PR #2, opened 2025-02-27
runtime: import code from zeam-runtime

gballet:
Import the current, committed state of zeam-runtime, so that the prover/verifier program can be built.

This does not support the full input yet since a common interface between zkvms needs to be designed (blockblaz/zeam-pm#9). The executable needs to be compiled before running the prover (which isn't implemented yet, so it's not a big deal at this point).

Fixes blockblaz/zeam-pm#5

### DIFF
--- a/pkgs/state-transition-runtime/build.zig
+++ b/pkgs/state-transition-runtime/build.zig
@@ -1,11 +1,29 @@
 const Builder = @import("std").Build;
 const std = @import("std");
 
+const zkvm_types = enum {
+    ceno,
+    powdr,
+    sp1,
+};
+
 pub fn build(b: *Builder) void {
     const target_query = .{ .cpu_arch = .riscv32, .os_tag = .freestanding, .abi = .none, .cpu_model = .{ .explicit = &std.Target.riscv.cpu.generic_rv32 } };
 
     const target = b.resolveTargetQuery(target_query);
     const optimize = b.standardOptimizeOption(.{});
+
+    // Declare the -Dzkvm option, which is a choice between all supported zkvms
+    const zkvm = b.option(zkvm_types, "zkvm", "zkvm target") orelse .powdr;
+    const zkvm_module = b.addModule("zkvm", .{
+        .optimize = optimize,
+        .target = target,
+        .root_source_file = b.path(switch (zkvm) {
+            .ceno => "src/ceno/lib.zig",
+            .powdr => "src/powdr/lib.zig",
+            .sp1 => "src/sp1/lib.zig",
+        }),
+    });
 
     // add ssz
     const ssz = b.dependency("ssz.zig", .{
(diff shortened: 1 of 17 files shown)


pkgs/state-transition-runtime/build.zig
```diff
@@ -39,26 +57,42 @@ pub fn build(b: *Builder) void {
...
+    // tests.root_module.addImport("ssz", ssz);
+    // tests.root_module.addImport("zeam-types", zeam_types);
+    // tests.root_module.addImport("zeam-transition-runtime", zeam_state_transition);
 
-    const run_tests = b.addRunArtifact(tests);
-    const test_step = b.step("test", "Run unit tests");
-    test_step.dependOn(&run_tests.step);
+    // const run_tests = b.addRunArtifact(tests);
+    // const test_step = b.step("test", "Run unit tests");
+    // test_step.dependOn(&run_tests.step);
```
gballet, line 97: I'm commenting this out, because these tests can never run (unless we are building this on riscv32). It might be a good idea to have the CI build them and run them in qemu, but this is an effort that is quite involved, and it can be tested another way by building the executable and running a prover instead. So I might end up removing it all, I just need to think about the best way to do this.

pkgs/state-transition-runtime/src/main.zig
```diff
@@ -30,18 +33,31 @@ pub fn main() noreturn {
     };
     _ = witnesses;
```
g11tech, line 34: we still have not accessed inputs and witnesses here right?

pkgs/state-transition-runtime/build.zig
```diff
@@ -1,12 +1,30 @@
...
+
 pub fn build(b: *Builder) void {
     const target_query = .{ .cpu_arch = .riscv32, .os_tag = .freestanding, .abi = .none, .cpu_model = .{ .explicit = &std.Target.riscv.cpu.generic_rv32 } };
 
     const target = b.resolveTargetQuery(target_query);
     const optimize = b.standardOptimizeOption(.{});
 
+    // Declare the -Dzkvm option, which is a choice between all supported zkvms
+    const zkvm = b.option(zkvm_types, "zkvm", "zkvm target") orelse .powdr;
+    const zkvm_module = b.addModule("zkvm", .{
```
g11tech, line 18: ok so we are inserting the target zkvm interface here that we have wrapped in each of individual folders?

pkgs/state-transition-runtime/src/main.zig
gballet: yes, this is coming in another PR that I'm currently working on, see blockblaz/zeam-pm#9

pkgs/state-transition-runtime/build.zig
gballet: not quite: depending on the `-Dzkvm=` option we passed to `zig build`, it will use the files corresponding to the correct zkvm. So it's only using one subdirectory at a time.

I want to expand the whole thing so that all the binaries are built and can be released, but this is not necessary at this stage.

g11tech: yes, this doesn't need to be part of this build but proving manager's build can trigger these builds one by one to generate the target binaries

> depending on the -Dzkvm= option we passed to zig build, it will use the files corresponding to the correct zkvm. So it's only using one subdirectory at a time

right what I meant was that one particular zk vm gets wrapped into "zkvm" module which is used by the zig `main`

g11tech (approved):
lfg :rocket:

PR #3, opened 2025-03-01
prove and verify with powdr

gballet:
WIP still

### DIFF
--- a/pkgs/state-proving-manager/src/manager.zig
+++ b/pkgs/state-proving-manager/src/manager.zig
@@ -1,15 +1,132 @@
-const types = @import("types")
-const state_transition = @import("state-transition")
+const std = @import("std");
+const types = @import("types");
+const state_transition = @import("state-transition");
+const allocator = std.heap.page_allocator;
 
+pub const ZKVMContext = struct {
+    program_path: []const u8,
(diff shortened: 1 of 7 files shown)


pkgs/state-proving-manager/src/manager.zig
```diff
@@ -1,15 +1,141 @@
...
+    var publics = std.ArrayList(u8).init(allocator);
+
+    var serialized = try std.ArrayList(u8).init(allocator);
+    defer serialized.deinit();
+    try ssz.serialize(types.SignedBeamBlock, block, serialized);
+
+    var input = try std.fmt.allocPrint(allocator, "{any}", .{serialized});
+
+    // Run the executable without proving, in order to recover the publics
+    var dryrun_process = std.process.Child.init(&[_][]const u8{
```
g11tech, line 31: ahh nice so we automating your poc demo here!
zkVMs would eventually be providing this programmatic api?

gballet: yes, it's currently possible to do that with rust libs - which I will work on integrating in a subsequent PR - right now I just want to do the whole end-to-end thing and calling external programs will do.

PR #4, opened 2025-03-03
build all zkvm binaries + cli at once

gballet:
Changes the build system to build all supported platforms in one go when typing `zig build`. We will further extend it to also build the full client down the road. This is a prerequisite for #3 

*EDIT*

This now builds every binary at once, using `zig build`. It's only building the tests for the state transition.

*TODO*

 - [x] build tests for all packages
 - [ ] rename the `pkgs/` subdirectory to `src/`, since all modules are no longer stored as independent packages
 - [ ] remove all other `src/` subdirectories
 - [x] rebase :(


g11tech:
i somehow feel the top level builds should be like triggering the all the package builds, and test trigger all package level tests

a monorepo kinda build system

gballet:
> i somehow feel the top level builds should be like triggering the all the package builds, and test trigger all package level tests
> 
> a monorepo kinda build system

yes, and I want to do that as well. I was just doing the first part, but I can expand.

g11tech (commented):
lgtm

PR #8, opened 2025-03-16
add and run clock event loop

g11tech:
add clock to primarily drive the client via libxev event loop

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -23,6 +23,14 @@ pub fn build(b: *Builder) !void {
         .target = target,
         .optimize = optimize,
     }).module("ssz.zig");
+    const zigcli = b.dependency("zigcli", .{
+        .target = target,
+        .optimize = optimize,
+    });
+    const xev = b.dependency("xev", .{
+        .target = target,
+        .optimize = optimize,
+    }).module("xev");
 
     // add zeam-params
     const params = b.addModule("zeam-params", .{
@@ -52,13 +60,6 @@ pub fn build(b: *Builder) !void {
         .optimize = optimize,
     });
 
-    // add beam node
-    const zeam_beam_node = b.addModule("zeam-node", .{
-        .target = target,
-        .optimize = optimize,
-        .root_source_file = b.path("pkgs/node/src/node.zig"),
-    });
-
     const st_lib = b.addStaticLibrary(.{
         .name = "zeam-state-transition",
         .root_source_file = b.path("pkgs/state-transition/src/lib.zig"),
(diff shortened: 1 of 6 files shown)


pkgs/node/src/clock.zig
```diff
@@ -1,3 +1,54 @@
...
+};
+
+fn timerCallback(
+    _: ?*void,
+    _: *xev.Loop,
+    _: *xev.Completion,
+    result: xev.Timer.RunError!void,
+) xev.CallbackAction {
+    _ = result catch unreachable;
+    std.debug.print("timer call back on event loop -----------\n", .{});
```
gballet, line 52: I recommend commenting out this line, so that it doesn't clutter the view of people working on other stuff

pkgs/node/src/utils.zig
```diff
@@ -0,0 +1,32 @@
...
+    mutex: Mutex,
+
+    const Self = @This();
+    pub fn init() !Self {
+        const loop = try xev.Loop.init(.{});
+        const mutex = Mutex{};
+
+        return Self{
+            .loop = loop,
+            .mutex = mutex,
```
gballet, line 19:
```suggestion
        return Self{
            .loop = loop,
            .mutex = Mutex{},
```

gballet: You are creating an object and copying it. Not a performance hit, but let's remove copies where we can.

pkgs/node/src/utils.zig
```diff
@@ -0,0 +1,32 @@
...
+
+pub const EventLoop = struct {
+    loop: xev.Loop,
+    // events from libp2p or other threads will also be pushed on it
+    mutex: Mutex,
+
+    const Self = @This();
+    pub fn init() !Self {
+        const loop = try xev.Loop.init(.{});
+        const mutex = Mutex{};
```
gballet, line 15: this mutex doesn't seem like it's being used at all :thinking:

build.zig
```diff
@@ -67,6 +68,14 @@ pub fn build(b: *Builder) !void {
...
     b.installArtifact(st_lib);
 
+    // add beam node
+    const zeam_beam_node = b.addModule("zeam-node", .{
+        .target = target,
+        .optimize = optimize,
+        .root_source_file = b.path("pkgs/node/src/lib.zig"),
+    });
+    zeam_beam_node.addImport("xev", xev);
+
```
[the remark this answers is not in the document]
g11tech: because node will later require other modules to be added as import

pkgs/node/src/utils.zig
g11tech: yes other events from other theads might also go into this loop

pkgs/node/src/utils.zig
g11tech: what is the better pattren, i don't want to maintain the pointer/reference in the main eventloop struct?

build.zig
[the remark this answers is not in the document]
gballet: hmmm yeah but I didn't get the impression that was the case. nvm.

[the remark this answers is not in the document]
gballet: ah yes of course, the state transition will be used in the future. makes sense.

pkgs/node/src/utils.zig
gballet: there's no pointer here, and they are equivalent. It's just saving one line of code and a copy (that is probably not happening anyway as I'm sure the compiler gets rid of it for us)

gballet (dismissed):
LGTM

PR #12, opened 2025-04-04
prepare chain config from zeam cli

g11tech:
adding boilerplates for Partial, merge/mix and to narrow cast the bigger struct to lower and use them in preparing the config for the initializing the node (followup PR)

### DIFF
--- a/pkgs/node/src/clock.zig
+++ b/pkgs/node/src/clock.zig
@@ -1,4 +1,6 @@
 const std = @import("std");
+const Allocator = std.mem.Allocator;
+
 const xev = @import("xev");
 
 const params = @import("@zeam/params");
@@ -12,6 +14,7 @@ pub const Clock = struct {
     current_slot_time_ms: isize,
     current_slot: isize,
     events: utils.EventLoop,
+    on_slot_cbs: std.ArrayList(*const fn (ud: ?*anyopaque, slot: isize) void),
 
     timer: xev.Timer,
     c: xev.Completion,
@@ -19,6 +22,7 @@ pub const Clock = struct {
     const Self = @This();
 
     pub fn init(
+        allocator: Allocator,
         genesis_time: usize,
     ) !Self {
         const events = try utils.EventLoop.init();
@@ -29,14 +33,7 @@ pub const Clock = struct {
         const current_slot = @divFloor(@as(isize, @intCast(std.time.milliTimestamp())) + CLOCK_DISPARITY_MS - genesis_time_ms, SECONDS_PER_SLOT_MS);
         const current_slot_time_ms = genesis_time_ms + current_slot * SECONDS_PER_SLOT_MS;
 
-        return Self{
-            .genesis_time_ms = genesis_time_ms,
-            .current_slot_time_ms = current_slot_time_ms,
-            .current_slot = current_slot,
-            .events = events,
-            .timer = timer,
-            .c = c,
-        };
+        return Self{ .genesis_time_ms = genesis_time_ms, .current_slot_time_ms = current_slot_time_ms, .current_slot = current_slot, .events = events, .timer = timer, .c = c, .on_slot_cbs = std.ArrayList(*const fn (ud: ?*anyopaque, slot: isize) void).init(allocator) };
     }
 
     pub fn tickSlot(self: *Self) void {
@@ -58,6 +55,8 @@ pub const Clock = struct {
             try self.events.run(.until_done);
         }
     }
+
+    pub fn onSlot() void {}
 };
 
 fn timerCallback(
--- a/pkgs/utils/src/cast.zig
+++ b/pkgs/utils/src/cast.zig
@@ -1,0 +1,15 @@
+const std = @import("std");
+
+pub fn Cast(To: type, from: anytype) To {
+    var result: To = undefined;
+    inline for (@typeInfo(To).@"struct".fields) |field| {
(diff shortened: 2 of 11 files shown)


pkgs/utils/src/cast.zig
```diff
@@ -0,0 +1,15 @@
+const std = @import("std");
+
+pub fn Cast(To: type, from: anytype) To {
+    var result: To = undefined;
+    inline for (@typeInfo(To).@"struct".fields) |field| {
+        const field_value = switch (@typeInfo(field.type)) {
+            .optional => @field(from, field.name),
+            // TODO: throw error instead of panic?
```
gballet, line 8: that'd be nice because the error can be handled at comptime, when the panic will be called at runtime iirc.

pkgs/utils/src/cast.zig
```diff
@@ -0,0 +1,15 @@
+const std = @import("std");
+
+pub fn Cast(To: type, from: anytype) To {
+    var result: To = undefined;
+    inline for (@typeInfo(To).@"struct".fields) |field| {
+        const field_value = switch (@typeInfo(field.type)) {
+            .optional => @field(from, field.name),
```
gballet, line 7: is this what you want to do here? because the way I see it, if the target field isn't an optional and the source field is, then you need to check that the source field isn't `null` and either copy the data if it's there, or return an error/panic otherwise.

But here, what I understand is that you're trying to do the opposite: if the target is an optional, then you assign the value of the source (whether it's the same or not). That should work out of the box without having to do it explicitly, right?

gballet: see https://zig.fly.dev/p/UROz_wVocn8 for what I mean

pkgs/utils/src/mix.zig
```diff
@@ -0,0 +1,78 @@
...
+
+fn indexByName(comptime fields: []const std.builtin.Type.StructField, name: []const u8) ?usize {
+    for (fields, 0..) |field, i| {
+        if (std.mem.eql(u8, field.name, name)) return i;
+    }
+    return null;
+}
+
+/// Mixes fields from structure extend into structure super
+pub fn Mix(comptime Super: type, comptime Extend: type) type {
```
gballet, line 24: I know this as a "Mix in" so I recommend renaming this `MixIn`

pkgs/utils/src/mix.zig
```diff
@@ -0,0 +1,78 @@
...
+    const superInfo = ensure(Super, .@"struct") orelse @panic("Super type must be a struct");
+    const extendInfo = ensure(Extend, .@"struct") orelse @panic("Extend type must be a struct");
```
gballet, line 26: not that I recommend doing that in this PR, but none of them need to be a struct, you could just `switch` on the type. Let's not complicate this PR though.

pkgs/utils/src/cast.zig
```diff
@@ -0,0 +1,15 @@
+const std = @import("std");
+
+pub fn Cast(To: type, from: anytype) To {
+    var result: To = undefined;
+    inline for (@typeInfo(To).@"struct".fields) |field| {
```
gballet, line 5: might be worth ignoring fields if they don't exist in one or the other structure? or make it an option at least.

pkgs/node/src/clock.zig
```diff
@@ -29,14 +33,7 @@ pub const Clock = struct {
...
 
-        return Self{
-            .genesis_time_ms = genesis_time_ms,
-            .current_slot_time_ms = current_slot_time_ms,
-            .current_slot = current_slot,
-            .events = events,
-            .timer = timer,
-            .c = c,
-        };
+        return Self{ .genesis_time_ms = genesis_time_ms, .current_slot_time_ms = current_slot_time_ms, .current_slot = current_slot, .events = events, .timer = timer, .c = c, .on_slot_cbs = std.ArrayList(*const fn (ud: ?*anyopaque, slot: isize) void).init(allocator) };
```
gballet, line 36: autoformat? it's less readable this way

pkgs/node/src/clock.zig
```diff
@@ -58,6 +55,8 @@ pub const Clock = struct {
             try self.events.run(.until_done);
         }
     }
+
+    pub fn onSlot() void {}
```
gballet, line 59: doesn't seem to be used anywhere

pkgs/utils/src/mix.zig
g11tech: most of this is copied snipped so yea we can trim it down later as you said

pkgs/node/src/clock.zig
g11tech: yea how to do it into multiline? wanted to ask you this (it auto formats)

pkgs/utils/src/cast.zig
g11tech: i needed to do this if target wasn't an optional and source was an optional non null type, to make sure the target type can be properly interpreted in else section. otherwise it was giving compile error

pkgs/utils/src/cast.zig
gballet: I mean to also ignore fields that are missing in the from, but it's ok, let's implement it when we need it.

pkgs/node/src/clock.zig
gballet: you need to add a comma after the last entry. I'll merge and push the fix on master.

gballet (approved):

PR #18, opened 2025-04-12
mock prove a chain segment with powdr

g11tech:
prove a mock chain segment

TODO 
- [ ] fix the zkvm input deserialization

image

### DIFF
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -57,33 +57,30 @@ pub fn main() !void {
             try clock.run();
         },
         .prove => |provecmd| {
-            const state = types.BeamState{
-                .genesis_time = 0,
-                .slot = 0,
-                .latest_block_header = .{
-                    .slot = 0,
-                    .proposer_index = 0,
-                    .parent_root = [_]u8{0} ** 32,
-                    .state_root = [_]u8{0} ** 32,
-                    .body_root = [_]u8{0} ** 32,
-                },
-            };
-            const block = types.SignedBeamBlock{
-                .message = .{
-                    .slot = 0,
-                    .proposer_index = 0,
-                    .parent_root = [_]u8{ 199, 128, 9, 253, 240, 127, 197, 106, 17, 241, 34, 55, 6, 88, 163, 83, 170, 165, 66, 237, 99, 228, 76, 75, 193, 95, 244, 205, 16, 90, 179, 60 },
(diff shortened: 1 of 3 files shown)


pkgs/state-transition/src/transition.zig
```diff
@@ -66,11 +66,15 @@ pub fn verify_signatures(signedBlock: types.SignedBeamBlock) !void {
 
 // TODO(gballet) check if beam block needs to be a pointer
 pub fn apply_transition(allocator: Allocator, state: *types.BeamState, signedBlock: types.SignedBeamBlock) !void {
+    const block = signedBlock.message;
+    if (block.slot <= state.slot) {
```
g11tech, line 70: one shouldn't be able to apply past slot blocks on the pre state, so fixed that

pkgs/cli/src/main.zig
```diff
@@ -57,33 +57,30 @@ pub fn main() !void {
...
+            // generate a mock chain with 2 blocks including genesis i.e. 1 block on top of genesis
+            const mock_config = types.GenesisSpec{
+                .genesis_time = 0,
+            };
+            const mock_chain = try sftFactory.genMockChain(allocator, 2, mock_config);
+
+            // starting beam state
+            var beam_state = mock_chain.genesis_state;
+            // block 0 is genesis so we have to apply block 1 onwards
+            for (1..mock_chain.blocks.len) |i| {
```
gballet, line 74:
```suggestion
            for (mock_chain.blocks, 1..) |block, i| {
```

pkgs/cli/src/main.zig
```diff
@@ -57,33 +57,30 @@ pub fn main() !void {
...
+                // this is a signed block
+                const block1 = mock_chain.blocks[i];
+                std.debug.print("block {d}: {any}", .{ i, block1 });
+                std.debug.print("\nprestate slot blockslot={d} stateslot={d}", .{ block1.message.slot, beam_state.slot });
+
+                _ = try stateProvingManager.prove_transition(beam_state, block1, options, allocator);
```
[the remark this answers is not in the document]
gballet: I debugged a bit and ran into a deserialization itself runs into a problem, I will keep digging.

gballet (commented):
I confirm there is an issue that has to do with deserialization. I need to dig deeper.

gballet (approved):
Not sure yet why the tests fail on a missing powdr library, since it seems to build, but I will investigate that in the main branch since the problem is already there. All I need right now, is to rebase my risc0 branch on top of that.

PR #19, opened 2025-04-13
get onslot callbacks on chain from clock

g11tech:
drive chain from clock

- [x] push the cb directly on the clock event loop

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -35,7 +35,7 @@ pub const BeamChain = struct {
         std.debug.print("chain received on slot cb at slot={d}\n", .{slot});
     }
 
-    pub fn onSlotWrapper(self: *Self) *OnSlotCbWrapper {
+    pub fn getOnSlotCbWrapper(self: *Self) *OnSlotCbWrapper {
         // need a stable pointer across threads
         const cb_ptr = self.allocator.create(OnSlotCbWrapper) catch @panic("mem alloc failed");
         cb_ptr.* = .{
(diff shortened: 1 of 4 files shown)


pkgs/node/src/chain.zig
```diff
@@ -4,18 +4,45 @@ const Allocator = std.mem.Allocator;
...
+    }
+
+    fn printSlot(self: *Self, slot: isize) void {
+        _ = self;
+        std.debug.print("chain received on slot cb at slot={d}\n", .{slot});
+    }
+
+    pub fn getOnSlotCbWrapper(self: *Self) *OnSlotCbWrapper {
+        // need a stable pointer across threads
+        const cb_ptr = self.allocator.create(OnSlotCbWrapper) catch @panic("mem alloc failed");
```
gballet, line 40: in general, avoid panics if you can.

pkgs/node/src/clock.zig
```diff
@@ -64,15 +95,9 @@ pub const Clock = struct {
         }
     }
 
-    pub fn onSlot() void {}
+    pub fn subscribeOnSlot(self: *Self, cb: *OnSlotCbWrapper) !void {
+        try self.on_slot_cbs.append(cb);
+        var c: xev.Completion = undefined;
+        try self.on_slot_cs.append(&c);
```
gballet, line 101: I don't know `xev.Completion` but I don't think that will work: you are passing a pointer to a stack-allocated object,

gballet (approved):

PR #20, opened 2025-04-20
zkvm: display stf logs in zkvm context

gballet:
`std.debut.print` will not work in a zkvm context, as there is no unix support in these environments. I have created a `log` function that behaves like it, and uses either `zkvm.io.print_str` or `std.debug.print` depending on the context it's run in.

Note that it will only be able to print a maximum of 512 characters at a time, but that is good enough for our current use case.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -187,21 +187,23 @@ fn build_zkvm_targets(b: *Builder, main_exe: *Builder.Step) !void {
     });
     zeam_types.addImport("@zeam/params", zeam_params);
 
-    // add state transition
-    const zeam_state_transition = b.addModule("@zeam/state-transition", .{
-        .root_source_file = b.path("pkgs/state-transition/src/lib.zig"),
-        .target = target,
-        .optimize = optimize,
-    });
-    zeam_state_transition.addImport("@zeam/types", zeam_types);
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -187,22 +187,24 @@ fn build_zkvm_targets(b: *Builder, main_exe: *Builder.Step) !void {
...
     for (zkvm_targets) |zkvm_target| {
         const zkvm_module = b.addModule("zkvm", .{
             .optimize = optimize,
             .target = target,
             .root_source_file = b.path(b.fmt("pkgs/state-transition-runtime/src/{s}/lib.zig", .{zkvm_target.name})),
         });
 
+        // add state transition, create a new module for each zkvm since
+        // that module depends on the zkvm module.
+        const zeam_state_transition = b.addModule("@zeam/state-transition", .{
```
g11tech, line 199: can't we only add the `zkvm` addImport to this module inside the loop and else keep it outside the loop?

gballet: no because the module itself is different for each zkvm binary, the import is a reference to the module so we can't do that afaik.

g11tech (approved):

PR #22, opened 2025-04-27
add a basic functional forkchoice

g11tech:
- [x] code complete
- [x] test cases
  - [x]  genStateBlockHeader
  - [x] fc onblock


pkgs/tests.zig
```diff
@@ -1,5 +1,5 @@
 comptime {
-    _ = @import("state-transition/src/lib.zig");
+    // _ = @import("state-transition/src/lib.zig");
```
g11tech, line 2: uncommenting causes 
image

testing boilerplate needs fixing

pkgs/node/src/forkchoice.zig
```diff
@@ -1,69 +1,298 @@
...
+        return false;
+    }
+
+    pub fn tickSlot(self: *Self, currentSlot: types.Slot) void {
+        if (self.fcStore.currentSlot >= currentSlot) {
+            return;
+        }
+
+        self.fcStore.currentSlot = currentSlot;
+        // reset attestations or process checkpoints as prespscribed in the specs
```
gballet, line 208: typo: prescribed

pkgs/tests.zig
gballet: it does

gballet: I have been investigating why this fails a bit already, the fact you broke it is going to help me further. Wait a bit before merging this PR, it's helpful for me to fix the issue in this state.

pkgs/utils/src/extend.zig
gballet, line 1: are you planning to merge this PR as is? because if so, I suggest you remove this PR and the comment about `Extend` above. No point in merging broken code.

pkgs/utils/src/extend.zig
```diff
@@ -0,0 +1,18 @@
...
+pub usingnamespace castFactory;
+
+pub fn Extend(To: type, from: anytype, with: anytype) To {
+    var result: To = undefined;
+    inline for (@typeInfo(To).@"struct".fields) |field| {
+        const extend_value = @field(with, field.name) orelse @field(from, field.name);
+
+        const field_value = switch (@typeInfo(field.type)) {
+            .optional => extend_value,
+            // TODO: throw error instead of panic?
```
gballet, line 12: yes, because this is runtime code

pkgs/utils/src/extend.zig
```diff
@@ -0,0 +1,18 @@
+const std = @import("std");
+const castFactory = @import("./cast.zig");
+pub usingnamespace castFactory;
+
+pub fn Extend(To: type, from: anytype, with: anytype) To {
```
gballet, line 5: I don't see anything obviously wrong, what is the error that is being produced?

g11tech: will create a separate extend PR and add a testcase which I would love to have to see/debug to make it work

pkgs/utils/src/extend.zig
g11tech: yea removed it for now

pkgs/tests.zig
g11tech: yea uncommenting `_ = @import("state-transition/src/lib.zig");`  will show up the error. but lets merge the PR in current state itself to unblock the dev and we can address this a bit down the line I guess

gballet (approved):

PR #24, opened 2025-05-11
implement mini 3sf for beamchain experimentation

g11tech:
translate https://github.com/ethereum/research/blob/master/3sf-mini/consensus.py for zeam 3sf experimentation

- [x] impl/translate state transition
- [x] generate, add, run/fix test scenarios

#### How to Run?
1. install risc0 version: 3.0.0-rc1
2. run `zig build -Doptimize=ReleaseFast install run -- prove`

### DIFF
--- a/pkgs/state-transition/src/lib.zig
+++ b/pkgs/state-transition/src/lib.zig
@@ -57,57 +57,6 @@ test "apply transition on mocked chain" {
     try std.testing.expect(std.mem.eql(u8, &post_state_root, &mock_chain.blocks[mock_chain.blocks.len - 1].message.state_root));
 }
 
-test "mock genesis and block production" {
-    // 1. setup genesis config
-    const test_config = types.GenesisSpec{
-        .genesis_time = 1234,
-        .num_validators = 4,
-    };
-
-    var arena_allocator = std.heap.ArenaAllocator.init(std.testing.allocator);
-    defer arena_allocator.deinit();
-    const allocator = arena_allocator.allocator();
-
-    const mock_chain = try genMockChain(allocator, 2, test_config);
-
-    // check genesis state root
-    var test_genesis_root: [32]u8 = undefined;
-    try ssz.hashTreeRoot(types.BeamState, mock_chain.genesis_state, &test_genesis_root, std.testing.allocator);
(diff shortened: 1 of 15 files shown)


pkgs/state-transition/src/lib.zig
```diff
@@ -17,28 +17,29 @@ pub const StateTransitionOpts = transition.StateTransitionOpts;
...
-}
+// test "ssz import" {
+//     const data: u16 = 0x5566;
+//     const serialized_data = [_]u8{ 0x66, 0x55 };
+//     var list = std.ArrayList(u8).init(std.testing.allocator);
+//     defer list.deinit();
+
+//     try ssz.serialize(u16, data, &list);
+//     try std.testing.expect(std.mem.eql(u8, list.items, serialized_data[0..]));
+// }
```
gballet, line 28: why is that commented out?

pkgs/state-transition/src/lib.zig
```diff
@@ -56,85 +57,36 @@ test "apply transition on mocked chain" {
...
-        try std.testing.expect(std.mem.eql(u8, &applied_block_root, &state_block_header_root));
-
-        if (i < mock_chain.blocks.len - 1) {
-            // apply the next block
-            const block = mock_chain.blocks[i + 1];
-            try apply_transition(allocator, &beam_state, block, .{});
-        }
-    }
-}
+// test "genStateBlockHeader" {
```
gballet, line 60: same here: I don't recommend disabling tests

g11tech: yes, uncommenting this

pkgs/state-transition/src/lib.zig
```diff
@@ -56,85 +57,36 @@ test "apply transition on mocked chain" {
     try std.testing.expect(std.mem.eql(u8, &post_state_root, &mock_chain.blocks[mock_chain.blocks.len - 1].message.state_root));
 }
 
-test "mock genesis and block production" {
```
g11tech, line 59: removed this with hardcoded data as the things are really dynamic for now

gballet (approved):

PR #26, opened 2025-06-02
abstract out logging functionality in an extendable way

g11tech:
a little abstraction on the logging

can be further extended to creation of the child scopes as well as add file transport in future PR work

result:

```
[debug] (zkvm): apply transition stateslot=0 blockslot=1
state transition completed
proof len=329557
[debug] (zeam): apply transition stateslot=0 blockslot=1
```

so most logs can be converted into this format going forward, even date/time can be added in the beginning so have the logs we are used to see in clients

### DIFF
--- a/pkgs/state-transition/src/transition.zig
+++ b/pkgs/state-transition/src/transition.zig
@@ -87,10 +87,10 @@ pub fn apply_transition(allocator: Allocator, state: *types.BeamState, signedBlo
     var logger = getLogger();
     logger.setActiveLevel(opts.activeLogLevel);
     const block = signedBlock.message;
-    logger.debug("apply transition stateslot={d} blockslot={d}\n", .{ state.slot, block.slot }) catch @panic("appply transition start log");
+    logger.debug("apply transition stateslot={d} blockslot={d}\n", .{ state.slot, block.slot });
 
     if (block.slot <= state.slot) {
-        logger.debug("slots are invalid for block {any}: {} >= {}\n", .{ block, block.slot, state.slot }) catch @panic("error printing block and state slots");
+        logger.debug("slots are invalid for block {any}: {} >= {}\n", .{ block, block.slot, state.slot });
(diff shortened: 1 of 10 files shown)


pkgs/utils/src/log.zig
```diff
@@ -0,0 +1,106 @@
...
+
+    const scope_prefix = "(" ++ switch (scope) {
+        std.log.default_log_scope => system_prefix,
+        else => system_prefix ++ "-" ++ @tagName(scope),
+    } ++ "): ";
+    const prefix = "[" ++ comptime level.asText() ++ "] " ++ scope_prefix;
+
+    if (builtin.target.os.tag == .freestanding) {
+        const io = @import("zkvm").io;
+        var buf: [1024]u8 = undefined;
```
gballet, line 21: It'd be better to use an allocator here, since some zkvms don't have a very large stack. You could do have a singleton logger that takes the allocator in its `init` funciton.

pkgs/state-transition/src/transition.zig
```diff
@@ -1,20 +1,17 @@
...
-    if (builtin.target.os.tag == .freestanding) {
-        const io = @import("zkvm").io;
-        var buf: [512]u8 = undefined;
-        io.print_str(try std.fmt.bufPrint(buf[0..], fmt, args));
-    } else {
-        std.debug.print(fmt, args);
-    }
-}
+// put the active logs at debug level for now by default
+pub const StateTransitionOpts = struct { activeLogLevel: std.log.Level = std.log.Level.debug };
```
gballet, line 14: This is ok as long as we're developing, but when we start making releases and showing our performance, we should not make debug the default. It should in fact depend on the build optimization.

g11tech: in the guest program I agree, but I like the ability to change log levels via env in the least (and with keystrokes or api call on a running client).

PR #31, opened 2025-06-26
drop powdr

gballet:
This one is more of a heartbreak than #30, given how much more time was invested in making this work. This being said, the powdr team decided to start using openvm as their vm frontend, so there is no longer a need to support this target.


gballet:
Only deactivate it for now, we might want to reactivate it for their WASM VM.

PR #35, opened 2025-07-29
setup multi chain single node via in process mock network

g11tech:


gballet (approved):
LGTM. But as copilot says in one of its comments, you should not be continuing if there's an error.

PR #42, opened 2025-08-13
integrate eth libp2p network via rust libp2p bridge

g11tech:
WIP


pkgs/node/src/chain.zig
```diff
@@ -168,7 +170,8 @@ test "process and add mock blocks into a node's chain" {
 
     const mock_chain = try stf.genMockChain(allocator, 5, chain_config.genesis);
     const beam_state = mock_chain.genesis_state;
-    var beam_chain = try BeamChain.init(allocator, chain_config, beam_state);
+    const nodeid = 10; // random value
```
gballet, line 173: @g11tech this is one of the test issues I had to fix, please review because I picked something at random

pkgs/params/src/lib.zig
```diff
@@ -14,5 +14,5 @@ const activePresetValues = @field(presets, @tagName(activePreset));
 pub const SECONDS_PER_SLOT = activePresetValues.SECONDS_PER_SLOT;
 
 test "test preset loading" {
-    try std.testing.expect(SECONDS_PER_SLOT == 4);
+    try std.testing.expect(SECONDS_PER_SLOT == mainnetPreset.preset.SECONDS_PER_SLOT);
```
gballet, line 17: @g11tech I also had to fix that, but please check if the choice makes sense

pkgs/node/src/chain.zig
g11tech: yes this is good :+1:

pkgs/params/src/lib.zig
g11tech: reverted the slot time back, it was a debugging artifact, thanks for pointing :+1: :pray:

g11tech:
Merging this for now

@noopur23 can you create issues for all the todos listed in the PR as well as the entire codebase, we should start handing out these to those who want to contribute

cc @gballet

PR #64, opened 2025-08-20
feat: add logger in chain and forkChoice

noopur23:
accept  logger pointer in chain and forkChoice and use it rather than std.debug.print

WIP

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -10,7 +10,6 @@ const networks = @import("@zeam/network");
 const params = @import("@zeam/params");
 
 const zeam_utils = @import("@zeam/utils");
-const getLogger = zeam_utils.getLogger;
 
 pub const fcFactory = @import("./forkchoice.zig");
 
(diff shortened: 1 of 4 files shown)


pkgs/cli/src/main.zig
```diff
@@ -197,13 +197,28 @@ pub fn main() !void {
...
+                .allocator = beam_node_1.chain.allocator,
+                .states = beam_node_1.chain.states,
+            }});
 
             try beam_node_1.run();
             try beam_node_2.run();
             try clock.run();
 
-            std.debug.print("forkchoice={any}\n", .{beam_node_1.chain.forkChoice});
+            std.debug.print("chainoptionsinfo: {any}\n", .{.{
```
g11tech, line 212: whats the issue here?

pkgs/node/src/chain.zig
```diff
@@ -26,10 +26,17 @@ pub const BeamChain = struct {
...
 
     const Self = @This();
-    pub fn init(allocator: Allocator, config: configs.ChainConfig, anchorState: types.BeamState, nodeId: u32) !Self {
-        const fork_choice = try fcFactory.ForkChoice.init(allocator, config, anchorState);
+    pub fn init(
+        allocator: Allocator,
+        config: configs.ChainConfig,
+        anchorState: types.BeamState,
+        nodeId: u32,
+        logger: *zeam_utils.ZeamLogger,
```
g11tech, line 37: create a params struct that you pass here and put config, anchorstate nodeid, logger into it

pkgs/node/src/chain.zig
```diff
@@ -82,14 +90,12 @@ pub const BeamChain = struct {
             },
         };
 
-        std.debug.print("\n\n\n node-{d}::going for block production opts={any} raw block={any}\n\n", .{ self.nodeId, opts, block });
+        // self.logger.debug("\n\n\n node-{d}::going for block production opts={any} raw block={any}\n\n", .{ self.nodeId, opts, block });
```
g11tech, line 93: also remove extra \ns when you are going for cleanup

pkgs/cli/src/main.zig
g11tech: also we can remove this and the above std debug as it was used for intial debugging (in followup PR)

noopur23: error: unable to format function body type, use '*const fn (comptime @Type(.enum_literal), log.Level, comptime log.Level, comptime []const u8, anytype) void' for a function pointer type
        .@"fn" => @compileError("unable to format function body type, use '*const " ++ @typeName(T) ++ "' for a function pointer type"),

g11tech (approved):

PR #90, opened 2025-08-23
feat: integrate metrics lib and add a sample e2e chain process block metric

bomanaps:
# PR Summary

- **Metrics Service:** Implemented  using `karlseguin/metrics.zig`.
- **Histogram:** Added a histogram `chain_onblock_duration_seconds` to measure `onBlock` and `apply_transition` execution time.
- **Metrics Server:** Metrics server now runs on port `9667`.
- **Grafana Dashboard:** Added `Zeam Application Metrics` dashboard with automatic provisioning.
- **Configuration Updates:** Updated `prometheus.yml`, added `pkgs/metrics/README.md`, and updated `.gitignore`.
closes https://github.com/blockblaz/zeam/issues/61. https://github.com/blockblaz/zeam/issues/61

Screenshot 2025-08-24 at 17 27 34

Screenshot 2025-08-24 at 17 27 17

### DIFF
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -154,14 +154,10 @@ pub fn main() !void {
         },
         .beam => |beamcmd| {
             try metrics.init(allocator);
-            
+
             // Start metrics HTTP server
-            if (!metrics.isZKVM()) {
-                try metricsServer.startMetricsServer(allocator, beamcmd.metricsPort);
-            } else {
-                std.log.info("Skipping HTTP server for ZKVM target", .{});
-            }
-            
+            try metricsServer.startMetricsServer(allocator, beamcmd.metricsPort);
+
             std.debug.print("beam opts ={any}\n", .{beamcmd});
 
             const mock_network = beamcmd.mockNetwork;


pkgs/metrics/src/noop.zig
```diff
@@ -0,0 +1,34 @@
+const std = @import("std");
+
+pub const Timer = struct {
```
g11tech, line 3: I am assuming noop is not being used as of now (since its primarily for future use in the STF which get proved )

bomanaps: Yes

pkgs/metrics/grafana/provisioning/dashboards/dashboards.yml
gballet, line 1: we shouldn't store the grafana dashboards here. We should geth ethpandaops to run them for us, they have their own infra for this - no need to bloat this repo with that stuff.

bomanaps:
Hello @gballet and @g11tech @noopur23 please can I get a review on this?

pkgs/cli/src/main.zig
```diff
@@ -35,6 +36,12 @@ const ZeamArgs = struct {
         beam: struct {
             help: bool = false,
             mockNetwork: bool = false,
+            metricsPort: u16 = 9667,
+        },
+        generate_prometheus_config: struct {
+            help: bool = false,
+            metricsPort: u16 = 9667,
+            output: []const u8 = "pkgs/metrics/prometheus/prometheus.yml",
```
gballet, line 44: default behavior should be to dump in the current directory. This only works because you are running it inside the zeam repo, but people will get zeam via docker images, binary distributions, etc... so don't assume `pkgs/metrics/prometheus` will be there.

pkgs/metrics/src/lib.zig
```diff
@@ -0,0 +1,195 @@
+const std = @import("std");
+const metrics_lib = @import("metrics");
+
+// Platform-specific time function
+fn getTimestamp() i128 {
+    // For freestanding targets, we might not have access to system time
+    // In that case, we'll use a simple counter or return 0
+    if (@import("builtin").target.os.tag == .freestanding) {
```
gballet, line 8: an issue with this, is that some zkvms are going to emulate linux, and freestanding won't be a good way to determine if something is in a zkvm context. It will be ok for now, but please implement a `isZKVM() bool` function, that uses this code, so that we only have one place to fix this in.

pkgs/cli/src/main.zig
```diff
@@ -35,6 +36,12 @@ const ZeamArgs = struct {
         beam: struct {
             help: bool = false,
             mockNetwork: bool = false,
+            metricsPort: u16 = 9667,
+        },
+        generate_prometheus_config: struct {
```
g11tech, line 41: can you also move remove this out for now and add this to a utils cli package that we will be adding

pkgs/metrics/src/lib.zig
```diff
@@ -0,0 +1,195 @@
...
+        // For freestanding environments, we can't measure real time
+        // Return 0 for now - in a real implementation you'd want a cycle counter
+        return 0;
```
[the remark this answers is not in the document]
g11tech: pls do not resolve the comments yourself unless you are 110% sure that you have resolved as per the solution asked.

for this current comment, since we will have a noop metrics and will not even call this in zkvm, can you remove this if part?

pkgs/metrics/src/lib.zig
```diff
@@ -0,0 +1,202 @@
...
+        .port = port,
+    };
+
+    const thread = try std.Thread.spawn(.{}, ServerContext.run, .{ctx});
+    thread.detach();
+}
+
+/// Generates a Prometheus configuration file content based on the metrics port.
+/// This can be used to create a prometheus.yml file that matches the CLI arguments.
+pub fn generatePrometheusConfig(allocator: std.mem.Allocator, metrics_port: u16) ![]u8 {
```
g11tech, line 180: remove this from the current PR and move to a followup PR which you will do for adding this functionality to utils cli

pkgs/metrics/src/lib.zig
```diff
@@ -0,0 +1,202 @@
...
+    if (isZKVM()) {
+        try writer.writeAll("# Metrics disabled for ZKVM target\n");
+        return;
+    }
+
+    try metrics_lib.write(&metrics, writer);
+}
+
+/// Starts a simple HTTP server to serve metrics on /metrics endpoint.
+pub fn startListener(allocator: std.mem.Allocator, port: u16) !void {
```
g11tech, line 116: this needs to move out from this file, to a file: routes.zig in the same directory, and it needs to accept an httpz instance to setup its route on that server.

httpz instance needs to started on the main cli and metrics need to be initialized/instantiated from there

pkgs/utils/src/cli.zig
gballet, line 1: so when I suggested to put it in `utils`, I meant to move it to a _subdirectory_ of `utils` since there might be more utilities in the future.

pkgs/cli/src/main.zig
gballet: so FYI I have moved that back in, I think it's fine to have that in there for now. Let's go through pq-devnet-0 and improve once we have launched it.

pkgs/metrics/src/lib.zig
gballet: right so @bomanaps we would start the server when running `zeam beam`, possibly guarded by a `--metrics` option.

pkgs/metrics/README.md
```diff
@@ -0,0 +1,120 @@
...
+1. **Clone the dashboard repository**:
+   ```sh
+   git clone https://github.com/blockblaz/zeam-dashboards.git
+   cd zeam-dashboards
+   ```
+
+2. **Generate Prometheus configuration**:
+   ```sh
+   # From your Zeam repository
+   ./zig-out/bin/zeam generate_prometheus_config --output prometheus/prometheus.yml
```
gballet, line 55:
```suggestion
   ./zig-out/bin/zeam prometheus genconfig -f prometheus/prometheus.yml
```

pkgs/cli/src/main.zig
```diff
@@ -153,8 +154,14 @@ pub fn main() !void {
         },
         .beam => |beamcmd| {
             try metrics.init(allocator);
-            // TODO: Phase 3 - Integrate new HTTP server architecture with metrics routes
-            // try metrics.startListener(allocator, beamcmd.metricsPort);
+            
+            // Start metrics HTTP server
+            if (!metrics.isZKVM()) {
```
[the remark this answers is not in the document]
gballet:
```suggestion
```

gballet (approved):
LGTM.

PR #92, opened 2025-08-24
Add macOS frameworks to Rust glue

bomanaps:
Add macOS-specific framework links (`CoreFoundation`, `SystemConfiguration`) in the Rust glue library build to ensure compatibility on macOS targets.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -15,8 +15,13 @@ const zkvm_targets: []const zkvmTarget = &.{
 };
 
 // Add the glue libs to a compile target
-fn addZkvmGlueLibs(b: *Builder, comp: *Builder.Step.Compile) void {
+fn addRustGlueLib(b: *Builder, comp: *Builder.Step.Compile, target: Builder.ResolvedTarget) void {
     comp.addObjectFile(b.path("rust/target/release/librustglue.a"));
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -203,6 +207,12 @@ pub fn build(b: *Builder) !void {
         .target = target,
     });
     manager_tests.root_module.addImport("@zeam/types", zeam_types);
+    addZkvmGlueLibs(b, manager_tests);
```
GrapeBaBa, line 210: Why add this to manager_tests? may be CI failed related

GrapeBaBa: But anyway, I tested you branch on linux and no errors, it is a bit odd.

gballet (approved):
So the reason why this was failing, is that now that the tests are also dependent on the rust library (which makes no sense to me at this point, but we can definitely imagine this being true in the future), we need to ensure that the library is built before we link the tests. Now, on local machines that works because typically the library has previously been built and the environment hasn't been "cleaned up". But in CI, the library doesn't currently exist. So by enforcing the right order, the problem disappears. This is not a MacOS thing.

This being said, I also added macos builds to CI, since this PR is primarily about that.

PR #94, opened 2025-08-24
ci: fix rust caching

### LINKED ISSUE #40: ci: fix rust dependency cache in risc0 job
The risc0 job doesn't seem to cache the dependencies, this needs to be fixed.

see https://github.com/blockblaz/zeam/actions/runs/16853057477/job/47742108064#step:5:56 for example

It might also be the case for the regular build, in which case this needs to be fixed too.

gballet:
Fixes #40

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Cache dependencies
       uses: Swatinem/rust-cache@v2
       with:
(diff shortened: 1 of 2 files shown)


gballet:
Doesn't seem to work, I wonder if this rust-specific github action really works

g11tech:
> Doesn't seem to work, I wonder if this rust-specific github action really works

it works on PR merges to main

PR #129, opened 2025-08-28
add timestamp to the logger in pretty print format

noopur23:
Prefixes datetime in MMM dd HH:mm:ss mss format 

Aug 28 21:08:41 000 [debug] (zeam-n2): processed missed_slots=0

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -44,13 +44,18 @@ pub fn build(b: *Builder) !void {
         .target = target,
         .optimize = optimize,
     }).module("xev");
+    const pg = b.dependency("datetime", .{
+        .target = target,
+        .optimize = optimize,
+    });
 
-    // add zeam-params
+    // add zeam-utils
     const zeam_utils = b.addModule("@zeam/utils", .{
         .target = target,
         .optimize = optimize,
         .root_source_file = b.path("pkgs/utils/src/lib.zig"),
     });
+    zeam_utils.addImport("datetime", pg.module("datetime"));
 
     // add zeam-params
     const zeam_params = b.addModule("@zeam/params", .{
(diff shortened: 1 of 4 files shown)


build.zig
```diff
@@ -288,6 +298,7 @@ fn build_zkvm_targets(b: *Builder, main_exe: *Builder.Step, host_target: std.Bui
             .optimize = optimize,
             .root_source_file = b.path("pkgs/utils/src/lib.zig"),
         });
+        zeam_utils.addImport("datetime", pg.module("datetime"));
```
g11tech, line 301: is not having this giving an error, basically zkvm targets are what runs inside zkvm i.e. freestanding os, and for that you have not used datetime, so ideally zig shouldn't even be asking for it when its building

pkgs/utils/src/log.zig
```diff
@@ -22,22 +23,30 @@ pub fn comptTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, com
         // TODO don't throw error because it somehow messes with creation of  noopLogger as noopLog
```
g11tech, line 23: remove this
```suggestion
```

g11tech: remove all three lines

pkgs/utils/src/log.zig
```diff
@@ -100,3 +106,26 @@ pub fn getScopedLogger(comptime scope: LoggerScope, activeLevel: ?std.log.Level)
 pub fn getLogger(activeLevel: ?std.log.Level) ZeamLogger {
     return ZeamLogger.init(std.log.default_log_scope, activeLevel orelse std.log.default_level);
 }
+
+pub fn getFormattedTimestamp(buf: []u8) []const u8 {
+    if (builtin.target.os.tag == .freestanding) {
```
g11tech, line 111: you don't need to do this if now since we dont even call in freestanding/inside zkvm

build.zig
```diff
@@ -44,13 +44,18 @@ pub fn build(b: *Builder) !void {
...
+    });
 
-    // add zeam-params
+    // add zeam-utils
     const zeam_utils = b.addModule("@zeam/utils", .{
         .target = target,
         .optimize = optimize,
         .root_source_file = b.path("pkgs/utils/src/lib.zig"),
     });
+    zeam_utils.addImport("datetime", datetime.module("datetime"));
```
g11tech, line 58: instead of doing .module later on at 2 places, just .module when you are first creating datetime

PR #132, opened 2025-08-31
feat:impl tools-cli and adding enrgen command

### LINKED ISSUE #130: Add a command in CLI to generate ENR
Original [from ](https://t.me/c/2873518363/332) TG discussion.

CLI design:
```sh
zeam enr gen --sk 0x1234567890abcdef... --udp 30303 --ip 192.168.1.100 --tcp 30304
```

@g11tech Is it like that

GrapeBaBa:
## Motivation
Currently we need a CLI command to generate ENR. Since discussed before, it need a new CLI which separates from `zeam` CLI. This PR introduces a new CLI which named `zeam-tools` and adds a sub-command `enrgen`. 

## Implementation
```sh
 USAGE:
     ./zig-out/bin/zeam-tools [OPTIONS] [COMMANDS]

 COMMANDS:
  enrgen     Generate a new ENR (Ethereum Node Record)

 OPTIONS:
  -h, --help                       Show help information
  -v, --version                    Show version information
```

```sh

 USAGE:
     ./zig-out/bin/zeam-tools enrgen [OPTIONS] 

 OPTIONS:
  -s, --sk STRING                  Secret key (hex string with or without 0x prefix)(required)
  -i, --ip STRING                  IPv4 address for the ENR(required)
  -u, --udp INTEGER                UDP port for discovery(required)
  -o, --out STRING                 Output file path (prints to stdout if not specified)(default: null)
  -h, --help                       Show help information for the enrgen command
```

Fix #130

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,6 +1,6 @@
 .{
     .name = .zeam,
-    .fingerprint = 0x243fd12cbfabbe73,
+    .fingerprint = 0x243fd12ce9ee69c2,
     .version = "0.0.0",
     .dependencies = .{
         .ssz = .{
@@ -24,8 +24,8 @@
             .hash = "metrics-0.0.0-W7G4eCG0AQCQXidzvV5kx4l0smr_WCw-8JLIwS_OHYoW",
         },
         .zig_enr = .{
-            .url = "git+https://github.com/blockblaz/enr#702fc2262cad4d387ef62385569534f8e20e499e",
-            .hash = "zig_enr-0.0.0-ak9ZXDk5AQCrVlosm0vlP__x0E5YX0oEhqaaYCf2Dq72",
+            .url = "git+https://github.com/blockblaz/enr#02f187591f8a9616623e0bc91d139f0d18c99bcf",
+            .hash = "zig_enr-0.0.0-cO2P20iiAQALUQA-fb_70u0hyoBJo7u_tdeZue8FuyuZ",
         },
     },
     .paths = .{""},


GrapeBaBa:
@gballet @g11tech can't request reviewer, ping you here.

pkgs/tools-cli/src/main.zig
```diff
@@ -0,0 +1,219 @@
...
+
+    pub const __messages__ = .{
+        .help = "Show help information",
+        .version = "Show version information",
+    };
+
+    const ENRGenCmd = struct {
+        sk: []const u8,
+        ip: []const u8,
+        udp: u16,
```
g11tech, line 31: i think instead of `udp` field we have `quic` field, see https://github.com/leanEthereum/leanSpec/blob/main/docs/client/networking.md

build.zig
```diff
@@ -185,6 +190,18 @@ pub fn build(b: *Builder) !void {
...
+        .name = "zeam-tools",
+        .root_source_file = b.path("pkgs/tools-cli/src/main.zig"),
+        .optimize = optimize,
+        .target = target,
+    });
+    tools_cli_exe.root_module.addImport("enr", enr);
+    tools_cli_exe.root_module.addImport("build_options", build_options.createModule());
+    tools_cli_exe.root_module.addImport("simargs", zigcli.module("simargs"));
+    b.installArtifact(tools_cli_exe);
+
```
gballet, line 204: the tool should have its own build, as we only want to build it if needed. You can add a `all` target that would build all the tools in the `pkgs/tools-cli` directory

build.zig.zon
```diff
@@ -19,6 +19,10 @@
             .url = "git+https://github.com/frmdstryr/zig-datetime?ref=master#3a39a21e6e34dcb0ade0ff828d0914d40ba535f3",
             .hash = "datetime-0.8.0-cJNXzP_YAQBxQ5hkNNP6ScnG5XsqciJmeP5RVV4xwCBA",
         },
+        .zig_enr = .{
```
gballet, line 22: didn't you need to regenerate the fingerprint?

pkgs/tools-cli/src/main.zig
gballet, line 1: There will presumably be more than one tool, so it should have another directory, e.g. `pkgs/tools/enrgen/src/main.zig`. I don't think we need `-cli`, tools are typically CLIs.

gballet (commented):
Looks like you need to add a dependency for this to compile in docker. My suggestion is to _not_ build this executable in docker, unless it's necessary for daily operations. If you disagree, fine, just add the dependency and call that `build-all` thing that I suggested.

pkgs/tools-cli/src/main.zig
GrapeBaBa: Current tools CLI design assume that multiple tools will be sub commands, So that we have only one CLI binary. Suggested dir seems not match this assumption. What is your opinion?

pkgs/tools-cli/src/main.zig
GrapeBaBa: Oh, it seems the spec is not good here, it doesn't need a quic field. there is already a standard udp field there.  Combine secp256k1, ip, udp, it can generate a ipv4 multiaddr. This is already used in beacon chain, i think we need change the spec here.

pkgs/tools-cli/src/main.zig
gballet: my opinion is that, unless I'm mistaken, this is not part of the main client. If it is, then this needs to be a subcommand and if that is the case, it needs to be in `pkgs/cli` and not aside.

GrapeBaBa: I agree this is a new CLI, but I don't think we should have a directory like `pkgs/tools/enrgen/src/main.zig`. What I thought is `pkgs/tools/src/main.zig`, the `enrgen` is a subcommand of this CLI. When we add new tools, we don't need add new `main.zig` and just add another subcommand in `pkgs/tools/src/main.zig`.

pkgs/tools-cli/src/main.zig
GrapeBaBa: ok, I saw here https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/discovery/enr_ext.rs#L9, let me add `quic` field.

PR #134, opened 2025-09-02
Enabling file logging and log rotation

noopur23:
WIP


pkgs/cli/src/main.zig
```diff
@@ -216,8 +216,8 @@ pub fn main() !void {
             var validator_ids_1 = [_]usize{1};
             var validator_ids_2 = [_]usize{2};
 
-            const logger1 = utilsLib.getScopedLogger(.n1, .debug);
-            const logger2 = utilsLib.getScopedLogger(.n2, .debug);
+            const logger1 = utilsLib.getScopedLogger(.n1, .debug, "./log");
```
g11tech, line 219: are the two loggers able to interleave into the same file?

also add the file name as a parameter in the cli args but with a default, and default file name should be `consensus.log`

pkgs/cli/src/main.zig
```diff
@@ -122,7 +128,7 @@ pub fn main() !void {
         },
         .prove => |provecmd| {
             std.debug.print("distribution dir={s}\n", .{provecmd.dist_dir});
-            const logger = utilsLib.getLogger(null);
+            const logger = utilsLib.getLogger(null, null, null);
```
g11tech, line 131: combine these two into one params struct that is passed or null

pkgs/utils/src/log.zig
```diff
@@ -4,7 +4,7 @@ const datetime = @import("datetime");
 
 // having activeLevel non comptime and dynamic allows us env based logging and even a keystroke activated one
 // on a running client, may be can be revised later
-pub fn compTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, comptime level: std.log.Level, comptime fmt: []const u8, args: anytype) void {
+pub fn compTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, comptime level: std.log.Level, comptime fmt: []const u8, args: anytype, file: ?std.fs.File) void {
```
g11tech, line 7: also add the fileActiveLevel in the last argument and change it to a params/struct so that stderr and log fiile can be run at different active log levels

for e.g. we might just want info logs on screen and full default logs in the file

pkgs/cli/src/main.zig
g11tech: also add another param into the last struct which is activeLogLevel for the file, call this struct FileParams 

also add a corresponding fileActiveLogLevel cli arg default set to debug, and change the logger's console active default level to info (also take it as a cli param with default info if thats not already there)

pkgs/utils/src/log.zig
```diff
@@ -31,86 +31,169 @@ pub fn compTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, comp
...
 const LoggerScope = enum {
     default,
     n1,
     n2,
     n3,
 };
 
+pub const FileParams = struct {
+    filePath: ?[]const u8,
+    fileName: ?[]const u8,
```
g11tech, line 69:
```suggestion
    filePath: []const u8,
    fileName: []const u8,
```
no need to make individual params optional

pkgs/utils/src/log.zig
```diff
@@ -31,86 +31,169 @@ pub fn compTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, comp
...
+};
+
 pub const ZeamLogger = struct {
     activeLevel: std.log.Level,
     scope: LoggerScope,
+    file: ?std.fs.File, // optional log file
+    filePath: ?[]const u8, // path to log file directory
+    fileName: []const u8, // log file name
```
g11tech, line 77: just store them as ?fileParams that gets passed in init

pkgs/utils/src/log.zig
```diff
@@ -4,8 +4,8 @@ const datetime = @import("datetime");
 
 // having activeLevel non comptime and dynamic allows us env based logging and even a keystroke activated one
 // on a running client, may be can be revised later
-pub fn compTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, comptime level: std.log.Level, comptime fmt: []const u8, args: anytype) void {
-    if (@intFromEnum(level) > @intFromEnum(activeLevel)) {
+pub fn compTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, fileActiveLevel: std.log.Level, comptime level: std.log.Level, comptime fmt: []const u8, args: anytype, file: ?std.fs.File) void {
```
g11tech, line 7: combine fileActiveLevel and file into one optional params struct only

and update the below early return condition to if (level > activelevel and (fileparams is null or level > fileparams.activelevl)

pkgs/utils/src/log.zig
```diff
@@ -31,86 +31,182 @@ pub fn compTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, comp
...
+    filePath: []const u8,
+    fileName: []const u8,
+};
+
 pub const ZeamLogger = struct {
     activeLevel: std.log.Level,
     scope: LoggerScope,
+    file: ?std.fs.File, // optional log file
+    fileParams: ?FileParams,
+    last_rotation_day: i64,
```
g11tech, line 80: combine them all into one holding optional fileparams

pkgs/utils/src/log.zig
```diff
@@ -4,8 +4,8 @@ const datetime = @import("datetime");
 
 // having activeLevel non comptime and dynamic allows us env based logging and even a keystroke activated one
 // on a running client, may be can be revised later
-pub fn compTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, comptime level: std.log.Level, comptime fmt: []const u8, args: anytype) void {
-    if (@intFromEnum(level) > @intFromEnum(activeLevel)) {
+pub fn compTimeLog(comptime scope: LoggerScope, activeLevel: std.log.Level, comptime level: std.log.Level, comptime fmt: []const u8, args: anytype, fileParams: ?FileParams) void {
```
g11tech, line 7: don't pass full fileParams but cast it down {fileActiveLevel, file} only so declare another struct and use that here,

gballet (approved):
LGTM but there are still come comments left by Gajinder that I feel should be addressed.

PR #136, opened 2025-09-04
feat: wireup the node for 3sf voting

g11tech:
just wires up the nodes and validators for 3sf voting and publishing

### DIFF
--- a/pkgs/network/src/interface.zig
+++ b/pkgs/network/src/interface.zig
@@ -54,10 +54,7 @@ pub const GossipMessage = union(GossipTopic) {
     const Self = @This();
     // figureout is there a generic way to find active enum
     pub fn getTopic(self: *const Self) GossipTopic {
-        switch (self.*) {
-            .block => return .block,
-            .vote => return .vote,
-        }
+        return std.meta.activeTag(self.*);
     }
 };
 
--- a/pkgs/network/src/interface.zig
+++ b/pkgs/network/src/interface.zig
@@ -110,7 +110,7 @@ pub const GenericGossipHandler = struct {
 
             const publishWrapper = try self.allocator.create(MessagePublishWrapper);
             publishWrapper.* = MessagePublishWrapper{ .handler = handler, .data = data, .networkId = self.networkId };
-            std.debug.print("\n\n\nnetwork-{d}:: schedueling ongossip publishWrapper={any} on loop for topic {any}\n", .{ self.networkId, topic, publishWrapper });
+            std.debug.print("\n\n\nnetwork-{d}:: schedueling ongossip publishWrapper={any} on loop for topic {any}\n\n", .{ self.networkId, topic, publishWrapper });
 
             // TODO: figure out why scheduling on the loop is not working for libp2p separate net instance
             // remove this option once resolved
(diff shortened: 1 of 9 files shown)


pkgs/network/src/interface.zig
```diff
@@ -107,7 +110,7 @@ pub const GenericGossipHandler = struct {
 
             const publishWrapper = try self.allocator.create(MessagePublishWrapper);
             publishWrapper.* = MessagePublishWrapper{ .handler = handler, .data = data, .networkId = self.networkId };
-            std.debug.print("\n\n\nnetwork-{d}:: schedueling ongossip publishWrapper={any} on loop for topic {any}\n", .{ self.networkId, topic, publishWrapper });
+            std.debug.print("\n\n\nnetwork-{d}:: schedueling ongossip publishWrapper={any} on loop for topic {any}\n\n", .{ self.networkId, topic, publishWrapper });
```
g11tech, line 113: @noopur23 add the following three tasks to your log EPIC task:

1. pass same logger to network which you pass to node (in case of mock network, create and pass its own logger with default scope) and change std.debug statements to the logger.debug ones
2. add logger to validator and pass node logger to validator as well and replace std.debug there as well
3. colore code the level (like [info] [debug]) into different colors, as well as timestamp and where we print the scope

rust/src/libp2p_bridge.rs
```diff
@@ -37,13 +37,21 @@ pub fn create_and_run_network(network_id: u32, zig_handler: u64, self_port: i32,
...
         // TODO: get the topic mapping from topic_id
-        let topic = gossipsub::IdentTopic::new("block");
+        let topic = match topic_id {
+            0 => gossipsub::IdentTopic::new("block"),
+            1 => gossipsub::IdentTopic::new("vote"),
+            unknown_id => {
+                println!("Invalid topic_id: {unknown_id}");
+                return;
+            },
+        };
```
gballet, line 53: this is more something that we will need to do during the cleanup than in this PR, but I suggest that we reduce our reliance on rust by doing this in zig and passing the topic/string as a parameter. Less surface to maintain in rust. But it's ok for now.

rust/src/libp2p_bridge.rs
```diff
@@ -137,11 +145,20 @@ pub async fn run_eventloop(&mut self) {
...
                     {
                         let topic = message.topic.as_str();
-                        let _topic_ptr = topic.as_ptr();
-                        let _topic_len = topic.len();
+                        let topic_id: u32 = match topic {
+                            "block" => 0,
+                            "vote" => 1,
+                            unknown_topic => {
+                                println!("\nrustbridge{0}:: unknown_topic={unknown_topic}\n", self.network_id);
+                                // Will return here return from event loop or return from the match poll to go
```
gballet, line 153: comment is perfectible

gballet (approved):

pkgs/node/src/chain.zig
```diff
@@ -128,19 +132,42 @@ pub const BeamChain = struct {
...
+    pub fn constructVote(self: *Self, opts: VoteConstructionParams) !types.Mini3SFVote {
+        const slot = opts.slot;
+        // this will be removed in followup PR to reflect latest spec changes
+        const validator_id = 0;
+
+        const head = self.forkChoice.get_proposal_head(slot);
+        const target = self.forkChoice.get_vote_target();
+
+        const vote = types.Mini3SFVote{
+            //
```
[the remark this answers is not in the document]
g11tech: just wanted to make it multiline

PR #144, opened 2025-09-08
extend forkchoice structures and accept new votes

g11tech:
get the forkchoice ready for accepting and applying new votes as per specs

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -135,7 +135,7 @@ pub const BeamChain = struct {
     pub fn constructVote(self: *Self, opts: VoteConstructionParams) !types.Mini3SFVote {
         const slot = opts.slot;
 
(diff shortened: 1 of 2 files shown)


pkgs/node/src/chain.zig
```diff
@@ -190,9 +190,12 @@ pub const BeamChain = struct {
         self.printSlot(self.forkChoice.fcStore.currentSlot);
     }
 
-    // import block assuming it is validated
+    // import block assuming it is gossip validated or synced
+    // this onBlock corresponds to spec's forkchoice's onblock with some functionality split between this and
+    // our implemented forkchoice's onblock. this is to parellize apply transition with other verifications
```
gballet, line 195:
```suggestion
    // our implemented forkchoice's onblock. this is to parallelize "apply transition" with other verifications
```

PR #151, opened 2025-09-09
activate flow to process new votes plus slight block production/publishing refactor

g11tech:


pkgs/node/src/chain.zig
```diff
@@ -149,6 +157,14 @@ pub const BeamChain = struct {
         return vote;
     }
 
+    // TODO: right now validator indepdently publishes to the network but move the gossip
```
g11tech, line 160: @noopur23 addressing this also in the same issue as the publishblock todo comment one

PR #163, opened 2025-09-11
move to pq devnet0 stf

g11tech:
@noopur23 can you get this PR & tests working

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -242,7 +242,12 @@ pub const BeamChain = struct {
             var cpost_state = try types.sszClone(self.allocator, types.BeamState, pre_state);
 
             // 2. apply STF to get post state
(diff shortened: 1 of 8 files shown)


pkgs/node/src/chain.zig
```diff
@@ -109,16 +109,16 @@ pub const BeamChain = struct {
         const pre_state = self.states.get(parent_root) orelse return BlockProductionError.MissingPreState;
         var post_state = try types.sszClone(self.allocator, types.BeamState, pre_state);
 
-        const timestamp = self.config.genesis.genesis_time + opts.slot * params.SECONDS_PER_SLOT;
+        // const timestamp = self.config.genesis.genesis_time + opts.slot * params.SECONDS_PER_SLOT;
```
g11tech, line 112:
```suggestion
        // keeping for later when execution will be integrated into lean
        // const timestamp = self.config.genesis.genesis_time + opts.slot * params.SECONDS_PER_SLOT;
```

Issue #139, opened 2025-09-07
add logger to validator and pass node logger to validator as well


noopur23:
@noopur23 add the following three tasks to your log EPIC task:

2. add logger to validator and pass node logger to validator as well and replace std.debug there as well

_Originally posted by @g11tech in https://github.com/blockblaz/zeam/pull/136#discussion_r2328760575_

PR #171, opened 2025-09-13
Add logger to validator, replace std.debug

bomanaps:
closes this https://github.com/blockblaz/zeam/issues/139


g11tech:
there are still some std.debug logs pending in mock and ethlibp2p, all logs to be converted

g11tech (approved):

PR #172, opened 2025-09-13
add color to console and file logger

### LINKED ISSUE #140: color code the level (like [info] [debug]) into different colors, as well as timestamp and where we print the scope
@noopur23 add the following three tasks to your log EPIC task:

3. colore code the level (like [info] [debug]) into different colors, as well as timestamp and where we print the scope

_Originally posted by @g11tech in https://github.com/blockblaz/zeam/pull/136#discussion_r2328760575_

noopur23:
- Add colors for console logging
- Add colors for file logging with option to turn it off from cli for non color supported editors
- Add testLogger tests - level set to debug and no file logging

closes #140


pkgs/utils/src/log.zig
```diff
@@ -104,21 +123,26 @@ pub const DEFAULT_FILE_BEHAVIOUR = FileBehaviourParams{
...
 
 pub const ZeamLogger = struct {
     activeLevel: std.log.Level,
     scope: LoggerScope,
     fileParams: ?FileParams,
+    testLogger: bool,
 
     const Self = @This();
-    pub fn init(scope: LoggerScope, activeLevel: std.log.Level, fileBehaviourParams: ?FileBehaviourParams) Self {
+    pub fn init(scope: LoggerScope, activeLevel: std.log.Level, fileBehaviourParams: ?FileBehaviourParams, testLogger: bool) Self {
```
g11tech, line 136: why is this extra flag testLogger needed, simply null should be passed as fileBehaviourParams in that case , which is just what getTestLogger would do

Issue #168, opened 2025-09-13
replace std.debug.print in mock and ethlibp2p network by logger


g11tech:
pass a logger during init of the network and use that to dump the logs including the the invalid ssz file's name that gets written by ethlibp2p network during parsing failure

PR #173, opened 2025-09-13
log invalid ssz dump file names and further logger cleanup

bomanaps:
fixes this https://github.com/blockblaz/zeam/issues/168


pkgs/cli/src/main.zig
```diff
@@ -212,7 +212,7 @@ pub fn main() !void {
                 network.* = try networks.Mock.init(allocator, loop, &logger1);
                 backend1 = network.getNetworkInterface();
                 backend2 = network.getNetworkInterface();
-                std.debug.print("---\n\n mock gossip {any}\n\n", .{backend1.gossip});
+                logger1.debug("---\n\n mock gossip {any}\n\n", .{backend1.gossip});
```
g11tech, line 215: remove extra \ns

PR #181, opened 2025-09-14
improve printSlot formatting

### LINKED ISSUE #103: reformat/pretty print the chain status in printSlot which is called at the start of the third interval (post attestation period)

anshalshukla:
Closes #103

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -179,11 +179,40 @@ pub const BeamChain = struct {
             return;
         };
 
-        self.logger.info("chain received on slot cb at slot={d} head={any} headslot={d}", .{
-            //
+        // Get additional chain information
+        const justified = self.forkChoice.fcStore.latest_justified;
+        const finalized = self.forkChoice.fcStore.latest_finalized;
+
+        // Calculate chain progress
+        const blocksBehind = if (slot > fcHead.slot) slot - fcHead.slot else 0;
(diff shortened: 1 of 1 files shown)


pkgs/node/src/chain.zig
```diff
@@ -179,11 +179,40 @@ pub const BeamChain = struct {
...
         };
 
-        self.logger.info("chain received on slot cb at slot={d} head={any} headslot={d}", .{
-            //
+        // Get additional chain information
+        const justified = self.forkChoice.fcStore.latest_justified;
+        const finalized = self.forkChoice.fcStore.latest_finalized;
+
+        // Calculate chain progress
+        const blocksBehind = if (slot > fcHead.slot) slot - fcHead.slot else 0;
```
GrapeBaBa, line 187: snake case variable is best practical follow the guide https://ziglang.org/documentation/0.14.1/#Style-Guide

g11tech:
can you snapshot for how it prints?

anshalshukla:
image

I've written a testcase for this, it's not a functional test, just an easy way to visualize the output. I can revert the commit if you think it's something we don't want to include in the codebase.

g11tech (approved):
approving even though we might improve the formatting to be a bit more succinct which is more of a UX problem

@noopur23 create a task : come up with a succinct representation for printing the chain status and assign it to me

PR #188, opened 2025-09-15
feat: add `override_genesis_time` flag in `node` cmd

GrapeBaBa:
This pull request introduces support for overriding the genesis time via CLI and configuration. At the same time fixs config and node tests not run bug.

### DIFF
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -341,3 +341,7 @@ pub fn main() !void {
         },
     }
 }
+
+test {
+    @import("std").testing.refAllDeclsRecursive(@This());
+}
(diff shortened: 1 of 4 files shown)


pkgs/cli/src/main.zig
```diff
@@ -335,9 +337,13 @@ pub fn main() !void {
             };
             defer start_options.deinit(allocator);
 
-            try node.loadGenesisConfig(allocator, custom_genesis, &start_options);
+            try node.loadGenesisConfig(allocator, custom_genesis, leancmd.override_genesis_time, &start_options);
```
g11tech, line 340: make this a boolean flag and if true pass time.now here to laodgenesisconfig else null

GrapeBaBa: That will lead to different genesis time for different nodes, I think it doesn't make sense.

g11tech: hmm make sense that way

g11tech (approved):

Issue #191, opened 2025-09-16
for printSlot debug log forkchoice in a visual way (upto a certain depth)


g11tech:
sort of a tree visualization:

0x23 (1) // has two child branches
   -> 0x52 (2) -> 0x34(3) -> (4) -> 0x67(5) // a chain from slot 2 to slot 5, with slot 4 missing, further has two branches
                                                            - (6) -> 0x77(7)
                                                            - (6) -> (7) -> 0x45(8)
   -> (2..5) -> 0x88(6) (a child chain of 0x23(1) with slot 2,3,4,5 missing and 0x88 at the child block at 6th slot)

  the github is reformatting it above (check the edit version) but this is how it should look like:

<img width="1056" height="260" alt="Image" src="https://github.com/user-attachments/assets/4dd53a5c-06db-447c-b099-62b4bb87b8ca" />

markolazic01:
looks like this was addressed in PR #426

PR #196, opened 2025-09-16
visualize forkchoice tree

ScottyPoi:
adds a visualization for the fork choice tree to the printSlot logging function as described in issue: #191. 

The visualization is configurable with tree_depth limit.  A tree_depth limit of `null` equals no depth limit.

A visualization with tree_depth = 2 should look something like:

```
0x1234 (0) // has 2 child branches
    ├── 0x2345 (1) // has 1 child branch
    │   └── 0x3456 (2) // ... (truncated at depth 2)
    └── [slots 3..4] ─┘ 0x5678 (5)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,7 +251,7 @@ jobs:
       run: zig build spectest:generate --summary all
 
     - name: Run all spec tests
-      run: zig build spectest:run --summary all -- --skip-expected-error-fixtures=true
+      run: zig build spectest:run --summary all
 
   dummy-prove:
     name: Dummy prove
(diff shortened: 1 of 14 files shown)


g11tech:
looks cool! some changes to make it look small and easy for eyes:
1. remove // comments, we can have a readme section to explain how to understand it
2. remove slots just [4..5] is sufficient, if only one slot is missing, just use [4] 
3. remove spacing before brackets () and [] both,
4. remove redundant 0x
5. use just two digits to represent root

g11tech:
@ScottyPoi can we truncate in opposite direction because most of the time relevance is to see the leaf forking, and if there are too many leaves include 3 by heaviest weight, and include 3 by most recent, so total 6 but keep this 3 number as a param

also relevant to know is justified is a parent in which of these branches

anshalshukla:
closing this in favor of #426

PR #197, opened 2025-09-17
add: module specific child loggers

### LINKED ISSUE #185: child logger scope addition for tagging the module printing log
so adding additional tag to each of the modules like [network], [chain], [forkchoice], [validator]

if one needs to zero in into a certain module using grep on console or log file, its not possible because we don't add tags to the log prefix that we build in the log file.

so we need to define another module enum, which one can create child of the top logger and any printing of the log through the child logger also prints out the additional module tag in the prefix

anshalshukla:
Closes #185

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -31,8 +31,10 @@ pub const BeamChain = struct {
     // from finalized onwards to recent
     states: std.AutoHashMap(types.Root, types.BeamState),
     nodeId: u32,
-    logger: *zeam_utils.ZeamLogger,
-    child_logger: zeam_utils.ChildLogger,
+    // This struct needs to contain the zeam_logger_config to be able to call `maybeRotate`
+    // For all other modules, we just need module_logger
+    zeam_logger_config: *zeam_utils.ZeamLoggerConfig,
+    module_logger: zeam_utils.ModuleLogger,
     registered_validator_ids: []usize = &[_]usize{},
 
     const Self = @This();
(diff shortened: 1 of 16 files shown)


anshalshukla:
image

pkgs/node/src/chain.zig
```diff
@@ -68,6 +68,9 @@ pub const BeamChain = struct {
         // forkchoice head
         const slot = @divFloor(time_intervals, constants.INTERVALS_PER_SLOT);
         const interval = time_intervals % constants.INTERVALS_PER_SLOT;
+        const zeam_logger = self.logger;
+        const logger = zeam_logger.child(.node);
```
g11tech, line 72: this logger will have scope of "chain"

pkgs/utils/src/log.zig
```diff
@@ -258,6 +291,87 @@ pub const ZeamLogger = struct {
...
+    pub fn child(self: *const Self, moduleTag: ModuleTag) ChildLogger {
+        return ChildLogger{
+            .parent = self,
+            .moduleTag = moduleTag,
+        };
+    }
+};
+
+/// Child logger that adds module-specific tags to log messages
+pub const ChildLogger = struct {
```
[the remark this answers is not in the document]
g11tech: hmmm, rename child logger as ModuleLogger, and why are you passing both logger and child logger at many places

pkgs/utils/src/log.zig
```diff
@@ -347,3 +461,24 @@ fn getLevelColor(comptime level: std.log.Level) []const u8 {
...
     };
 }
+
+fn getModuleTagName(tag: ModuleTag) []const u8 {
+    return switch (tag) {
+        .cli => "cli",
+        .chain => "chain",
+        .configs => "configs",
+        .forkchoice => "forkchoice",
+        .gossip_handler => "gossip_handler",
```
g11tech, line 471: just gossip is good enough
```suggestion
        .gossip_handler => "gossip",
```

pkgs/utils/src/log.zig
```diff
@@ -347,3 +461,24 @@ fn getLevelColor(comptime level: std.log.Level) []const u8 {
...
+        .metrics => "metrics",
+        .network => "network",
+        .node => "node",
+        .params => "params",
+        .state_proving_manager => "state_proving_manager",
+        .state_transition => "stf",
+        .state_transition_runtime => "stf_runtime",
```
g11tech, line 478:
```suggestion
        .state_transition_runtime => "runtime",
```

anshalshukla:
> hmmm, rename child logger as ModuleLogger, and why are you passing both logger and child logger at many places

I've changed the structure quite a bit, have changed ZeamLogger to ZeamLoggerConfig which stores the configuration, ModuleLogger stores optional module tag and logging is performed via it.

g11tech (dismissed):
lgtm :heart:

PR #198, opened 2025-09-17
fix: make start node test friendly and fix several memory leak

GrapeBaBa:
This PR makes starting node more testing friendly. Such as #194 it need access `BeamNode` to check finalize, but right now can't do it with `startNode` function in the `node.zig`. At the same time the refactor seems make `Clock`, `BeamNode` lifecycle more easy. Also try to fix some potential memory leak issue.

### DIFF
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -343,7 +343,7 @@ pub fn main() !void {
 
             var lean_node: node.Node = undefined;
 
-            try node.Node.init(&lean_node, allocator, &start_options);
+            try lean_node.init(allocator, &start_options);
             defer lean_node.deinit();
             try lean_node.run();
         },
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -326,7 +326,7 @@ pub fn main() !void {
         .node => |leancmd| {
             var logger = utils_lib.getLogger(console_log_level, utils_lib.FileBehaviourParams{ .fileActiveLevel = log_file_active_level, .filePath = log_filepath, .fileName = log_filename });
 
-            var start_options: node.StartNodeOptions = .{
+            var start_options: node.NodeOptions = .{
                 .node_id = leancmd.node_id,
                 .metrics_enable = leancmd.metrics_enable,
                 .metrics_port = leancmd.metrics_port,
@@ -341,7 +341,11 @@ pub fn main() !void {
 
             try node.buildStartOptions(allocator, leancmd, &start_options);
 
-            try node.startNode(allocator, &start_options);
+            var lean_node: node.Node = undefined;
+
+            try node.Node.init(&lean_node, allocator, &start_options);
+            defer lean_node.deinit();
+            try lean_node.run();
         },
     }
 }
--- a/pkgs/cli/src/node.zig
+++ b/pkgs/cli/src/node.zig
@@ -23,7 +23,7 @@ const NodeCommand = @import("main.zig").NodeCommand;
 
 const prefix = "zeam_";
 
-pub const StartNodeOptions = struct {
+pub const NodeOptions = struct {
     node_id: u32,
     bootnodes: []const []const u8,
     validator_indices: []usize,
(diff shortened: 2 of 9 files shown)


pkgs/types/src/lib.zig
```diff
@@ -122,6 +122,13 @@ pub const BeamState = struct {
     // a flat representation of the justifications map
     justifications_roots: []Root,
     justifications_validators: []u8,
+
+    pub fn deinit(self: *BeamState, allocator: Allocator) void {
+        // historical_block_hashes and justified_slots are slices so need to be freed
+        // justifications_roots and justifications_validators not freed for now as they are not allocated
+        allocator.free(self.historical_block_hashes);
+        allocator.free(self.justified_slots);
```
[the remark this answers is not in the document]
GrapeBaBa: And I remembered the types ser/dser using ssz lib can't have allocator field, since the ssz lib will reflect all the fields in the types.

pkgs/cli/src/node.zig
```diff
@@ -33,19 +33,159 @@ pub const StartNodeOptions = struct {
...
+    network: networks.EthLibp2p,
+    beam_node: BeamNode,
+    clock: Clock,
+    enr: ENR,
+    options: *const NodeOptions,
+    allocator: std.mem.Allocator,
+
+    const Self = @This();
+
+    pub fn init(self: *Self, allocator: std.mem.Allocator, options: *const NodeOptions) !void {
```
g11tech, line 57: this is a weird form of init where one passes the reference to a self which is unitialized

need @gballet 's opinion on this

pkgs/cli/src/node.zig
```diff
@@ -33,19 +33,159 @@ pub const StartNodeOptions = struct {
...
+        var anchorState = try sft.genGenesisState(allocator, chain_config.genesis);
+        errdefer anchorState.deinit(allocator);
+
+        // TODO we seem to be needing one loop because then the events added to loop are not being fired
+        // in the order to which they have been added even with the an appropriate delay added
+        // behavior of this further needs to be investigated but for now we will share the same loop
+        self.loop = try xev.Loop.init(.{});
+
+        const self_node_index = options.validator_indices[0];
+        try ENR.decodeTxtInto(&self.enr, options.bootnodes[self_node_index]);
```
g11tech, line 90: abstract away all the network related processing to another function

pkgs/types/src/lib.zig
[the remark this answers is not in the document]
GrapeBaBa: It may need more consider as not all the slices allocated such as `justifications_roots` and `justifications_validators `

pkgs/cli/src/main.zig
```diff
@@ -341,7 +341,11 @@ pub fn main() !void {
 
             try node.buildStartOptions(allocator, leancmd, &start_options);
 
-            try node.startNode(allocator, &start_options);
+            var lean_node: node.Node = undefined;
+
+            try node.Node.init(&lean_node, allocator, &start_options);
```
g11tech, line 346:
```suggestion
            try lean_node.init(allocator, &start_options);
```
this should work

pkgs/cli/src/node.zig
g11tech: ok, then we need to move a whole lot of inits to this form

pkgs/cli/src/node.zig
```diff
@@ -33,19 +33,159 @@ pub const StartNodeOptions = struct {
...
+        self.loop.deinit();
+    }
+
+    pub fn run(self: *Node) !void {
+        try self.network.run();
+        try self.beam_node.run();
+
+        const ascii_art =
+            \\
+            \\  ███████╗███████╗ █████╗ ███╗   ███╗
```
g11tech, line 156: cool art :heart:

pkgs/cli/src/node.zig
g11tech: you need to address this @GrapeBaBa then we can probably merge in the PR

pkgs/cli/src/node.zig
spiral-ladder: think it also depends on how large the struct is as the link mentions, if the struct isn't large in size then this form is not necessary.

PR #208, opened 2025-09-18
fix: flattening order

### LINKED ISSUE #201: sort the justification roots in asc order before re-flattening the justifications
right now we iterate over keys and flatten the justifications map but as per spec the keys (justifications_root have to be in asc order)

https://github.com/blockblaz/zeam/blob/main/pkgs/state-transition/src/transition.zig#L270:L276

Image

so fix the flattening order by:
1. first getting all keys in justifications_roots and then sorting them
2. iterate over roots in order and flatten to justifications_validators

anshalshukla:
Closes #201


anshalshukla:
Not sure, if there is a good way to add tests for this as specs doesn't contain any functional tests

Issue #99, opened 2025-08-24
create a params struct to pass in Chain and put config, anchorstate nodeid, logger into it and pass via it, may be anchorstate also needs to be pointer,


noopur23:
create a params struct that you pass here and put config, anchorstate nodeid, logger into it and pass via it, may be anchorstate also needs to be pointer,

on second thoughts, do it as a followup PR, not only for this but for other structs as well, may be one PR for each such struct

_Originally posted by @g11tech in https://github.com/blockblaz/zeam/pull/64#discussion_r2290280944_

PR #209, opened 2025-09-18
Refactor Chain init params

noopur23:
part of #99


pkgs/node/src/chain.zig
```diff
@@ -392,11 +396,11 @@ test "printSlot output demonstration" {
     // Create a mock chain with some blocks
     const mock_chain = try stf.genMockChain(allocator, 3, chain_config.genesis);
     const beam_state = mock_chain.genesis_state;
-    const nodeid = 42; // Test node ID
+    const nodeId = 42; // Test node ID
```
GrapeBaBa, line 399: snake case for variable

pkgs/node/src/node.zig
```diff
@@ -20,7 +20,7 @@ const LevelDB = struct {};
 
 const NodeOpts = struct {
     config: configs.ChainConfig,
-    anchorState: types.BeamState,
+    anchorState: *const types.BeamState,
```
GrapeBaBa, line 23: snake case for field

PR #211, opened 2025-09-18
increase the input size for risc0 proving

g11tech:
fix the risc0 run aparenly broken because of invalid deserilization of input


pkgs/types/src/lib.zig
```diff
@@ -12,7 +12,7 @@ pub const ValidatorIndex = u64;
 pub const Bytes48 = [48]u8;
 
 //update signature size to 4000 after ssz is fixed
-pub const SIGSIZE = 40;
+pub const SIGSIZE = 4000;
```
g11tech, line 15: fyi: size was reverted to 40 in main to test the risc0 with other stf changes that are going in

Issue #215, opened 2025-09-19
Add option for configurable preset using flags


chetanyb:
- create various preset for running the node(eg minimal, devnet, mainnet etc)
- allow to select and use anyone with flags
- set default in case no flag is given

PR #231, opened 2025-09-23
Add configurable preset using flags

chetanyb:
Closes https://github.com/blockblaz/zeam/issues/215

### Usage

```bash
zig build -Dpreset=mainnet
```

```bash
zig build -Dpreset=minimal
```


anshalshukla:
Based on discussion with @chetanyb I'm closing this as it is not something we require right now and the PR in current state is not mergable and would require equal effort to resolve conflicts as much to create a new one.

PR #241, opened 2025-09-27
feat: init spectest module

GrapeBaBa:


pkgs/types/src/lib.zig
```diff
@@ -121,6 +121,49 @@ pub const BeamState = struct {
     justifications_roots: JustificationsRoots,
     justifications_validators: JustificationsValidators,
 
+    pub fn withJustifications(self: *BeamState, allocator: Allocator, justifications: *const std.AutoHashMapUnmanaged(Root, []u8)) !void {
```
g11tech, line 124: also move loadJustiffications here as well

PR #243, opened 2025-09-28
fix: store states as ptr

### LINKED ISSUE #234: move self.states cache to pointer of states along with blockInfo's poststate
move self.states cache to pointer of states along with blockInfo's poststate

https://github.com/blockblaz/zeam/pull/223/files#diff-fad5783414afdcea3ec80abc66e5cb701cc2487b72c4f52ecb8181df4e71c296R307

anshalshukla:
Closes #234

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ prometheus.yml
 # Data files (RocksDB)
 data/
 .DS_Store
+
+# VSCode files
+.vscode/
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -264,6 +264,23 @@ pub fn main() !void {
 
             var backend1: networks.NetworkInterface = undefined;
             var backend2: networks.NetworkInterface = undefined;
+
+            // These are owned by the network implementations and will be freed in their deinit functions
+            // We will run network1 and network2 after the nodes are running to avoid race conditions
+            var network1: *networks.EthLibp2p = undefined;
+            var network2: *networks.EthLibp2p = undefined;
+            var listen_addresses1: []Multiaddr = undefined;
+            var listen_addresses2: []Multiaddr = undefined;
+            var connect_peers: []Multiaddr = undefined;
+            defer {
+                for (listen_addresses1) |addr| addr.deinit();
+                allocator.free(listen_addresses1);
+                for (listen_addresses2) |addr| addr.deinit();
+                allocator.free(listen_addresses2);
+                for (connect_peers) |addr| addr.deinit();
+                allocator.free(connect_peers);
+            }
+
             if (mock_network) {
                 var network: *networks.Mock = try allocator.create(networks.Mock);
                 network.* = try networks.Mock.init(allocator, loop, logger1_config.logger(.network));
@@ -271,12 +288,10 @@ pub fn main() !void {
                 backend2 = network.getNetworkInterface();
                 logger1_config.logger(null).debug("--- mock gossip {any}", .{backend1.gossip});
             } else {
-                var network1: *networks.EthLibp2p = try allocator.create(networks.EthLibp2p);
+                network1 = try allocator.create(networks.EthLibp2p);
                 const key_pair1 = enr_lib.KeyPair.generate();
                 const priv_key1 = key_pair1.v4.toString();
-                const listen_addresses1 = &[_]Multiaddr{try Multiaddr.fromString(allocator, "/ip4/0.0.0.0/tcp/9001")};
-                // these addresses are converted to a slice in the `run` function of `EthLibp2p` so it can be freed safely after `run` returns
-                defer for (listen_addresses1) |addr| addr.deinit();
+                listen_addresses1 = try allocator.dupe(Multiaddr, &[_]Multiaddr{try Multiaddr.fromString(allocator, "/ip4/0.0.0.0/tcp/9001")});
                 const network_name1 = try allocator.dupe(u8, chain_config.spec.name);
                 errdefer allocator.free(network_name1);
                 network1.* = try networks.EthLibp2p.init(allocator, loop, .{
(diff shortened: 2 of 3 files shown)


pkgs/node/src/chain.zig
```diff
@@ -172,7 +175,9 @@ pub const BeamChain = struct {
         // 3. cache state to save recompute while adding the block on publish
         var block_root: [32]u8 = undefined;
         try ssz.hashTreeRoot(types.BeamBlock, block, &block_root, self.allocator);
-        try self.states.put(block_root, post_state);
+        const post_state_ptr = try self.allocator.create(types.BeamState);
+        post_state_ptr.* = post_state;
```
g11tech, line 179: post_state should already be an allocated state which you cloned in sszClone to begin with, so you don't need to re-alloc it

pkgs/node/src/chain.zig
```diff
@@ -308,10 +311,10 @@ pub const BeamChain = struct {
             block.slot,
         });
 
-        const post_state = blockInfo.postState orelse computedstate: {
+        const post_state = if (blockInfo.postState) |post_state_ptr| post_state_ptr.* else computedstate: {
             // 1. get parent state
-            const pre_state = self.states.get(signedBlock.message.parent_root) orelse return BlockProcessingError.MissingPreState;
-            var cpost_state = try types.sszClone(self.allocator, types.BeamState, pre_state);
+            const pre_state_ptr = self.states.get(signedBlock.message.parent_root) orelse return BlockProcessingError.MissingPreState;
+            var cpost_state = try types.sszClone(self.allocator, types.BeamState, pre_state_ptr.*);
```
g11tech, line 317: if you do the ssz change and make cpost_state as allocated pointer then post_state can directly be state pointer

pkgs/node/src/chain.zig
```diff
@@ -47,10 +47,12 @@ pub const ProducedBlock = struct {
 
 pub const BeamChain = struct {
     config: configs.ChainConfig,
+    anchor_state: types.BeamState,
+
     forkChoice: fcFactory.ForkChoice,
     allocator: Allocator,
     // from finalized onwards to recent
-    states: std.AutoHashMap(types.Root, types.BeamState),
+    states: std.AutoHashMap(types.Root, *types.BeamState),
```
GrapeBaBa, line 55: what is the purpose of this change, it seems add an extra pointer lookup

GrapeBaBa:
> I think the key advantage is that when we put `*BeamState` vs `BeamState` is that the structs internal data is shallow copied in case of `BeamState` whereas it's just the pointer that get's copied with `*BeamState`, so its more computational and memory efficient.

can `getPtr` function of `map` help? I thought making the `states` map as the owner of all the`BeamState`, it still can get the any `BeamState` pointer anywhere.

pkgs/node/src/chain.zig
```diff
@@ -94,11 +99,13 @@ pub const BeamChain = struct {
     pub fn deinit(self: *Self) void {
         var it = self.states.iterator();
         while (it.next()) |entry| {
-            entry.value_ptr.deinit(self.allocator);
+            entry.value_ptr.*.deinit(self.allocator);
```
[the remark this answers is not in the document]
anshalshukla: deinit already takes a pointer argument to Self, but since i've modified the value to`*BeamState`, the entry give me a pointer to value `**BeamState`, a cleaner way can be get the dereference value in the closure itself `|*entry|`.  I don't think it can be null as the value is kept as a pointer and not a nullable pointer so it should be safe to dereference it.

GrapeBaBa:
> `getPtr` can help get the pointer to the value present in the map but I don't think it's possible to shift ownership of the `BeamState` variable to map, I think it's not a good idea as well because then it will become significantly harder to manage that piece of memory.
> 
> I looked into the std library code and it internally calls
> 
> ```
>         pub fn putContext(self: *Self, allocator: Allocator, key: K, value: V, ctx: Context) Allocator.Error!void {
>             const result = try self.getOrPutContext(allocator, key, ctx);
>             result.value_ptr.* = value;
>         }
> ```
> 
> So the result will basically get the pointer to store value based on the presence of the key and then the value is stored, in case of `*BeamState` it would be just one pointer as compared to `BeamState` where it will be multiple pointer and copies of simple data types, plus maybe some additional padding.

I didn't get it. It indeed transfer the ownership to the map, the `BeamState` will in the memory of the `map`. Why is it harder to manage the that piece of memory? For store `*BeamState`, you allocated a memory for `BeamState` first and store the `*BeamState` in the map, it need extra `*BeamState` `destroy`.

GrapeBaBa:
> hmm, I was thinking in more rust terms there. But i think the crux is shallow copy vs pointer copy.

If it is the only concern, I felt it still can get pointer copy if needed even no copy in current approach and do some refactor in other code which using such as `fetchPut`, `getOrPut`. But I am not sure since I didn't dive into the whole code.

pkgs/node/src/chain.zig
```diff
@@ -29,14 +29,14 @@ pub const VoteConstructionParams = struct {
 
 pub const ChainOpts = struct {
     config: configs.ChainConfig,
-    anchorState: *const types.BeamState,
+    anchorState: types.BeamState,
```
g11tech, line 32: why are we not passing pointer here?

pkgs/node/src/chain.zig
```diff
@@ -47,10 +47,12 @@ pub const ProducedBlock = struct {
 
 pub const BeamChain = struct {
     config: configs.ChainConfig,
+    anchor_state: *types.BeamState,
+
     forkChoice: fcFactory.ForkChoice,
     allocator: Allocator,
     // from finalized onwards to recent
-    states: std.AutoHashMap(types.Root, types.BeamState),
+    states: std.AutoHashMap(types.Root, *types.BeamState),
```
g11tech, line 55: can we also make it a *const types.BeamState map?

anshalshukla: This should not be possible as the map is the owner here, if I make it `*const types.BeamState` I won't be able to free the memory as `deinit()` cannot be called on `const`

pkgs/node/src/chain.zig
```diff
@@ -29,14 +29,14 @@ pub const VoteConstructionParams = struct {
 
 pub const ChainOpts = struct {
     config: configs.ChainConfig,
-    anchorState: *const types.BeamState,
```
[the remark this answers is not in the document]
anshalshukla: I've made `struct BeamChain` owner of the ChainOpts's anchorState so that deinit can be performed, I think I had introduced const keyword in one of my earlier PR's and it went through because we weren't de-allocating it

PR #246, opened 2025-09-29
Fix RocksDB double free issue

bomanaps:
Prevents use-after-free during DB lifetime and remove the shutdown crash when multiple DB instances are opened (e.g., two nodes with separate --db_path).


anshalshukla:
@GrapeBaBa @g11tech @g11tech I agree with Kai, I think it's better to free cfs as soon as the open function ends and not wait for db to `deinit` as the `cfs` handles are only associated params we need for all the db operations, which we've already stored in `cf_handles`

Issue #248, opened 2025-09-29
Impl spectest for `process_header`


GrapeBaBa:
https://github.com/leanEthereum/leanSpec/blob/main/tests/lean_spec/subspecs/containers/test_state.py#L530-L607 , try to migrate `process_block_header` code from `transition.zig` to `BeamState.zig`

PR #256, opened 2025-09-30
feat: impl spectest for `genesis_state`

GrapeBaBa:
This pull request refactors the genesis state initialization and deinitialization logic for the blockchain state and block types, and adds comprehensive roundtrip and genesis generation tests for state justifications. The changes improve memory management, enforce consistent initialization patterns, and enhance test coverage and reliability for state transitions.

Fix #255

PR #257, opened 2025-09-30
refactor: move functions closer to type declarations

anshalshukla:
- Sort of prereq for #248
- Moves functions closer to type declarations
- Cleans type package and splits it into individual modules

### DIFF
--- a/pkgs/types/src/block.zig
+++ b/pkgs/types/src/block.zig
@@ -101,10 +101,29 @@ pub const BeamBlock = struct {
             .slot = self.slot,
             .proposer_index = self.proposer_index,
             .parent_root = self.parent_root,
-            .state_root = utils.ZERO_HASH,
+            .state_root = self.state_root,
             .body_root = body_root,
         };
         return header;
+    }
+
+    // computing latest block header to be assigned to the state for processing the block
+    pub fn blockToLatestBlockHeader(self: *const Self, allocator: Allocator, header: *BeamBlockHeader) !void {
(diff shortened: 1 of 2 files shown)


pkgs/cli/src/main.zig
```diff
@@ -235,7 +235,7 @@ pub fn main() !void {
             chain_options.num_validators = num_validators;
             // transfer ownership of the chain_options to ChainConfig
             const chain_config = try ChainConfig.init(Chain.custom, chain_options);
-            const anchorState = try sft_factory.genGenesisState(gpa.allocator(), chain_config.genesis);
+            const anchorState = try types.genGenesisState(gpa.allocator(), chain_config.genesis);
```
GrapeBaBa, line 238: I did similar thing in the #256 , and make the function in the `BeamState` struct, we may follow the style like leanSpec which make more util functions into the struct.

anshalshukla: @GrapeBaBa I've followed slightly different pattern with `genGenesisBlock` and `genGenesisState`, I've avoided self pointer in these fucntions to avoid first assigning the variable to undefined and then calling the function.

GrapeBaBa: I also felt it is wired before until I looked at this https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#cache-invalidation, it is actual recommended approach for large struct initialization. Currently i assume the `BeamState` and `BeamBlock` large struct.

Issue #250, opened 2025-09-29
Impl spectest for `state transition full`


GrapeBaBa:
https://github.com/leanEthereum/leanSpec/blob/main/tests/lean_spec/subspecs/containers/test_state.py#L675-L725

PR #259, opened 2025-09-30
Implement state-transition-full spectest

ScottyPoi:
Fix: #250


pkgs/spectest/src/containers/state.zig
```diff
@@ -4,6 +4,8 @@ const configs = @import("@zeam/configs");
 const types = @import("@zeam/types");
 const ssz = @import("ssz");
 const params = @import("@zeam/params");
+const state_transition = @import("@zeam/state-transition");
```
GrapeBaBa, line 7: Can we follow the leanSpec style to migrate utils function into `BeamState` member functions like previous test. It may wait other related the spec test merged.

ScottyPoi: you're referring to `process_slot` and other functions currently imported from `@zeam/state_transition`?

Yeah, I can do that

gballet (changes requested):
I don't think it's the right way to proceed (same thing with #260): there are these spec tests that will be output, and so there should be a `spectest` executable that ingests said spec tests and runs them. I'm not sure what the state of the leanSpec branch is, I've started an inquiry into that, but imo it's better to write a generic executable than rewriting all tests twice - in case there are some issues there. Maybe I'm projecting too much what happens with the execution clients, but it seems like the right way to proceed.

PR #275, opened 2025-10-03
feat: node connect/disconnect event handling

### LINKED ISSUE #267: maintain connected peers
as the peer connect/disconnect the events should be passed to the zig, where map of connected peers by peerid should be maintained in node/network.zig

and this should be displayed in the printSlot status

then in a followup PR, onInterval of chain/valirator should only be called if >=1 peer connected unless `--forceSynced` parameter which will always keep onInterval running

anshalshukla:
Closes #267

### DIFF
--- a/pkgs/network/src/ethlibp2p.zig
+++ b/pkgs/network/src/ethlibp2p.zig
@@ -105,6 +105,24 @@ export fn handleMsgFromRustBridge(zigHandler: *EthLibp2p, topic_str: [*:0]const 
     };
 }
 
+export fn handlePeerConnectedFromRustBridge(zigHandler: *EthLibp2p, peer_id: [*:0]const u8) void {
+    const peer_id_slice = std.mem.span(peer_id);
+    zigHandler.logger.info("network-{d}:: Peer connected: {s}", .{ zigHandler.params.networkId, peer_id_slice });
+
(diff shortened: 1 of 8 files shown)


pkgs/network/src/ethlibp2p.zig
```diff
@@ -254,17 +283,29 @@ pub const EthLibp2p = struct {
         _ = data;
     }
 
+    pub fn subscribePeerEvents(ptr: *anyopaque, handler: interface.OnPeerEventCbHandler) anyerror!void {
```
GrapeBaBa, line 286: nit: for the interface implementation, it may not need `anyerror`

pkgs/node/src/node.zig
```diff
@@ -37,11 +42,16 @@ pub const BeamNode = struct {
     validator: ?validators.BeamValidator = null,
     nodeId: u32,
     logger: zeam_utils.ModuleLogger,
+    connected_peers: *std.StringHashMap(PeerInfo),
```
GrapeBaBa, line 45: Is this `connected_peers` managed in the network module such as more natural?

GrapeBaBa:
LGTM

g11tech (approved):
taking in this PR so as to have a better log experience but can you fix the `beam` command issue in the followup PR @anshalshukla

Issue #277, opened 2025-10-04
metrics - lean_head_slot


noopur23:
Name -	lean_head_slot
Type - Gauge
Usage - Latest slot of the lean chain
Sample collection event - On get fork choice head

PR #293, opened 2025-10-04
Meterics for lean_head_slot

bomanaps:
Closes this https://github.com/blockblaz/zeam/issues/277
lean_head_slot
Screenshot 2025-10-04 at 10 53 41

increase(lean_head_slot[5m])
Screenshot 2025-10-04 at 10 56 38

3. delta(lean_head_slot[1m])
Screenshot 2025-10-04 at 10 57 22

### DIFF
--- a/pkgs/api/src/lib.zig
+++ b/pkgs/api/src/lib.zig
@@ -34,9 +34,11 @@ var g_initialized: bool = false;
 const Metrics = struct {
     chain_onblock_duration_seconds: ChainHistogram,
     block_processing_duration_seconds: BlockProcessingHistogram,
+    lean_head_slot: LeanHeadSlotGauge,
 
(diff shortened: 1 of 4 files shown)


bomanaps:
@gballet @g11tech

bomanaps:
@gballet

pkgs/api/src/lib.zig
```diff
@@ -121,6 +124,13 @@ pub const routes = @import("./routes.zig");
 pub const events = @import("./events.zig");
 pub const event_broadcaster = @import("./event_broadcaster.zig");
 
+/// Sets the lean head slot metric.
+/// This should be called whenever the fork choice head is updated.
+pub fn setLeanHeadSlot(slot: u64) void {
+    if (!g_initialized or isZKVM()) return;
```
g11tech, line 130: if its not initialized, all the metrics calls are already no op right?

bomanaps: Yes

g11tech: then lets remove this if, please comment and note this particular bit on the top of the metrics file, also remove noop metric since that isn't needed anymore since not initializing results into a no-op

Issue #307, opened 2025-10-09
State Transition Metrics


noopur23:
State Transition Metrics

- [ ] #282 
- [ ] #283 
- [ ] #284 
- [ ] #285 
- [ ] #286 
- [ ] #287 
- [ ] #288 
- [ ] #289

PR #310, opened 2025-10-10
impl state transition metrics

bomanaps:
### 1. `lean_latest_justified_slot` (Gauge)
Added a gauge to track the latest **justified checkpoint slot** from the state.  
After each state transition in `chain.zig`, we read `post_state.latest_justified.slot` and call  
`api.setLeanLatestJustifiedSlot()` to update the metric.

### 2. `lean_latest_finalized_slot` (Gauge)
Added a gauge to track the latest **finalized checkpoint slot** from the state.  
After each state transition in `chain.zig`, we read `post_state.latest_finalized.slot` and call  
`api.setLeanLatestFinalizedSlot()` to update the metric.

### 3. `lean_state_transition_time_seconds` (Histogram)
Created a timer to measure the **total duration of state transitions** from start to finish.  
We start a timer before calling `stf.apply_raw_block()` or `stf.apply_transition()` in `chain.zig`,  
then observe the duration after completion.

### 4. `lean_state_transition_slots_processed_total` (Counter)
Added a counter to track the **total number of slots processed**, including empty slots.  
After state transition completes, we calculate `post_state.slot - parent_slot` and call  
`api.addSlotsProcessed()` to increment the counter.

### 5. `lean_state_transition_slots_processing_time_seconds` (Histogram)
Implemented a **callback pattern** to time the `process_slots()` function execution.  
We added `recordSlotsProcessingTime` callback to `StateTransitionOpts`, which  
`process_slots()` uses to report its duration to `api.observeSlotsProcessingTime()`.

### 6. `lean_state_transition_block_processing_time_seconds` (Histogram)
Implemented a **callback pattern** to time the `process_block()` function execution.  
We added `recordBlockProcessingTime` callback to `StateTransitionOpts`, which  
`process_block()` uses to report its duration to `api.observeBlockProcessingTime()`.

### 7. `lean_state_transition_attestations_processed_total` (Counter)
Added a counter to track the **total number of attestations processed** across all blocks.  
After state transition in `chain.zig`, we count `block.body.attestations.constSlice().len` and call  
`api.addAttestationsProcessed()` to increment the counter.

### 8. `lean_state_transition_attestations_processing_time_seconds` (Histogram)
Implemented a **callback pattern** to time the `process_attestations()` function execution.  
We added `recordAttestationsProcessingTime` callback to `StateTransitionOpts`, passed it through  
`process_operations()`, and `process_attestations()` uses it to report its duration to  
`api.observeAttestationsProcessingTime()`.

- **Metrics # 1–4:** Use **chain layer calculation** (read/calculate after STF).  
- **Metrics # 5–8:** Use **callback pattern** (STF calls back to API with timing data).

Closes https://github.com/blockblaz/zeam/issues/307

### DIFF
--- a/pkgs/types/src/state.zig
+++ b/pkgs/types/src/state.zig
@@ -208,9 +208,8 @@ pub const BeamState = struct {
             self.slot += 1;
         }
 
-        const slots_processed: u64 = @intCast(slot - start_slot);
-
         if (comptime !zeam_metrics.isZKVM()) {
+            const slots_processed: u64 = @intCast(slot - start_slot);
             zeam_metrics.metrics.lean_state_transition_slots_processed_total.incrBy(slots_processed);
         }
     }
@@ -285,9 +284,9 @@ pub const BeamState = struct {
     fn process_attestations(self: *Self, allocator: Allocator, attestations: Attestations, logger: zeam_utils.ModuleLogger) !void {
         const attestations_timer = zeam_metrics.lean_state_transition_attestations_processing_time_seconds.start();
         defer _ = attestations_timer.observe();
-        const attestation_count: u64 = @intCast(attestations.constSlice().len);
 
         if (comptime !zeam_metrics.isZKVM()) {
+            const attestation_count: u64 = @intCast(attestations.constSlice().len);
             zeam_metrics.metrics.lean_state_transition_attestations_processed_total.incrBy(attestation_count);
         }
 


pkgs/api/src/lib.zig
```diff
@@ -58,33 +76,60 @@ pub const Timer = struct {
...
 };
 
 /// A wrapper struct that exposes a `start` function to match the existing API.
 pub const Histogram = struct {
-    is_chain: bool,
+    context: *anyopaque,
+    observeFn: *const fn (*anyopaque, f32) void,
 
     pub fn start(self: *const Histogram) Timer {
         return Timer{
```
gballet: see if you can make `Timer` a generic type so that you can simply define an `observe` function and call it directly from `Timer.observe`, instead of using a function pointer.

pkgs/api/src/lib.zig
```diff
@@ -58,33 +76,60 @@ pub const Timer = struct {
...
+/// Initialized with no-op defaults; properly initialized in init() after the metrics are created.
+/// This allows them to be used safely even if init() is never called (e.g., in tests).
+pub var chain_onblock_duration_seconds: Histogram = .{
+    .context = undefined,
+    .observeFn = &observeNoop,
+};
+pub var block_processing_duration_seconds: Histogram = .{
+    .context = undefined,
+    .observeFn = &observeNoop,
+};
```
gballet: I see, you are forced to use function pointers based on the noop. But then I suggest you do use a `?Histogram` and only report if it's not `null`

noopur23:
lean_state_transition_time_seconds is not implemented

pkgs/node/src/chain.zig
```diff
@@ -195,7 +197,19 @@ pub const BeamChain = struct {
         self.module_logger.debug("node-{d}::going for block production opts={any} raw block={s}", .{ self.nodeId, opts, block_str });
 
         // 2. apply STF to get post state & update post state root & cache it
-        try stf.apply_raw_block(self.allocator, post_state, &block, self.block_building_logger);
+        const stf_timer = zeam_metrics.chain_onblock_duration_seconds.start();
```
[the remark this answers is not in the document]
bomanaps: `chain_onblock_duration_seconds` is already measured where it runs. In chain.zig we start the timer right before calling `stf.apply_raw_block(...)` `/ stf.apply_transition(...)` and observe it immediately after, so the histogram keeps tracking the full onBlock duration. That placement is intentional we use this metric for the node-level  how long does onBlock take view, including the bookkeeping around the STF. for STF-only timing we now record lean_state_transition_time_seconds, plus the per-stage histograms inside the transition itself.

pkgs/node/src/chain.zig
```diff
@@ -369,22 +376,25 @@ pub const BeamChain = struct {
...
             const cpost_state = try self.allocator.create(types.BeamState);
-            try types.sszClone(self.allocator, types.BeamState, pre_state.*, cpost_state);
+            try types.sszClone(self.allocator, types.BeamState, parent_state.*, cpost_state);
 
             // 2. apply STF to get post state
             var validSignatures = true;
             stf.verify_signatures(signedBlock) catch {
                 validSignatures = false;
             };
+            const transition_timer = zeam_metrics.lean_state_transition_time_seconds.start();
```
g11tech: remove timer, and undo parent_state changes, timer is directly to be used inside STF functions

pkgs/state-transition/src/transition.zig
```diff
@@ -27,23 +29,41 @@ pub const StateTransitionOpts = struct {
 //     return;
 // }
 
-// not active in PQ devnet0 - zig will automatically prune this from code
-fn process_execution_payload_header(state: *types.BeamState, block: types.BeamBlock) !void {
-    const expected_timestamp = state.genesis_time + block.slot * params.SECONDS_PER_SLOT;
-    if (expected_timestamp != block.body.execution_payload_header.timestamp) {
-        return StateTransitionError.InvalidExecutionPayloadHeaderTimestamp;
+pub fn is_justifiable_slot(finalized: types.Slot, candidate: types.Slot) !bool {
```
g11tech: why is this function still here?

pkgs/state-transition/src/transition.zig
```diff
@@ -27,23 +29,41 @@ pub const StateTransitionOpts = struct {
...
+    state: *types.BeamState,
+    block: *types.BeamBlock,
+    opts: StateTransitionOpts,
+) !void {
+    const logger = opts.logger;
+    const transition_timer = zeam_metrics.lean_state_transition_time_seconds.start();
+    defer _ = transition_timer.observe();
 
-    // process block and modify the pre state to post state
+    try state.process_slots(allocator, block.slot, logger);
```
g11tech: why are you explictly calling process slot, process_block does all that, as I said earlier restore back all STF changes except collection of metrics at the right places

pkgs/node/src/chain.zig
```diff
@@ -390,23 +393,19 @@ pub const BeamChain = struct {
         });
 
         const post_state = if (blockInfo.postState) |post_state_ptr| post_state_ptr else computedstate: {
-            // 1. get parent state
```
g11tech: why are these comments removed?

PR #315, opened 2025-10-13
add attestation validation

bomanaps:
Add attestation validation matching leanSpec specification with tests

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -325,8 +325,8 @@ pub const BeamChain = struct {
 
                 self.module_logger.debug("chain received vote onGossip cb: {s}", .{signed_vote_str});
 
-                // Validate attestation before processing
-                self.validateAttestation(signed_vote) catch |err| {
+                // Validate attestation before processing (gossip = not from block)
+                self.validateAttestation(signed_vote, false) catch |err| {
                     self.module_logger.debug("Gossip attestation validation failed: {any}", .{err});
                     return; // Drop invalid gossip attestations
                 };
@@ -389,8 +389,8 @@ pub const BeamChain = struct {
             block.slot,
         });
         for (block.body.attestations.constSlice()) |signed_vote| {
-            // Validate attestation before processing
-            self.validateAttestation(signed_vote) catch |e| {
+            // Validate attestation before processing (from block = true)
+            self.validateAttestation(signed_vote, true) catch |e| {
                 self.module_logger.err("invalid attestation in block: validator={d} error={any}", .{ signed_vote.validator_id, e });
                 // Skip invalid attestations but continue processing the block
                 continue;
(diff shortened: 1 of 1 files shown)


pkgs/node/src/chain.zig
```diff
@@ -608,3 +713,473 @@ test "printSlot output demonstration" {
...
+    defer beam_chain.deinit();
+
+    // Add one block
+    const block = mock_chain.blocks[1];
+    try beam_chain.forkChoice.onInterval(block.message.slot * constants.INTERVALS_PER_SLOT, false);
+    try beam_chain.onBlock(block, .{});
+
+    // Create attestation with unknown source block
+    const unknown_root = [_]u8{0xFF} ** 32;
+    const invalid_vote = types.SignedVote{
```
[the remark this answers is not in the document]
g11tech: similarly for other test cases

pkgs/node/src/chain.zig
```diff
@@ -461,13 +467,112 @@ pub const BeamChain = struct {
...
+                vote.target.slot,
+            });
+            return AttestationValidationError.TargetCheckpointSlotMismatch;
+        }
+
+        // 4. Validate attestation is not too far in the future
+        //    current_slot = Slot(self.time // SECONDS_PER_INTERVAL)
+        //    assert vote.slot <= Slot(current_slot + Slot(1))
+        const max_allowed_slot = self.forkChoice.fcStore.timeSlots + constants.MAX_FUTURE_SLOT_TOLERANCE;
+        if (vote.slot > max_allowed_slot) {
```
g11tech: need to check this condition, because by this I can attest to next slot for which block has not even been minted, which doesn't sound right

pkgs/node/src/chain.zig
```diff
@@ -461,13 +467,112 @@ pub const BeamChain = struct {
...
+
+        const target_idx = self.forkChoice.protoArray.indices.get(vote.target.root) orelse {
+            self.module_logger.debug("Attestation validation failed: unknown target block root=0x{s}", .{
+                std.fmt.fmtSliceHexLower(&vote.target.root),
+            });
+            return AttestationValidationError.UnknownTargetBlock;
+        };
+
+        const head_idx = self.forkChoice.protoArray.indices.get(vote.head.root) orelse {
+            self.module_logger.debug("Attestation validation failed: unknown head block root=0x{s}", .{
```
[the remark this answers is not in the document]
bomanaps: Done https://github.com/leanEthereum/leanSpec/pull/75

pkgs/node/src/chain.zig
```diff
@@ -461,7 +467,114 @@ pub const BeamChain = struct {
...
+        //    Per leanSpec store.py:
+        //    - General validation (line 140): assert vote.slot <= Slot(current_slot + Slot(1))
+        //    - Gossip processing (line 177): assert vote.slot <= time_slots
+        //
+        //    Gossip attestations must be for current or past slots only. Validators vote
+        //    in interval 1 of the current slot, so they cannot vote for future slots.
+        //    Block attestations can be more lenient since the block itself was validated.
+        const current_slot = self.forkChoice.fcStore.timeSlots;
+        const max_allowed_slot = if (is_from_block)
+            current_slot + constants.MAX_FUTURE_SLOT_TOLERANCE // Block attestations: allow +1
```
g11tech:
```suggestion
            current_slot - constants.MAX_FUTURE_SLOT_TOLERANCE // Block attestations: allow +1
```

this is the correct condition, i.e. an attestation/vote packed in a proposal block at a slot can't be any later than the previous slot

can you also do the PR in the spec to fix this if thats there

bomanaps: In the spec where have this  `assert vote.slot <= Slot(current_slot + Slot(1)), "Attestation too far in future"` which means it should be `assert vote.slot <= Slot(current_slot - Slot(1)), "Attestation from current or future slot"`  Line https://github.com/leanEthereum/leanSpec/blob/161c0493e980c4f26b70c3f873d0f17b378f6268/src/lean_spec/subspecs/forkchoice/store.py#L178  gossip check would become redundant after this fix (since -1 is stricter than <= current_slot), so should I leave it  as a redundant safety check?

g11tech (approved):

PR #316, opened 2025-10-13
Hashsig

anshalshukla:
- Introduces the xmss package to enable hash-based signature (hashsig) key generation, signing, and verification.
- The xmss package integrates with the hashsig library through Rust bindings, using opaque code blocks in Zig to prevent accidental key exposure or misuse.
- Adds feature flags to Rust crates to prevent testing errors caused by conflicts in libgue.


build.zig
```diff
@@ -34,6 +61,17 @@ pub fn build(b: *Builder) !void {
     // Get git commit hash as version
     const git_version = b.option([]const u8, "git_version", "Git commit hash for version") orelse "unknown";
 
+    // build rust project for different features
+    const rust_builds = struct {
+        networking: *Builder.Step.Run,
+        zkvms: *Builder.Step.Run,
+        crypto: *Builder.Step.Run,
+    }{
```
gballet: you don't need to define a structure, just 3 standalone variables.

PR #328, opened 2025-10-24
feat: add block fetching when validation fails

### LINKED ISSUE #321: add the head block to be pulled via req/resp if the validation fails because of the head root missing in forkchoice
add the head block to be pulled via req/resp if the validation fails because of the head root missing in forkchoice

_Originally posted by @g11tech in https://github.com/blockblaz/zeam/pull/315#discussion_r2445233644_

anshalshukla:
Closes #321

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -433,9 +433,7 @@ pub const BeamChain = struct {
             // Validate attestation before processing (from block = true)
             self.validateAttestation(attestation, true) catch |e| {
                 if (e == AttestationValidationError.UnknownHeadBlock) {
-                    missing_roots.append(attestation.data.head.root) catch |append_err| {
-                        self.module_logger.warn("Failed to track missing root: {any}", .{append_err});
-                    };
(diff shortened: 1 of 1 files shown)


pkgs/node/src/chain.zig
```diff
@@ -402,10 +406,20 @@ pub const BeamChain = struct {
...
+        errdefer missing_roots.deinit();
+
         const signatures = signedBlock.signature.constSlice();
+
         for (block.body.attestations.constSlice(), 0..) |attestation, index| {
             // Validate attestation before processing (from block = true)
             self.validateAttestation(attestation, true) catch |e| {
+                if (e == AttestationValidationError.UnknownHeadBlock) {
+                    missing_roots.append(attestation.data.head.root) catch |append_err| {
+                        self.module_logger.warn("Failed to track missing root: {any}", .{append_err});
```
g11tech: this error is unlkely, only in case of mem overflow, just use try and let the entire onblock fail, no need for excessive logging

g11tech:
> can you also add the behavior as described here:
> 
> #330

pls complete this as a followup PR

PR #329, opened 2025-10-24
split rust libraries into several smaller packages + use dummy prover for normal operations

gballet:
This PR is trying to reduce the overall compilation time by moving the prover compilation outside of the main development loop.

 * it introduces a build option to select the prover at build time
 * it adds a `none` prover, which is the default This means that proving is disabled in the default zeam binary.
 * it only builds the prover-enabled binaries in CI if the main build succeeds.

Replaces #303

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -60,6 +60,9 @@ pub fn build(b: *Builder) !void {
     const prover_option = b.option([]const u8, "prover", "Choose prover: none, risc0, openvm, or all (default: none)") orelse "none";
     const prover = std.meta.stringToEnum(ProverChoice, prover_option) orelse .none;
 
+    // LTO option (disabled by default for faster builds)
+    const enable_lto = b.option(bool, "lto", "Enable Link Time Optimization (slower builds, smaller binaries)") orelse false;
+
     // add ssz
     const ssz = b.dependency("ssz", .{
         .target = target,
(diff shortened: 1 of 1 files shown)


gballet:
@GrapeBaBa the macos build fails, I dont' know why. I know that it calls `objcopy` and somehow that fails on mac. I believe that you are using a mac, could you please see if you get the same error when doing `zig buid` ?

anshalshukla:
@gballet it does fail at `objcopy` on my mac as well.

build.zig
```diff
@@ -459,17 +490,65 @@ pub fn build(b: *Builder) !void {
...
+        .none => "rust/target/release/liblibp2p_glue.a",
+        .risc0 => "rust/target/risc0-release/librisc0_glue.a rust/target/risc0-release/liblibp2p_glue.a",
+        .openvm => "rust/target/openvm-release/libopenvm_glue.a rust/target/openvm-release/liblibp2p_glue.a",
+        .all => "rust/target/release/librisc0_glue.a rust/target/release/liblibp2p_glue.a rust/target/release/libopenvm_glue.a",
+    };
+
+    const weaken_cmd = b.fmt(
+        \\for lib in {s}; do \
+        \\  if [ -f "./$lib" ]; then \
+        \\    objcopy -w \
```
gballet: looks like obcopy doesn't work on mac, at least not that installed version on this specific macos

PR #333, opened 2025-10-27
implement xmss with the hashsig rust crate

gballet:
This is a rebase of #316, but I didn't want to overwrite the source branch just in case I missed something.

@anshalshukla I have simplified the integration, and as far as I understand I no longer need the features. Please double-check what I did makes sense.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -314,7 +314,6 @@ pub fn build(b: *Builder) !void {
 
     try build_zkvm_targets(b, &cli_exe.step, target);
 
-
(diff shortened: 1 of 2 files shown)


anshalshukla:
Yup, it looks much cleaner now

gballet:
This isn't merged yet, because of the linter issues, but also because the build is broken on fedora.

gballet:
The fedora issue is due to a newer version of the linker, which is stricter. I won't stop the merge of this PR just for this, I'll fix it in a later PR.

PR #343, opened 2025-11-01
Add node-scoped logging to RustBridge

bomanaps:
closes https://github.com/blockblaz/zeam/issues/97

Screenshot 2025-11-01 at 07 39 05


rust/libp2p-glue/src/logger.rs
```diff
@@ -0,0 +1,63 @@
...
+        LogLevel::Debug => 0,
+        LogLevel::Info => 1,
+        LogLevel::Warn => 2,
+        LogLevel::Error => 3,
+    }
+}
+
+// Build a plain message string (optionally with a [module] prefix) and forward to Zig
+
+fn log_with_level(level: LogLevel, network_id: u32, module: Option<&str>, message: &str) {
```
g11tech: remove module since its no more used here,

PR #350, opened 2025-11-06
fix: build Docker images natively to avoid Zig HTTP bug

### LINKED ISSUE #261: amd64 img not building anymore
got a message from Barnabas on this 

Image

https://github.com/ethpandaops/eth-client-docker-image-builder/actions/runs/18087107869/job/51459998960

chetanyb:
### Problem
  Docker builds fail intermittently due to [Zig HTTP connection pool bug](https://github.com/ziglang/zig/issues/21316).

  ### Solution
  - Modified CI to build zeam natively first, then package with `Dockerfile.prebuilt`
  - Avoids Zig HTTP bug by not fetching dependencies inside Docker
  - Currently builds AMD64 only (CI runs on `ubuntu-latest`)

  ### Future: ARM64 + Multi-Arch
  To add ARM64 and multi-arch manifest later:
  1. Add self-hosted ARM64 runner
  2. Create parallel build jobs for AMD64/ARM64
  3. Add manifest job to combine both architectures

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,8 +246,8 @@ jobs:
     needs: [lint, build, test, build-all-provers]
     strategy:
       matrix:
-        arch: [amd64, arm64]
-    runs-on: [ ubuntu-latest, matrix.arch ]
+        builder: [ubuntu-latest, ubuntu-22.04-arm]
+    runs-on: ${{ matrix.builder }}
     steps:
     - uses: actions/checkout@v4
       with:


Dockerfile.prebuilt
```diff
@@ -0,0 +1,84 @@
...
+    if [ -d "/lib/$LIBDIR" ]; then \
+        mkdir -p "/runtime-libs/lib/$LIBDIR" && \
+        for lib in libc.so.6 libm.so.6 libpthread.so.0 libdl.so.2 librt.so.1 libgcc_s.so.1 libstdc++.so.6; do \
+            [ -f "/lib/$LIBDIR/$lib" ] && cp -L "/lib/$LIBDIR/$lib" "/runtime-libs/lib/$LIBDIR/" || true; \
+        done; \
+    fi && \
+    if [ -f "$LDSO" ]; then \
+        mkdir -p "/runtime-libs$(dirname $LDSO)" && \
+        cp -L "$LDSO" "/runtime-libs$LDSO"; \
+    fi
```
gballet: These libraries should already be present in the container, since you build on ubuntu and you also use Ubuntu in the container.

Dockerfile.prebuilt
```diff
@@ -0,0 +1,84 @@
...
+
+# Runtime stage - using scratch for minimal size
+FROM scratch AS runtime
+
+# Copy the architecture-specific libraries and loader
+COPY --from=runtime-prep /runtime-libs/ /
+
+# Copy SSL certificates for HTTPS
+COPY --from=runtime-prep /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
```
gballet: Can't you just apt install these ? The http issues are only for zig, or also for the system packaging as well?

README.md
```diff
@@ -62,6 +62,18 @@ out for this space for annoucements on the same or reach out to us via our [comm
...
 ### Build instructions
 
+#### Docker
+
+Docker images are built in CI using `Dockerfile.prebuilt`, which packages pre-built binaries. This avoids intermittent build failures caused by a [Zig HTTP connection pool bug](https://github.com/ziglang/zig/issues/21316) when building inside Docker.
+
+To build a Docker image locally:
+```bash
+# Build zeam natively first
+zig build -Doptimize=ReleaseFast -Dgit_version="$(git rev-parse --short HEAD)"
```
gballet: In not at my desk so I can't check, but you shouldn't have to pass the version, it should be added for you iirc.

chetanyb: I just tested this again and the `-Dgit_version` parameter must be passed explicitly since zig doesn't auto-capture git metadata. Testing: without the flag `--version` shows "unknown", with it shows the actual commit hash.

chetanyb:
> @chetanyb it builds for arm64 and amd64 but it doesn't seem to push the images to any remote repository. See this message:
> 
> ```
>  WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
> ```

yes, that workflow is not in this PR as @noopur23 is handling that along with the blockblaz repo PAT. She will be adding the reusable workflow once the PR is merged.

PR #352, opened 2025-11-09
signed and verified block and attestations with xmss signatures

chetanyb:
## Desription
- Add changes from https://github.com/leanEthereum/leanSpec/pull/117/
- Wire up the Rust hash-sig XMSS scheme via a Zig FFI layer, exposing keygen/sign/verify plus bincode serialization helpers so other modules can operate on opaque handles safely.
- Extend state-transition flow to carry  XMSS signatures per attestation, run `stf.verifySignatures` before applying blocks, and have forkchoice consume the verified signatures during block processing.
- Add TestKeyManager, full XMSS cycle tests, and XMSS-aware mock-chain generation so deterministic fixtures produce real signatures; update the build to link the new glue library and surface the testing module.

### DIFF
--- a/pkgs/node/src/node.zig
+++ b/pkgs/node/src/node.zig
@@ -358,6 +358,7 @@ pub const BeamNode = struct {
                 self.chain.forkChoice.fcStore.time,
                 itime_intervals,
             });
+            return;
         }
 
         // till its time to attest atleast for first time don't run onInterval,


pkgs/testing/signature_helpers.zig
```diff
@@ -0,0 +1,82 @@
...
+        const keys = try allocator.alloc(xmss.KeyPair, num_validators);
+        errdefer allocator.free(keys);
+
+        const num_active_epochs = max_slot + 1;
+
+        for (keys, 0..) |*key, i| {
+            const seed = try std.fmt.allocPrint(allocator, "test_validator_{d}", .{i});
+            defer allocator.free(seed);
+
+            key.* = try xmss.KeyPair.generate(
```
g11tech: they generate same keypair each time?

chetanyb: yes

pkgs/configs/src/lib.zig
```diff
@@ -51,52 +51,59 @@ const ChainConfigError = error{
...
-    try std.testing.expect(genesis_config.genesis_time == 1704085200);
-    try std.testing.expect(genesis_config.num_validators == 9);
-
-    const genesis_config_override = try genesisConfigFromYAML(yaml, 1234);
-    try std.testing.expect(genesis_config_override.genesis_time == 1234);
-    try std.testing.expect(genesis_config_override.num_validators == 9);
-}
-
-test "custom dev chain" {
-    const dev_spec =
```
g11tech: also restore such parsing in a followup PR here or in the mock test sequence in the state transition

PR #360, opened 2025-11-12
fix: genMockChain leak in prover

gballet:


gballet:
except it's not reopening :facepalm:

PR #363, opened 2025-11-12
add a dummy prover

gballet:
PRs got merged, that broke the prover. This PR implements a dummy prover, to be run in tests, so that these errors can be caught.


g11tech (approved):
lgtm, amazing now we can cover proving manager in tests as well

PR #365, opened 2025-11-13
fix: allocs in mocked chain

gballet:


pkgs/state-transition/src/mock.zig
```diff
@@ -12,28 +12,28 @@ const transition = @import("./transition.zig");
 const MockChainData = struct {
     genesis_config: types.GenesisSpec,
     genesis_state: types.BeamState,
-    blocks: []types.SignedBlockWithAttestation,
-    blockRoots: []types.Root,
+    blocks: std.ArrayList(types.SignedBlockWithAttestation),
```
g11tech: aren't they toOwnedSlices?

PR #368, opened 2025-11-14
feat: recursive parent block syncing

### LINKED ISSUE #330: track depth of a block by root request and recursively sync parents till a particular configurable depth
currently we ensureBlocksByRootRequest on missing parent blocks, and when the response comes we do processBlockByRootChunk to add them onto chain, but again if parent blocks are missing, we need to sync parent (so a new ensureBlocksByRootRequest needs to be initiated) , while caching the current recieved block till we either get process the entire chain, or we hit a depth where we discard the requests

So `pending_block_roots` should now be a map to depth from the first child that started the fetch, also cache already fetched blocks into another root to block map to be added to node/network 
 - and whenever we need to do ensureBlocksByRootRequest, we check in the fetched blocks as well  to filter out those we have already fetched

anshalshukla:
Closes #330 
- Recursively fetches parent block upto the depth of `MAX_BLOCK_FETCH_DEPTH` or until `MAX_CACHED_BLOCKS` are found.
- Handles multiple blocks at same slot number to handle equivocation logic in future.
- Updates `PendingBlockRootMap` to store depth against each block root.
- Adds `FetchedBlockMap` to cache blocks that has been fetched but cannot be processed due to `MissingPreState`
- Refactors `node.zig` tests to move reusable functions into separate  `testing.zig` file.
- Adds test cases to verify that the chain auto processes it future blocks if available in cache.

### DIFF
--- a/pkgs/node/src/constants.zig
+++ b/pkgs/node/src/constants.zig
@@ -11,7 +11,7 @@ pub const MAX_FUTURE_SLOT_TOLERANCE = 1;
 
 // Maximum depth for recursive block fetching
 // When fetching parent blocks, we stop after this many levels to avoid infinite loops
-pub const MAX_BLOCK_FETCH_DEPTH = 64;
+pub const MAX_BLOCK_FETCH_DEPTH = 512;
 
 // Maximum number of blocks to keep in the fetched blocks cache
 // This prevents unbounded memory growth from malicious peers sending orphaned blocks


pkgs/node/src/chain.zig
```diff
@@ -381,15 +405,16 @@ pub const BeamChain = struct {
 
                 // Validate attestation before processing (gossip = not from block)
                 self.validateAttestation(signed_attestation.message, false) catch |err| {
-                    self.module_logger.warn("gossip attestation validation failed: {any}", .{err});
-                    return; // Drop invalid gossip attestations
+                    self.module_logger.debug("Gossip attestation validation failed: {any}", .{err});
```
g11tech: why to debug from warn?

pkgs/node/src/node.zig
```diff
@@ -155,7 +157,34 @@ pub const BeamNode = struct {
...
+    }
+
+    fn handleGossipProcessingResult(self: *Self, result: chainFactory.GossipProcessingResult) void {
+        // Process successfully imported blocks to retry any cached descendants
+        if (result.processed_block_root) |processed_root| {
+            self.logger.debug(
+                "Gossip block 0x{s} successfully processed, checking for cached descendants",
+                .{std.fmt.fmtSliceHexLower(processed_root[0..])},
+            );
+            self.processCachedDescendants(processed_root);
```
g11tech: if a block was rejected because it was before finalized, we also need to prune out all its descendants

g11tech: in that case we also ignore missing attestation roots

pkgs/node/src/node.zig
```diff
@@ -177,8 +278,54 @@ pub const BeamNode = struct {
                 });
             }
 
+            // Try to add the block to the chain
             const missing_roots = self.chain.onBlock(signed_block.*, .{}) catch |err| {
-                self.logger.warn("failed to import block fetched via RPC 0x{s} from peer {s}{}: {any}", .{
+                // Check if the error is due to missing parent
+                if (err == chainFactory.BlockProcessingError.MissingPreState) {
```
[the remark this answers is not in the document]
g11tech: similar case for onblock on gossip as well

pkgs/node/src/node.zig
```diff
@@ -239,7 +395,7 @@ pub const BeamNode = struct {
                                 self.node_registry.getNodeNameFromPeerId(block_ctx.peer_id),
                             });
 
-                            self.processBlockByRootChunk(block_ctx, &block_resp);
+                            try self.processBlockByRootChunk(block_ctx, &block_resp);
```
g11tech: what are the failure cases here?

pkgs/node/src/node.zig
```diff
@@ -525,8 +682,8 @@ pub const BeamNode = struct {
             });
             defer self.allocator.free(missing_roots);
 
-            self.fetchBlockByRoots(missing_roots) catch |err| {
-                self.logger.warn("failed to fetch {d} missing block(s): {any}", .{ missing_roots.len, err });
+            self.fetchBlockByRoots(missing_roots, 0) catch |err| {
+                self.logger.warn("Failed to fetch {d} missing block(s): {any}", .{ missing_roots.len, err });
```
g11tech: keep lowercase

pkgs/node/src/node.zig
```diff
@@ -711,6 +864,199 @@ test "Node peer tracking on connect/disconnect" {
...
+
+    // Advance forkchoice time to block1 slot and add block1 to the chain
+    try node.chain.forkChoice.onInterval(block1_slot * constants.INTERVALS_PER_SLOT, false);
+    const missing_roots1 = try node.chain.onBlock(block1, .{});
+    defer allocator.free(missing_roots1);
+
+    // Verify block1 is now in the chain
+    try std.testing.expect(node.chain.forkChoice.hasBlock(block1_root));
+
+    // Now call processCachedDescendants with block1_root. This should discover
```
g11tech: do any cases need more rigorous testing? or any other pending test cases, create an issue and assign to mercy

g11tech (commented):
create a task to add a sim where we start one node late (using a cli param `--testParentSyncing` for the beam sim) 

then the second node should syncup using this and finalize i.e. the normal beam sim resumes post second node is synced, assign it to mercy

pkgs/node/src/node.zig
anshalshukla: It doesn't return an error in most of the cases and just logs it expect following cases:
1. allocation errors
2. ssz clone errors
3. cache map put errors
4. fetchBlockByRoots errors

pkgs/node/src/node.zig
```diff
@@ -165,9 +213,142 @@ pub const BeamNode = struct {
...
+        while (stack.items.len > 0) {
+            const current_root = stack.pop().?;
+            // Collect immediate cached children first (don't mutate the map during iteration).
+            var children = std.ArrayList(types.Root).init(self.allocator);
+            defer children.deinit();
+
+            // TODO: Optimize this to use a more efficient data structure
+            // It is a O(n^2) operation to find all children of a parent and their descendants
+            // We should use a more efficient data structure to store the cached blocks
+            // along with the children roots so it's linear time operation in terms of the number of children
```
anshalshukla: https://github.com/blockblaz/zeam/pull/368/files#r2649541261

Issue #372, opened 2025-11-15
Fix validator manifest to use hex pubkeys and populate genesis_validators in config.yaml


noopur23:
1. fix validator-manifest.yaml to actually use hex bytes in the pubkeys and not file name in hash-sig-cli for lean-quickstart
2. dump those pubkeys in config.yaml under property: genesis_validators

noopur23:
closing this issue, as issue needs to be done in lean-quickstart https://github.com/blockblaz/lean-quickstart/issues/39

PR #373, opened 2025-11-16
update the `node` command to consume `devnet1` configuration

bomanaps:
closes https://github.com/blockblaz/zeam/issues/372

### DIFF
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -294,16 +294,11 @@ fn mainInner() !void {
             };
             var chain_options = (try json.parseFromSlice(ChainOptions, gpa.allocator(), chain_spec, options)).value;
 
-            const time_now_ms: usize = @intCast(std.time.milliTimestamp());
-            const time_now: usize = @intCast(time_now_ms / std.time.ms_per_s);
-
-            chain_options.genesis_time = time_now;
-
             // Create key manager FIRST to get validator pubkeys for genesis
             const key_manager_lib = @import("@zeam/key-manager");
             // Using 3 validators: so by default beam cmd command runs two nodes to interop
             const num_validators: usize = 3;
-            var key_manager = try key_manager_lib.getTestKeyManager(allocator, num_validators, 10000);
+            var key_manager = try key_manager_lib.getTestKeyManager(allocator, num_validators, 1000);
             defer key_manager.deinit();
 
             // Get validator pubkeys from keymanager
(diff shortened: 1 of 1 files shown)


bomanaps:
Zig YAML parser has a bug quoted strings starting with 0x are parsed as floats when the YAML comes from an inline string literal. but the same YAML loaded from a file is parsed correctly, which  breaks our handling of hex pubkeys.
For our use case (genesis_validators list of 52-byte hex public keys), In validator-manifest.yaml: keep pubkeys with 0x prefix (quoted) for human readability/cross-client docs. In config.yaml: write 104-char hex strings WITHOUT 0x (quoted), which the YAML parser consistently treats as .string. In parsing code: accept either format by stripping optional 0x during decode: hexToBytes52(input): if starts with 0x, strip validate length (104 chars) and decode Notes.
I also tested Unquoted 0x: always becomes a number → unusable Block scalars (|): work but add whitespace/newline management, and are awkward in sequences Single quotes around 0x: still not reliable for inline YAML

bomanaps:
> I think only option 2 need to be implemented in Zeam repo, the option 1 may need to change the code in the leanSpec repo. So we don't need [hashsig_manifest.zig](https://github.com/blockblaz/zeam/pull/373/files#diff-78da4d3ceebcd5ca5914428e209a2e97af10a7d23087d4db7b485a74584d6f51) and the new CLI in Zeam. Is it correct?

Yes thought about this but then the library generates random keys, not deterministic derivation from existing privkeys, so in other sense hash-sig-cli should handle: CLI for library's random key generation
Zeam should handle: Deterministic derivation from existing privkeys (workflow-specific)

GrapeBaBa:
LGTM

noopur23:
which yaml parser are you using ? have you tried some other libraries for the "0x" bug ?

build.zig
```diff
@@ -336,6 +337,7 @@ pub fn build(b: *Builder) !void {
     cli_exe.root_module.addImport("enr", enr);
     cli_exe.root_module.addImport("yaml", yaml);
     cli_exe.root_module.addImport("@zeam/key-manager", zeam_key_manager);
+    cli_exe.root_module.addImport("@zeam/xmss", zeam_xmss);
```
g11tech: why is this needed in cli?

g11tech:
@GrapeBaBa pls take over the PR while @bomanaps make the corresponding changes in lean-quickstart to fix the validator manifest and to also list all the pubkeys (correctly by index) in the config.yaml we generate

PR #375, opened 2025-11-16
fix: capture all runtime libraries in Dockerfile.prebuilt

chetanyb:
## Problem

  The Docker image built with `Dockerfile.prebuilt` fails at runtime with:
  exec /app/zig-out/bin/zeam: no such file or directory

  This was reported when attempting to run the prebuilt Docker image on Linux.

  ## Root Cause

  The library copying logic uses a regex that only matches paths starting with `/lib/`:
  ```bash
  grep -oP '(\/lib\/[^\s]+)'
  ```


gballet (commented):
That doesn't seem right: if the `zeam` executable isn't found, this is not a problem of copying the libraries. Which Linux are you building this on?

This is what I get on ubuntu aarch64:

```
> ldd zig-out/bin/zeam
        linux-vdso.so.1 (0x0000f0b770557000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000f0b770350000)
        /lib/ld-linux-aarch64.so.1 (0x0000f0b77051e000)
```

The first library doesn't exist and shouldn't be copied. The other two need to, but once again I don't think this is the fix for the problem you give in the description.

g11tech (approved):

PR #382, opened 2025-11-23
fix: set signature length to 3116

### LINKED ISSUE #381: Zeam's Signature type is using FixedBytes<4000> but shouldn't it be FixedBytes<3116>?
Zeam's Signature type is using FixedBytes<4000> but shouldn't it be FixedBytes<3116>?

anshalshukla:
Closes #381

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -341,6 +341,7 @@ pub fn build(b: *Builder) !void {
     cli_exe.step.dependOn(&build_rust_lib_steps.step);
     addRustGlueLib(b, cli_exe, target, prover);
     cli_exe.linkLibC(); // for rust static libs to link
+    cli_exe.linkLibCpp(); // for rocksdb C++ library to link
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -150,18 +150,18 @@ jobs:
 
 
     - name: Cache Zig packages
-      uses: actions/cache@v3
+      uses: actions/cache@v4
       with:
         path: ~/.cache/zig
-        key: ${{ runner.os }}-zig-packages-${{ hashFiles('build.zig.zon') }}
+        key: ${{ runner.os }}-zig-packages-${{ github.sha }}
```
g11tech: this effectively removes caching right?

anshalshukla: ugh, turns out it's a github runner bug. I think we can stay with this for now until it is fixed.

https://github.com/actions/runner/issues/4134

PR #383, opened 2025-11-23
fix ci build caching issues and add arm64 docker images

ch4r10t33r:
- zeam failure on arm64 should be resolved with this fix
Screenshot 2025-11-23 at 10 53 24
zeam docker images fail to run on arm64 machines. The fail with the above error. This PR sets RUSTFLAGS="-C target-cpu=generic" when building for ARM64, so Rust compiles with QEMU-compatible instructions instead of CPU-specific features that QEMU can't emulate.
- updates to ci to fix the caching build failures on the MacOS build step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,10 +69,20 @@ jobs:
         rustc --version
         cargo --version
 
+    - name: Get Cargo.lock hash
+      id: cargo-lock-hash
+      run: |
+        if [ -f rust/Cargo.lock ]; then
+          HASH=$(sha256sum rust/Cargo.lock 2>/dev/null | cut -d' ' -f1 || shasum -a 256 rust/Cargo.lock | cut -d' ' -f1)
+          echo "hash=${HASH}" >> $GITHUB_OUTPUT
+        else
+          echo "hash=default" >> $GITHUB_OUTPUT
+        fi
+
     - name: Cache dependencies
       uses: Swatinem/rust-cache@v2
       with:
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+        key: ${{ runner.os }}-cargo-${{ steps.cargo-lock-hash.outputs.hash }}
 
     - name: Check Rust formatting
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
gballet: Are these changes necessary? I don't see how modifying the caching would solve an arm-specific build.

Also,bid necessary, explain why that is in the pr description, to speed up the review process. Please.

Dockerfile
```diff
@@ -43,6 +43,17 @@ RUN if [ "$TARGETARCH" = "amd64" ]; then \
     fi
 ENV PATH="/root/.risc0/bin:${PATH}"
 
+# Configure Rust for ARM64 QEMU compatibility
+# Use generic CPU target to avoid illegal instructions when running under QEMU emulation
+ARG TARGETARCH
+RUN if [ "$TARGETARCH" = "arm64" ]; then \
+        rustup target add aarch64-unknown-linux-gnu; \
```
gballet: Why is that necessary? The docker arm image should be built on an arm Ubuntu machine, so that should be installed automatically.

.github/workflows/risc0.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/.cache/zig
-        key: ${{ runner.os }}-zig-packages-${{ hashFiles('build.zig.zon') }}
+        key: ${{ runner.os }}-zig-packages-${{ hashFiles('**/build.zig.zon') }}
```
gballet: That might be a useful fix, but I'd like to have it in another PR to speed things up.

Dockerfile
ch4r10t33r: Mihir on zeam telegram reported that he was facing this error

`qemu: uncaught target signal 4 (Illegal instruction) - core dumped`

Dockerfile
```diff
@@ -30,8 +30,8 @@ RUN ZIG_VERSION="0.14.1" && \
     mv "zig-${ZIG_ARCH}-linux-${ZIG_VERSION}" /opt/zig && \
     ln -s /opt/zig/zig /usr/local/bin/zig
 
-# Install Rust 1.85+
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.85.0
+# Install Rust nightly (required by build.zig which uses +nightly)
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain nightly
```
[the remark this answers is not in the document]
gballet: But it wasn't, which is what surprises me

.github/workflows/ci.yml
g11tech: weird why hashFiles sudddently started failing

ch4r10t33r: I don't know. This has been failing since this morning. Even @anshalshukla 's PR had this issue.

Dockerfile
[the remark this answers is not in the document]
ch4r10t33r: Ok, I dug into this. It appears that rustup detects nightly is used in build.zig and automatically installs it, that is why the build succeeded. So do we retain this for clarity or remove it? @gballet

PR #401, opened 2025-11-27
genesis state validator index

turuslan:
According to logs from [lean-quickstart](https://github.com/blockblaz/lean-quickstart) test,
zeam genesis state/block hash didn't match ream.
Reason was that validator indices were not initialized.


g11tech (dismissed):
thanks @turuslan for debugging this, :heart: much apperciated

unnawut:
I missed this PR by 3 hours 😭

PR #402, opened 2025-11-27
fix: Resolve Docker run stalling issue in Zeam

ch4r10t33r:
- Missing or inaccessible config directory: buildStartOptions causes the program to stall/hang when run via Docker (mostly on MacOS).
- Added decorators to print SimArgs to make the output more human readable.


chetanyb (dismissed):
LGTM, worked for me

g11tech (approved):

PR #413, opened 2025-12-01
fix: fix the spec test for latest format

GrapeBaBa:
**Motivation**

Make the all the spec tests works with latest lean spec test vector format.

**Description**
- Added `lexicographicHeadAmong` support for new test vector
- Aligned the new test vector json format to parse block correctly
- Fix `validators` index assignment from parsed json index value


g11tech (approved):
lgtm :rocket:

PR #417, opened 2025-12-03
feat: add node registry to track peer_id, validator_id and improve logging

### LINKED ISSUE #411: read all nodes from validator config into a map and log their names
peerID => node name,
validators ids => node name
   
So Log the node name for any log generated for peer interfactions where ENR or validator id is available 
 - gossip objects like block/attestation from propser id,  atttestation's validator id
 - from peerID
 
Incase node name is not available, no need to add it to the log

anshalshukla:
Closes #411

### DIFF
--- a/pkgs/utils/src/log.zig
+++ b/pkgs/utils/src/log.zig
@@ -488,7 +488,7 @@ test "OptionalNode formatter" {
         });
 
         const result = fbs.getWritten();
-        try testing.expectEqualStrings("(\x1b[38;5;201malice\x1b[0m) Peer connected: peer123, total peers: 5", result);
+        try testing.expectEqualStrings("(" ++ Colors.peer ++ "alice" ++ Colors.reset ++ ") Peer connected: peer123, total peers: 5", result);
     }
 
     // Test with node name null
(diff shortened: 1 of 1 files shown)


pkgs/node/src/chain.zig
```diff
@@ -328,10 +340,11 @@ pub const BeamChain = struct {
                 //check if we have the block already in forkchoice
                 const hasBlock = self.forkChoice.hasBlock(block_root);
 
-                self.module_logger.info("chain received gossip block for slot={any} blockroot={any} hasBlock={any}", .{
-                    //
+                self.module_logger.info("{}chain received gossip block for slot={any} blockroot={any} proposer={d} hasBlock={any}", .{
+                    zeam_utils.OptionalNode.init(self.getNodeNameForLog(block.proposer_index)),
```
g11tech: this is this node's log, node registry is not to print the name of this node, but to print the objects recieved from which node so one can see what our peers are sending us to easily identify who is sending what to us,

so this log should read: chain received gossip block from node={} for slot={}... and we fill the name of the node there from the node registery

Issue #415, opened 2025-12-02
consume  ssz encoded validator keys in zeam


noopur23:
Based on https://github.com/blockblaz/hash-sig-cli/issues/27

consume  ssz encoded validator keys in zeam

PR #418, opened 2025-12-04
add ssz validator key consumption

chetanyb:
## Description
This PR adds support for consuming ssz encoded validator keys while keeping existing json parsing as fallback. Closes https://github.com/blockblaz/zeam/issues/415

## Validation
- Tested locally using: https://github.com/blockblaz/lean-quickstart/pull/63

### DIFF
--- a/pkgs/cli/src/node.zig
+++ b/pkgs/cli/src/node.zig
@@ -70,6 +70,7 @@ pub const NodeOptions = struct {
     logger_config: *LoggerConfig,
     database_path: []const u8,
     hash_sig_key_dir: []const u8,
+    node_registry: *node_lib.NodeNameRegistry,
 
     pub fn deinit(self: *NodeOptions, allocator: std.mem.Allocator) void {
         for (self.bootnodes) |b| allocator.free(b);
@@ -77,6 +78,8 @@ pub const NodeOptions = struct {
         allocator.free(self.validator_indices);
         allocator.free(self.local_priv_key);
         allocator.free(self.hash_sig_key_dir);
+        self.node_registry.deinit();
+        allocator.destroy(self.node_registry);
     }
 };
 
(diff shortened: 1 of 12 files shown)


pkgs/cli/src/constants.zig
```diff
@@ -23,3 +23,6 @@ pub const DEFAULT_NODE_KEY: []const u8 = "./key";
 
 /// Maximum size (in bytes) for hash-sig key JSON blobs ingested by the CLI
 pub const MAX_HASH_SIG_KEY_JSON_SIZE: usize = 128 * 1024 * 1024;
+
+/// Maximum size (in bytes) for hash-sig key SSZ blobs ingested by the CLI
+pub const MAX_HASH_SIG_KEY_SSZ_SIZE: usize = 128 * 1024 * 1024;
```
gballet: it's exactly the same as `MAX_HASH_SIG_KEY_JSON_SIZE` and it's huuuuuuuuuuuuge.

If it really makes sense to have the exact same value for a binary encoding as for the JSON encoding, then just rename the variable above `MAX_HASH_SIG_ENCODED_KEY_SIZE`. If it really makes sense to differentiate the two, use a smaller value.

pkgs/cli/src/node.zig
```diff
@@ -317,36 +317,73 @@ pub const Node = struct {
...
+                defer self.allocator.free(secret_json);
+
+                var keypair = try xmss.KeyPair.fromJson(
+                    self.allocator,
+                    secret_json,
+                    public_json,
+                );
+                errdefer keypair.deinit();
+
+                try self.key_manager.addKeypair(validator_index, keypair);
```
gballet: this could be deduplicated for better readability

g11tech (approved):
lgtm for now, any changes requested by Guillaume to be taken as followup PR

Issue #292, opened 2025-10-04
UX - Node Monitoring Visual Tool


anshalshukla:
On the 3rd October Zeam call, we discussed ways to improve the infra runner experience. Some of the ideas included:

- Add a CLI flag to provide clear, structured monitoring of node state and metrics.

- Develop a dedicated CLI tool for node monitoring and parameter updates—similar to Geth’s IPC, but designed to be more visual, intuitive, and cross-platform (Linux, Mac, Windows). (maybe ncurses)

- Explore a web-based dashboard, where node parameters can be queried and displayed in a user-friendly format. This could be hosted locally when the node is running, or triggered with specific CLI flags.

We’re also open to further suggestions to enhance the experience.

The crux of this initiative is that users often struggle to parse raw CLI logs to understand the current status of a node—we want to make this much simpler and more accessible.

PR #426, opened 2025-12-14
forkchoice grafana visualization

chetanyb:
## Summary

Adds real-time fork choice tree visualization via `/api/forkchoice/graph` endpoint with thread-safety fixes to forkchoice module. Implements locked/unlocked pattern for 5 critical functions and eliminates unsafe direct field access.

#### Visualization API
  - New HTTP endpoint serving Grafana node-graph compatible JSON
  - Configurable history via ?slots=N (default: 50, max: 200)
  - Color-coded arc borders representing consensus states:
    - 🟣 Purple: Finalized blocks (canonical chain)
    - 🔵 Blue: Justified checkpoint
    - 🟠 Orange: Current head
    - 🟢 Green: Normal blocks (refer [TODO](https://github.com/blockblaz/zeam/blob/a7b602682c605592f84faf603802583f4b1b79e2/pkgs/cli/src/api_server.zig#L232-L233))
    - ⚫ Gray: Orphaned blocks (historical forks)
  - Arc completion represents validator weight
  - New endpoint /lean/states/finalized (SSZ checkpoint state).
  - Unified API server for `/metrics`, `/health`, `/events`, `/api/forkchoice/graph`, `/lean/states/finalized`.

 #### Thread Safety

  **Locking Pattern Implementation**
  - Fixed 5 functions missing thread-safety protection:
    - `getCanonicalView` (READ operations - lockShared)
    - `getCanonicalityAnalysis` (READ operations - lockShared)
    - `getCanonicalAncestorAtDepth` (READ operations - lockShared)
    - `rebase` (WRITE operations - exclusive lock)
    - `confirmBlock` (WRITE operations - exclusive lock)

  - Implemented locked/unlocked pattern:
    - Internal `*Unlocked` versions for efficient internal calls
    - Public wrappers with appropriate locks
    - No nested locking patterns detected

  **Eliminated Direct Field Access**
  - Replaced unsafe direct access in chain.zig with thread-safe getters:
    - `self.forkChoice.head` → `getHead()`
    - `self.forkChoice.fcStore` → `getLatestJustified()`, `getLatestFinalized()`
    - `self.forkChoice.protoArray.*` → `getNodeCount()`, `getBlockSlot()`

  **Concurrency Benefits**
  - Snapshot uses shared lock (allows concurrent reads)
  - Multiple readers don't block each other
  - Visualization doesn't block block processing

 #### API Server Reliability
  - Per-connection arena allocator prevents memory leaks on request failures
  - Made server stoppable with proper lifecycle management
  - Fixed potential use-after-free when accessing forkchoice pointer
  - Non-blocking accept loop with graceful shutdown
  - SSE lifecycle fixes:
      - heartbeat uses mutexed connection writes
      - broadcaster no longer frees connections on send failure
      - safe shutdown drain for SSE threads
      - removeGlobalConnection for cleanup
  - Rate limiter eviction is safe (no mutation during iteration).
  - chain can be attached after startup via handle.

## Security & Performance

  - Max slot cap (200) prevents excessive memory/lock time
  - Shared locks minimize contention
  - Lock-free JSON processing on snapshot copy
  - Per‑IP rate limiting + global in‑flight cap for graph endpoint (429 on exceed).
  - SSE connection cap (32) to avoid unbounded threads (503 on exceed).
  - Rate‑limit map bounded (256 IPs) with TTL cleanup + cooldown.

## Screenshot sample

Grafana panel showing fork-choice tree visualization

  ---
  ## NOTES
 - Companion [Grafana dashboards](https://github.com/blockblaz/zeam-dashboards)
 - Can be useful for https://github.com/blockblaz/zeam/issues/292

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -1334,6 +1334,7 @@ pub const BeamChain = struct {
         no_peers,
         behind_peers: struct {
             head_slot: types.Slot,
+            finalized_slot: types.Slot,
             max_peer_finalized_slot: types.Slot,
         },
     };
@@ -1362,10 +1363,20 @@ pub const BeamChain = struct {
             }
         }
 
-        // We must also be synced with peers (at or past max peer finalized slot)
+        // Check 1: our head is behind peer finalization — we don't even have finalized blocks
         if (our_head_slot < max_peer_finalized_slot) {
             return .{ .behind_peers = .{
                 .head_slot = our_head_slot,
+                .finalized_slot = our_finalized_slot,
+                .max_peer_finalized_slot = max_peer_finalized_slot,
+            } };
+        }
+
+        // Check 2: our finalization is behind peer finalization — we may be on a divergent fork
+        if (our_finalized_slot < max_peer_finalized_slot) {
+            return .{ .behind_peers = .{
+                .head_slot = our_head_slot,
+                .finalized_slot = our_finalized_slot,
                 .max_peer_finalized_slot = max_peer_finalized_slot,
             } };
         }
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ data/
 
 # VSCode files
 .vscode/
+
+# Tool versions
+.tool-versions
--- a/pkgs/node/src/constants.zig
+++ b/pkgs/node/src/constants.zig
@@ -21,6 +21,13 @@ pub const MAX_CACHED_BLOCKS = 1024;
 // Set to 7200 slots (approximately 8 hours in Lean, assuming 4 seconds per slot)
 pub const FORKCHOICE_PRUNING_INTERVAL_SLOTS: u64 = 7200;
 
+// Forkchoice visualization constants
 pub const MAX_FC_DISPLAY_DEPTH = 100;
 pub const MAX_FC_DISPLAY_BRANCH = 10;
 pub const MAX_FC_CHAIN_PRINT_DEPTH = 5;
+
+// Timeout for pending RPC requests in seconds.
+// If a peer does not respond within this duration, the request is finalized and retried
+// with a different peer. 2 slots at 4s/slot is generous for latency while ensuring
+// stuck sync chains recover quickly.
+pub const RPC_REQUEST_TIMEOUT_SECONDS: i64 = 8;
--- a/pkgs/node/src/network.zig
+++ b/pkgs/node/src/network.zig
@@ -43,7 +43,16 @@ pub const PendingRPC = union(enum) {
     }
 };
 
-pub const PendingRPCMap = std.AutoHashMap(u64, PendingRPC);
+pub const PendingRPCEntry = struct {
+    request: PendingRPC,
+    created_at: i64,
+
+    pub fn deinit(self: *PendingRPCEntry, allocator: Allocator) void {
+        self.request.deinit(allocator);
+    }
+};
+
+pub const PendingRPCMap = std.AutoHashMap(u64, PendingRPCEntry);
 // key: block root, value: depth
 pub const PendingBlockRootMap = std.AutoHashMap(types.Root, u32);
 // key: block root, value: pointer to block
@@ -64,6 +73,7 @@ pub const Network = struct {
     pending_block_roots: PendingBlockRootMap,
     fetched_blocks: FetchedBlockMap,
     fetched_block_children: ChildrenMap,
+    timed_out_requests: std.ArrayList(u64),
 
     const Self = @This();
 
@@ -94,10 +104,13 @@ pub const Network = struct {
             .pending_block_roots = pending_block_roots,
             .fetched_blocks = fetched_blocks,
             .fetched_block_children = fetched_block_children,
+            .timed_out_requests = std.ArrayList(u64).init(allocator),
         };
     }
 
     pub fn deinit(self: *Self) void {
+        self.timed_out_requests.deinit();
+
         var rpc_it = self.pending_rpc_requests.iterator();
         while (rpc_it.next()) |entry| {
             entry.value_ptr.deinit(self.allocator);
@@ -236,7 +249,7 @@ pub const Network = struct {
             defer request_ids_to_remove.deinit();
 
             while (rpc_it.next()) |rpc_entry| {
-                const pending_peer_id = switch (rpc_entry.value_ptr.*) {
+                const pending_peer_id = switch (rpc_entry.value_ptr.request) {
                     .status => |*ctx| ctx.peer_id,
                     .blocks_by_root => |*ctx| ctx.peer_id,
                 };
@@ -441,7 +454,10 @@ pub const Network = struct {
 
         const request_id = try self.sendStatus(peer_id, status, handler);
 
-        self.pending_rpc_requests.put(request_id, pending) catch |err| {
+        self.pending_rpc_requests.put(request_id, PendingRPCEntry{
+            .request = pending,
+            .created_at = std.time.timestamp(),
+        }) catch |err| {
             pending.deinit(self.allocator);
             return err;
         };
@@ -493,7 +509,10 @@ pub const Network = struct {
             return err;
         };
 
-        self.pending_rpc_requests.put(request_id, pending) catch |err| {
+        self.pending_rpc_requests.put(request_id, PendingRPCEntry{
+            .request = pending,
+            .created_at = std.time.timestamp(),
+        }) catch |err| {
             pending.deinit(self.allocator);
             return err;
         };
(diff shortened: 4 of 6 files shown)


chetanyb:
@g11tech ﻿﻿﻿﻿﻿Current design allows the observability API to briefly hold shared locks. Should we add rate limiting to prevent potential write starvation from excessive requests, or is the [max slot cap](https://github.com/blockblaz/zeam/pull/426/files#diff-8923b7591b9e2f4d59d63bb5844a69ce46e7aab282bd194771f58983511c0f7dR121) sufficient?

pkgs/cli/src/api_server.zig
```diff
@@ -2,6 +2,25 @@ const std = @import("std");
 const api = @import("@zeam/api");
 const constants = @import("constants.zig");
 const event_broadcaster = api.event_broadcaster;
+const node = @import("@zeam/node");
+
+// Global chain reference for API access
+var global_chain: ?*node.BeamChain = null;
```
g11tech: i dont like global chain, you should register the routes on the api server using the forkchoice instance, not even chain

pkgs/cli/src/api_server.zig
```diff
@@ -63,11 +82,235 @@ fn handleConnection(connection: std.net.Server.Connection, allocator: std.mem.Al
...
+        const role = if (is_finalized)
+            "finalized"
+        else if (is_justified)
+            "justified"
+        else if (is_head)
+            "head"
+        else if (is_orphaned)
+            "orphaned"
+        else
+            "normal";
```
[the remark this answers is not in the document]
chetanyb: Setting role in each block would scatter the logic and make priority less obvious. Keeping as-is.

pkgs/cli/src/api_server.zig
```diff
@@ -63,11 +82,235 @@ fn handleConnection(connection: std.net.Server.Connection, allocator: std.mem.Al
...
+
+        // Normalized weight for arc (0.0 to 1.0, draws partial circle border)
+        // Represents fraction of circle filled (0.5 = half circle, 1.0 = full circle)
+        const arc_weight: f64 = if (max_weight > 0)
+            @as(f64, @floatFromInt(pnode.weight)) / @as(f64, @floatFromInt(max_weight))
+        else
+            0.0;
```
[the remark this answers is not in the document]
chetanyb: uninitialized variables are undefined behavior, setting explicitly to 0.0 is required:

pkgs/cli/src/api_server.zig
```diff
@@ -2,47 +2,99 @@ const std = @import("std");
...
+        .thread = thread,
+        .ctx = ctx,
+    };
 }
 
-/// Handle individual HTTP connections in a separate thread
-fn handleConnection(connection: std.net.Server.Connection, allocator: std.mem.Allocator) void {
-    defer connection.stream.close();
-
+fn routeConnection(connection: std.net.Server.Connection, allocator: std.mem.Allocator, forkchoice: *node.fcFactory.ForkChoice, ctx: *SimpleMetricsServer) void {
```
g11tech: since ctx is already passed and has forkchoice, why forkchoice is needed as an arg

pkgs/cli/src/api_server.zig
```diff
@@ -2,47 +2,99 @@ const std = @import("std");
...
     if (std.mem.eql(u8, request.head.target, "/events")) {
-        // Handle SSE connection - this will keep the connection alive
-        SimpleMetricsServer.handleSSEEvents(connection.stream, allocator) catch |err| {
-            std.log.warn("SSE connection failed: {}", .{err});
+        if (!ctx.tryAcquireSSE()) {
+            _ = request.respond("Service Unavailable\n", .{ .status = .service_unavailable }) catch {};
+            connection.stream.close();
+            return;
+        }
+        _ = std.Thread.spawn(.{}, SimpleMetricsServer.handleSSEConnection, .{ connection.stream, allocator, ctx }) catch |err| {
```
g11tech: do we need to span a new thread?

pkgs/cli/src/api_server.zig
```diff
@@ -56,45 +108,284 @@ fn handleConnection(connection: std.net.Server.Connection, allocator: std.mem.Al
...
     } else {
         _ = request.respond("Not Found\n", .{ .status = .not_found }) catch {};
     }
 }
 
+fn handleForkChoiceGraph(
+    request: *std.http.Server.Request,
+    allocator: std.mem.Allocator,
+    forkchoice: *node.fcFactory.ForkChoice,
+    ctx: *SimpleMetricsServer,
```
g11tech: if ctx is passed, why need extra forkchoice arg

pkgs/node/src/chain.zig
```diff
@@ -402,6 +413,9 @@ pub const BeamChain = struct {
             \\+---------------------------------------------------------------+
             \\  Latest Justified:   Slot {d:>6} | Root: 0x{any}
             \\  Latest Finalized:   Slot {d:>6} | Root: 0x{any}
+            \\+---------------------------------------------------------------+
+            \\  ForkChoice Tree:
+            \\{s}
```
g11tech: we don't wanna print the tree here, but may be summary/analysis (which could be a followup PR)

KatyaRyazantseva:
The Grafana dashboard looks good. I've noticed some gaps between slots that might be related to scrape interval or Grafana lagging, but it's not critical.

tree

Should be tested on scenarios like losing peers, non-finality. From my side didn't look healthy. Not sure what the expectations are here. What do the numbers (2, 1) in the circle mean? Peers, votes? We can test the long-running behavior on devnets once this PR is released.

non-finality

Some dashboard improvements that could be done (or skipped):

- Colorize the finalized part of the chain (not only block) if possible.
- Make it more universal by 'Export to use at another instance' for those who have their own Grafana setup. With this, data source can be chosen on the Grafana side.
- Document dashboard setup details: yesoreyeram-infinity-datasource plugin needed, basic setup - Datasource, URL, Format, Rows.
- Adjust legend to show client names and readable fork choice statuses.

ch4r10t33r:
> * Client name mapping isn’t possible right now: the node only knows peer IDs, which aren’t tied to client identity. We can consider exposing an optional mapping in the spec later.

Can we consider to share this information via ReqResp after the initial handshake? @g11tech

pkgs/cli/src/api_server.zig
```diff
@@ -66,55 +175,32 @@ const ApiServer = struct {
...
-        } else if (std.mem.eql(u8, request.head.target, "/lean/states/finalized")) {
-            // Handle finalized checkpoint state endpoint
-            self.handleFinalizedCheckpointState(&request) catch |err| {
-                self.logger.warn("failed to handle finalized checkpoint state request: {}", .{err});
-                _ = request.respond("Internal Server Error\n", .{ .status = .internal_server_error }) catch {};
-            };
-        } else {
-            _ = request.respond("Not Found\n", .{ .status = .not_found }) catch {};
+        // Allow active SSE threads to drain before destroying context
+        while (true) {
```
g11tech: loop better written as while self.see_active !== 0 rather than while true

pkgs/cli/src/api_server.zig
```diff
@@ -8,17 +8,47 @@ const utils_lib = @import("@zeam/utils");
...
+const MAX_SSE_CONNECTIONS: usize = 32;
+const MAX_GRAPH_INFLIGHT: usize = 2;
+const RATE_LIMIT_RPS: f64 = 2.0;
+const RATE_LIMIT_BURST: f64 = 5.0;
+const RATE_LIMIT_MAX_ENTRIES: usize = 256; // Max tracked IPs to bound memory.
+const RATE_LIMIT_CLEANUP_THRESHOLD: usize = RATE_LIMIT_MAX_ENTRIES / 2; // Trigger lazy cleanup.
+const RATE_LIMIT_STALE_NS: u64 = 10 * std.time.ns_per_min; // Evict entries idle past TTL.
+const RATE_LIMIT_CLEANUP_COOLDOWN_NS: u64 = 60 * std.time.ns_per_s;
+
+pub const APIServerHandle = struct {
```
g11tech: a separate structure is not needed, why can't thread me moved to apiserver?

pkgs/cli/src/api_server.zig
```diff
@@ -66,61 +167,32 @@ const ApiServer = struct {
...
+        }) {
+            std.time.sleep(ACCEPT_POLL_NS);
         }
     }
 
     /// Handle metrics endpoint
-    fn handleMetrics(self: *const Self, request: *std.http.Server.Request) void {
-        var metrics_output = std.ArrayList(u8).init(self.allocator);
+    fn handleMetrics(_: *const Self, request: *std.http.Server.Request, allocator: std.mem.Allocator) void {
+        var metrics_output = std.ArrayList(u8).init(allocator);
```
anshalshukla: why did we replace self.allocator with a separate comment?

Issue #447, opened 2025-12-22
add test cases for forkchoice functions


g11tech:
set a mock forkchoice with a interesting tree (with forks, missed slots, orphans) and add test cases

- [ ] getCanonicalAncestorAtDepth
- [ ] getCanonicalityAnalysis

in forkchoiice

PR #448, opened 2025-12-22
Test forkchoice with mock tree, forks, missed slots, orphans

bomanaps:
closes https://github.com/blockblaz/zeam/issues/447

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -16,6 +16,18 @@ const zkvm_targets: []const zkvmTarget = &.{
 
 const ProverChoice = enum { dummy, risc0, openvm, all };
 
+fn setTestRunLabel(b: *Builder, run_step: *std.Build.Step.Run, name: []const u8) void {
+    run_step.step.name = b.fmt("test {s}", .{name});
+}
+
+fn setTestRunLabelFromCompile(b: *Builder, run_step: *std.Build.Step.Run, compile_step: *std.Build.Step.Compile) void {
+    const source_name = if (compile_step.root_module.root_source_file) |root_source|
+        root_source.getDisplayName()
+    else
+        compile_step.step.name;
+    setTestRunLabel(b, run_step, source_name);
+}
+
 // Add the glue libs to a compile target
 fn addRustGlueLib(b: *Builder, comp: *Builder.Step.Compile, target: Builder.ResolvedTarget, prover: ProverChoice) void {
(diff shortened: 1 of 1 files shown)


pkgs/node/src/forkchoice.zig
```diff
@@ -904,3 +924,327 @@ test "forkchoice block tree" {
...
+    //   G = 0x11 (slot 4, parent C)     <- FORK starts here
+    //   H = 0x22 (slot 6, parent G)
+    //   I = 0x33 (slot 7, parent H)
+    //
+    // Node indices in protoArray:
+    //   A=0, B=1, C=2, D=3, E=4, F=5, G=6, H=7, I=8
+    //
+    // ============================================================================
+
+    var arena_allocator = std.heap.ArenaAllocator.init(std.testing.allocator);
```
gballet: just use `std.testing.allocator` in tests, because the arena allocator is good for performance, but it won't tell you if one particular allocation isn't freed.

pkgs/node/src/forkchoice.zig
```diff
@@ -904,3 +924,327 @@ test "forkchoice block tree" {
...
+    defer arena_allocator.deinit();
+    const allocator = arena_allocator.allocator();
+
+    const mock_chain = try stf.genMockChain(allocator, 2, null);
+
+    const spec_name = try allocator.dupe(u8, "beamdev");
```
gballet: for example, here

pkgs/node/src/forkchoice.zig
```diff
@@ -532,14 +550,16 @@ pub const ForkChoice = struct {
...
+    /// Depth 0 returns the head itself. Traverses parent pointers (not slot arithmetic),
+    /// so missed slots don't affect depth counting. If depth exceeds chain length,
+    /// clamps to genesis.
     pub fn getCanonicalAncestorAtDepth(self: *Self, min_depth: usize) !ProtoBlock {
         var depth = min_depth;
         var current_idx = self.protoArray.indices.get(self.head.blockRoot) orelse return ForkChoiceError.InvalidHeadIndex;
-        if (current_idx < depth) {
-            current_idx = 0;
-            depth = 0;
-        }
```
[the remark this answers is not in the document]
bomanaps: the while loop's current_idx > 0 condition already handles this case, but I can still put it back

Issue #452, opened 2025-12-24
Add comprehensive test coverage for ForkChoice rebase function with attestation tracking validation


noopur23:
add test cases for the rebase function

1. Node Relationship Integrity
- [ ] **Parent pointers** are correctly updated for all remaining nodes
- [ ] **bestChild pointers** are correctly maintained/updated  
- [ ] **bestDescendant pointers** are correctly maintained/updated
- [ ] **Node indices** are properly remapped after removal of finalized nodes

2. Weight Preservation
- [ ] **Node weights** remain unchanged conceptually (same relative voting power)
- [ ] **Weight distribution** is preserved across the rebased tree
- [ ] **No weight leakage** occurs during the rebase operation
- [ ] **Weight arithmetic** is correct after index remapping

3. Attestation Vote Tracker Integrity
- [ ] **`attestations` HashMap** is correctly updated with new node indices
- [ ] **Vote targets** are remapped to correct post-rebase indices  
- [ ] **Vote counts** are preserved for all remaining nodes
- [ ] **Orphaned votes** (pointing to removed nodes) are properly handled
- [ ] **AttestationTracker** state remains consistent

4. Edge Cases
- [ ] **Rebase at genesis** (no nodes to remove)
- [ ] **Rebase with complex forks** (multiple branches affected)
- [ ] **Rebase with missed slots** (sparse node distribution)
- [ ] **Rebase with heavy attestation load** (many votes to remap)

_Originally posted by @g11tech in https://github.com/blockblaz/zeam/issues/451#issuecomment-3687692387_

PR #466, opened 2026-01-02
add comprehensive rebase tests for nodes weights and attestations

bomanaps:
closes https://github.com/blockblaz/zeam/issues/452


pkgs/state-transition/src/lib.zig
```diff
@@ -16,6 +16,7 @@ pub const verifySingleAttestation = transition.verifySingleAttestation;
 
 const mockImport = @import("./mock.zig");
 pub const genMockChain = mockImport.genMockChain;
+pub const MockChainData = mockImport.MockChainData;
```
g11tech: why is this required here?

Issue #471, opened 2026-01-07
Network Metrics


bomanaps:
https://github.com/leanEthereum/leanMetrics/pull/13

PR #472, opened 2026-01-07
add metrics for peer connection and disconnection events

bomanaps:
closes https://github.com/blockblaz/zeam/issues/471. https://github.com/leanEthereum/leanMetrics/pull/13

### DIFF
--- a/rust/libp2p-glue/src/lib.rs
+++ b/rust/libp2p-glue/src/lib.rs
@@ -1057,20 +1057,13 @@ impl Network {
                                 &format!("Outgoing connection failed: peer={} error={:?} result={}", peer_str, error, result),
                             );
 
-                            // Notify Zig of failed connection attempt
+                            // Notify Zig of failed connection attempt and handle reconnection
                             if let Some(pid) = peer_id {
                                 let peer_id_cstr = match CString::new(pid.to_string()) {
                                     Ok(cstr) => cstr,
                                     Err(_) => {
+                                        // Invalid peer_id string - can't communicate with Zig, don't retry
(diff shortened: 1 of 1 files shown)


pkgs/metrics/src/lib.zig
```diff
@@ -169,6 +177,28 @@ pub fn incrementLeanAttestationsInvalid(is_from_block: bool) void {
     };
 }
 
+/// Sets the current number of connected peers gauge
+pub fn setConnectedPeers(count: usize) void {
```
g11tech: why can't we directly use `metrics.lean_connected_peers.set`? 
it will automatically be a noop if its not initialized, and this is not going to get called from zkvm

similarly for others, I don't want extra wrapper functions encapsulating basic stuff

similarly for other metrics and even the older metrics, remove all the extra wrapper functions

rust/libp2p-glue/src/lib.rs
```diff
@@ -984,11 +1041,49 @@ impl Network {
...
                             );
 
+                            // Notify Zig of failed connection attempt
+                            if let Some(pid) = peer_id {
+                                let peer_id_cstr = match CString::new(pid.to_string()) {
+                                    Ok(cstr) => cstr,
+                                    Err(_) => {
+                                        logger::rustLogger.error(self.network_id, &format!("invalid_peer_id_string={}", pid));
+                                        // Continue to reconnection logic below
+                                        if let Some((addr, attempt)) = RECONNECT_ATTEMPTS
```
g11tech: if peer id is incorrect tgen why would we reattempt connection?

PR #479, opened 2026-01-12
hosting and consuming checkpoint sync server API

### LINKED ISSUE #469: serve finalized checkpoint state on an api and use it to checkpoint start the client
serve the checkpoint state in octect stream like how its done in beacon - use the similar api route to serve it
as well as use this to checkpoint start the client i.e. use this as an anchorstate to start the client 

(so  client should automatically sync to other peers using block by roots on the missing parents which is already the feature we have, but get it working if its not)

ch4r10t33r:
- Add `/lean/states/finalized` API endpoint to serve finalized checkpoint state as SSZ binary
- Add `--checkpoint-sync-url` CLI parameter to enable checkpoint sync from another zeam node
- Implement client-side checkpoint state download, verification, and anchor state initialization

**Motivation**
Closes #469

Checkpoint sync allows new zeam nodes to bootstrap from a trusted node's finalized state rather than syncing from genesis. This significantly reduces initial sync time for nodes joining the network.

Attached screenshot demonstrating the working of the endpoint
Screenshot 2026-01-13 at 18 37 30
Screenshot 2026-01-13 at 18 37 52

### DIFF
--- a/pkgs/network/src/ethlibp2p.zig
+++ b/pkgs/network/src/ethlibp2p.zig
@@ -1,5 +1,4 @@
 const std = @import("std");
-const json = std.json;
 const Allocator = std.mem.Allocator;
 const Thread = std.Thread;
 
@@ -10,7 +9,6 @@ const multiformats = @import("multiformats");
 const Multiaddr = multiformats.multiaddr.Multiaddr;
 const uvarint = multiformats.uvarint;
 const zeam_utils = @import("@zeam/utils");
-const jsonToString = zeam_utils.jsonToString;
 
 const interface = @import("./interface.zig");
 const NetworkInterface = interface.NetworkInterface;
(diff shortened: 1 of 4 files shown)


pkgs/cli/src/api_server.zig
```diff
@@ -2,43 +2,121 @@ const std = @import("std");
...
 /// Simple metrics server that runs in a background thread
-pub fn startAPIServer(allocator: std.mem.Allocator, port: u16) !void {
-    // Initialize the global event broadcaster
+pub fn startAPIServer(allocator: std.mem.Allocator, port: u16, logger_config: *LoggerConfig) !void {
+    // Initialize the global event broadcaster for SSE events
+    // This is idempotent - safe to call even if already initialized elsewhere (e.g., node.zig)
     try event_broadcaster.initGlobalBroadcaster(allocator);
 
+    // Create a logger instance for the API server
+    const logger = logger_config.logger(.metrics);
```
g11tech:
```suggestion
    const logger = logger_config.logger(.api_server);
```
so add a new child module in util/log

pkgs/cli/src/node.zig
```diff
@@ -184,11 +187,25 @@ pub const Node = struct {
...
+            try verifyCheckpointStateRoot(allocator, self.anchor_state, self.logger);
+
+            self.logger.info("checkpoint sync completed successfully, using state at slot {d} as anchor", .{self.anchor_state.slot});
+        } else {
+            // Try to load the latest finalized state from the database, fallback to genesis
+            db.loadLatestFinalizedState(self.anchor_state) catch |err| {
+                self.logger.warn("failed to load latest finalized state from database: {any}", .{err});
+                try self.anchor_state.genGenesisState(allocator, chain_config.genesis);
+            };
+        }
```
anshalshukla: I think it's better to put it in try block and if the checkpoint download fail we can log and fall back to load from db/genesis. RIght now if the remote node doesn't server the node will crash

pkgs/cli/src/node.zig
```diff
@@ -491,6 +513,192 @@ pub fn buildStartOptions(
...
+    const authority = after_scheme[0..path_start];
+    const path = after_scheme[path_start..];
+
+    // Split authority into host:port
+    const colon_pos = std.mem.indexOfScalar(u8, authority, ':');
+    const host = if (colon_pos) |pos| authority[0..pos] else authority;
+    const port_str = if (colon_pos) |pos| authority[pos + 1 ..] else null;
+    const port: u16 = if (port_str) |p_str|
+        try std.fmt.parseInt(u16, p_str, 10)
+    else if (std.mem.eql(u8, scheme, "https")) @as(u16, 443) else @as(u16, 80);
```
anshalshukla: We should use std library uri parser instead of manually doing it. If there is something specific that is requires manual parsing, maybe add a helper function that can be reused

pkgs/cli/src/node.zig
```diff
@@ -491,6 +513,192 @@ pub fn buildStartOptions(
...
+    const address = try std.net.Address.resolveIp(host, port);
+
+    // Connect to server
+    var stream = try std.net.tcpConnectToAddress(address);
+    defer stream.close();
+
+    // Build HTTP request
+    const request = try std.fmt.allocPrint(
+        allocator,
+        "GET {s} HTTP/1.1\r\n" ++
```
anshalshukla: We should use std.http.Client here

pkgs/cli/src/node.zig
```diff
@@ -491,6 +513,192 @@ pub fn buildStartOptions(
...
+    var headers_complete = false;
+    var content_length: ?usize = null;
+    var content_start: usize = 0;
+
+    while (!headers_complete) {
+        const bytes_read = try stream.read(buffer[total_read..]);
+        if (bytes_read == 0) break;
+        total_read += bytes_read;
+
+        // Look for end of headers (CRLF CRLF)
```
anshalshukla: does std http client not support this? looks like standard operations that default client should support i believe

anshalshukla (changes requested):

anshalshukla:
LGTM!

pkgs/cli/src/node.zig
```diff
@@ -212,6 +233,12 @@ pub const Node = struct {
             .logger_config = options.logger_config,
             .node_registry = options.node_registry,
         });
+
+        // Start API server after chain is initialized so we can pass the chain pointer
+        if (options.metrics_enable) {
```
g11tech: a bit weird to have metrics enabled flag to sart api server, create an issue for later cleanup

ch4r10t33r: Yes, I have a separate issue for this. https://github.com/blockblaz/zeam/issues/486

pkgs/cli/src/node.zig
```diff
@@ -600,32 +600,46 @@ fn downloadCheckpointState(
...
 }
 
-/// Verifies and logs checkpoint state information
-/// Computes the state root and block root for the downloaded state
-/// Note: The actual checkpoint verification happens during forkchoice initialization
-/// where the block root is compared against the anchor block root
+/// Verifies checkpoint state against the genesis configuration
+/// Validates that the downloaded state is consistent with expected genesis parameters
+/// Also computes and logs the state root and block root
 fn verifyCheckpointStateRoot(
```
g11tech:
```suggestion
fn verifyCheckpointState(
```

Issue #460, opened 2025-12-30
Add sim test to test parent block syncing


anshalshukla:
Originally addressed [here](https://github.com/blockblaz/zeam/pull/368#pullrequestreview-3614577865) by @g11tech

anshalshukla:
Currently in `"CLI beam command with mock network - complete integration test"` we have a two node setup which runs two nodes and wait for them to achieve finalization. Let's modify this setup to be a 4 node setup and spin up 3 node in the beginning, as soon as the 3 node setup reaches finalization, we should start the 4th node. Once all the 4 nodes successfully reach another finalized slot we can consider it to be successful.

PR #484, opened 2026-01-13
add 3-node parent sync integration test for networking

bomanaps:
closes https://github.com/blockblaz/zeam/issues/460

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -22,8 +22,8 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       version: ${{ steps.get_tags_labels.outputs.version }}
-      zeam_tag: ${{ steps.get_tags_labels.outputs.zeam_tag }}
-      has_network_tag: ${{ steps.get_tags_labels.outputs.has_network_tag }}
+      devnet_tag: ${{ steps.get_tags_labels.outputs.devnet_tag }}
+      has_devnet_tag: ${{ steps.get_tags_labels.outputs.has_devnet_tag }}
 
     steps:
     - name: Checkout code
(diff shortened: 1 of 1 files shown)


pkgs/cli/src/main.zig
```diff
@@ -323,8 +323,9 @@ fn mainInner() !void {
 
             // Create key manager FIRST to get validator pubkeys for genesis
             const key_manager_lib = @import("@zeam/key-manager");
-            // Using 3 validators: so by default beam cmd command runs two nodes to interop
-            const num_validators: usize = 3;
+            // Using 4 validators for 4-node setup with initial sync testing
+            // Nodes 1,2,3 start immediately; Node 4 starts after finalization to test sync
+            const num_validators: usize = 4;
```
g11tech: why do we need 4?

pkgs/cli/test/integration.zig
```diff
@@ -566,6 +566,111 @@ test "SSE events integration test - wait for justification and finalization" {
     std.debug.print("SUCCESS: SSE events integration test completed\n", .{});
 }
 
+test "4-node initial sync integration test - node 4 syncs after finalization" {
```
g11tech:
```suggestion
test "3-node initial sync integration test - node 3 syncs after finalization" {
```

pkgs/cli/src/main.zig
```diff
@@ -503,9 +515,99 @@ fn mainInner() !void {
                 .node_registry = registry_2,
             });
 
+            // Node 3 setup - delayed start for initial sync testing
+            // This node starts after nodes 1,2 reach finalization and will sync from peers
+            // We use a DelayedNodeStarter to start node 3 after finalization is reached
+            const DelayedNodeStarter = struct {
```
g11tech: why do we need all this, why can't we just create a beam node 3 outside and assign it in the `beam_node`  here and in checkAndStart just call `beam_node.run()` when its time to start the node3

pkgs/cli/test/integration.zig
```diff
@@ -566,6 +566,111 @@ test "SSE events integration test - wait for justification and finalization" {
     std.debug.print("SUCCESS: SSE events integration test completed\n", .{});
 }
 
+test "3-node initial sync integration test - node 3 syncs after finalization" {
```
g11tech: why do we need to add a new test, can't the old integration test be just updated,
as I can see anytime we run beam command the third node will be delayed started

pkgs/node/src/node.zig
```diff
@@ -1369,6 +1395,7 @@ test "Node: pruneCachedBlockSubtree prunes cached descendants even if root is no
 
     var ctx = try testing.NodeTestContext.init(allocator, .{});
     defer ctx.deinit();
+    ctx.fixupDbLogger(); // Fix dangling logger pointer after struct has stable address
```
g11tech: why isn't this addressed in the init itself?

PR #490, opened 2026-01-16
rebase justifications in state from finalized slot

### LINKED ISSUE #450: Fix justifications unpacking/repacking logic to handle finalized slot boundaries in state transition
the justifications pruning in state in spec, so while packing in the state, the justifications and justification slots start from finalized slot + 1,

so this info needs to be used to unpack justifications from the parent state using parent's finalized, and needs to be repacked into the post state that is being constructed in the STF using the post state's finalized (updated if new finalization)

GrapeBaBa:
Fix #450


pkgs/types/src/state.zig
```diff
@@ -252,17 +286,21 @@ pub const BeamState = struct {
             self.latest_finalized.root = staged_block.parent_root;
         }
 
-        // extend historical block hashes and justified slots structures using SSZ Lists directly
+        // extend historical block hashes using SSZ Lists directly
         try self.historical_block_hashes.append(staged_block.parent_root);
-        // if parent is genesis it is already justified
-        try self.justified_slots.append(if (self.latest_block_header.slot == 0) true else false);
```
g11tech: also this is not replicated in the extension because parent is not marked true if its genesis.

as noted in the below comment, we don't need to do extend checks, we can simply append because finalized has not moved yet and these operations are append to the current structure which is already with current finalized as base

pkgs/types/src/state.zig
```diff
@@ -252,17 +286,21 @@ pub const BeamState = struct {
...
+        const finalized_slot: Slot = self.latest_finalized.slot;
+        const last_materialized_slot: Slot = staged_block.slot - 1;
+        try self.extendJustifiedSlots(finalized_slot, last_materialized_slot);
 
         const block_slot: usize = @intCast(staged_block.slot);
         const missed_slots: usize = @intCast(block_slot - self.latest_block_header.slot - 1);
         for (0..missed_slots) |i| {
             _ = i;
             try self.historical_block_hashes.append(utils.ZERO_HASH);
-            try self.justified_slots.append(false);
```
g11tech: restore as noted in above comments, also restore in spec if modified there

pkgs/types/src/state.zig
```diff
@@ -252,18 +286,17 @@ pub const BeamState = struct {
...
         try self.historical_block_hashes.append(staged_block.parent_root);
-        // if parent is genesis it is already justified
-        try self.justified_slots.append(if (self.latest_block_header.slot == 0) true else false);
 
         const block_slot: usize = @intCast(staged_block.slot);
         const missed_slots: usize = @intCast(block_slot - self.latest_block_header.slot - 1);
         for (0..missed_slots) |i| {
             _ = i;
             try self.historical_block_hashes.append(utils.ZERO_HASH);
-            try self.justified_slots.append(false);
```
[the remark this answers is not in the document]
GrapeBaBa: already covered in extend

pkgs/types/src/state.zig
```diff
@@ -180,6 +180,40 @@ pub const BeamState = struct {
...
+        const base: Slot = finalized_slot + 1;
+        const relative_index: Slot = target_slot - base;
+        const required_capacity: Slot = relative_index + 1;
+        const current_len: Slot = @intCast(self.justified_slots.len());
+        if (required_capacity <= current_len) {
+            return;
+        }
+        const gap_size: Slot = required_capacity - current_len;
+        var i: Slot = 0;
+        while (i < gap_size) : (i += 1) {
```
g11tech: this is better written as for loop

pkgs/types/src/state.zig
```diff
@@ -317,6 +350,30 @@ pub const BeamState = struct {
         errdefer justifications.deinit(allocator);
         try self.getJustification(allocator, &justifications);
 
+        var finalized_slot: Slot = self.latest_finalized.slot;
+
+        var root_to_slots: std.AutoHashMapUnmanaged(Root, std.ArrayListUnmanaged(Slot)) = .empty;
```
g11tech: move this block of code of filling root to slots to a separate fn, also create a PR for adding/maintaining/pruning  a global optional cache (because prover will not pass a cache and hence the cache will need to be constructed there) which gets passed to the process block since this mapping will stay same across blocks

pkgs/types/src/state.zig
```diff
@@ -420,15 +472,48 @@ pub const BeamState = struct {
...
+                                    if (slot_value > finalized_slot) {
+                                        keep = true;
+                                        break;
+                                    }
+                                }
+                                if (!keep) {
+                                    try roots_to_remove.append(entry.key_ptr.*);
+                                }
+                            } else {
+                                try roots_to_remove.append(entry.key_ptr.*);
```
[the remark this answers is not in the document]
g11tech: why do missed slots duplicate the root? we only push zero hashes when there is missed slot, and need to make sure zero hash is not added in justifications in the first case

PR #91, opened 2025-08-24
fix:fix build error on macOS since missing link CoreFoundation

GrapeBaBa:
This pull request makes platform-specific improvements to the build process for macOS and includes a minor cleanup in the `lib.zig` file. The primary focus is on ensuring that the necessary macOS frameworks are linked for both the main CLI executable and the CLI tests.

Build system improvements for macOS compatibility:

* [`build.zig`](diffhunk://#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR159-R161): Added logic to link the `CoreFoundation` framework when building for macOS, both for the main CLI executable and for CLI tests, ensuring compatibility with macOS system libraries. [[1]](diffhunk://#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR159-R161) [[2]](diffhunk://#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR223-R225)

Code cleanup:

* [`pkgs/state-transition-runtime/src/risc0/lib.zig`](diffhunk://#diff-a2111c0b4d4fcad243ca6d490e472aaa5ae006820f049bb9b583eac4b85cf5cdL76): Removed an unnecessary blank line for minor code tidiness.

PR #493, opened 2026-01-16
feat: Add OCI labels for git commit and branch to docker images

KatyaRyazantseva:
Adds OCI  labels to both Dockerfile and Dockerfile.prebuilt for better image traceability:
- `org.opencontainers.image.ref.name` - git branch
- `org.opencontainers.image.revision` - git commit

Labels are set via build args with `unknown` as default, so existing builds continue to work.

Updated README with build example for publishing to registries.

Build with labels:
```bash
docker build -f Dockerfile.prebuilt \
  --build-arg GIT_COMMIT=$(git rev-parse HEAD) \
  --build-arg GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD) \
  -t zeam:local .
```

Related to #91

Screenshot 2026-01-16 at 17 03 14

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -225,6 +225,7 @@ jobs:
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
         build-args: |
+          GIT_COMMIT=${{ github.sha }}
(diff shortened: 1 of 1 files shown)


g11tech:
you will need to modify our docker image release workflow as well cc @noopur23 

cc @gballet any opinions here?

PR #494, opened 2026-01-17
implement leanVM aggregation for devnet2

anshalshukla:
- Implements signature aggregation and makes zeam devnet2 ready
- Makes typing explicit in hashsig bindings

Aggregation Algorithm:
Three phase algorithm to aggregate attestation
 1. Find all the signatures in signature_map/gossip_signatures_map and return missing validators
 2. Find missing validators in aggregated_payloads
 3. Remove validators already covered by aggregated_payloads from signatures_map before actually aggregating


pkgs/key-manager/src/lib.zig
```diff
@@ -141,6 +130,36 @@ pub const KeyManager = struct {
...
+        return keypair.public_key;
+    }
+
+    /// Sign an attestation and return the raw signature handle (for aggregation)
+    /// Caller must call deinit on the returned signature when done
+    pub fn signAttestationWithHandle(
+        self: *const Self,
+        attestation: *const types.Attestation,
+        allocator: Allocator,
+    ) !xmss.Signature {
```
g11tech: return is not a pointer?

pkgs/node/src/chain.zig
```diff
@@ -766,6 +790,18 @@ pub const BeamChain = struct {
...
+
+            // Prune gossip_signatures and aggregated_payloads for finalized attestations
+            self.forkChoice.pruneSignatureMaps(latest_finalized.slot) catch |err| {
+                self.module_logger.warn("failed to prune signature maps: {any}", .{err});
+            };
+        }
+
+        // Store aggregated payloads from the block for future block building
+        if (signedBlock) |block| {
+            self.storeAggregatedPayloads(block) catch |err| {
```
g11tech: this could be stored right when chain did the aggregation right?

pkgs/types/src/block.zig
```diff
@@ -297,92 +302,337 @@ pub fn createBlockSignatures(allocator: Allocator, num_aggregated_attestations:
...
 
-pub fn aggregateSignedAttestations(
-    allocator: Allocator,
-    signed_attestations: []const attestation.SignedAttestation,
-) !AggregatedAttestationsResult {
-    var aggregated_attestations = try AggregatedAttestations.init(allocator);
-    errdefer aggregated_attestations.deinit();
+        // Process each group
+        for (groups.items) |*group| {
+            const data_root = group.data_root;
```
g11tech: would it be usedful to move all the signature related code/tests to a new file blocksignatures?

anshalshukla: I'll do it to make it less cluttered as it will continue to become more cluttered with subsequent devnets/ changes.

Issue #491, opened 2026-01-16
leanMetrics: node lifecycle, sync status, reorgs, and finalization


bomanaps:
Full details - https://github.com/leanEthereum/leanMetrics/pull/16

PR #495, opened 2026-01-19
add node lifecycle and reorg metrics

bomanaps:
closes https://github.com/blockblaz/zeam/issues/491

### DIFF
--- a/pkgs/node/src/forkchoice.zig
+++ b/pkgs/node/src/forkchoice.zig
@@ -799,10 +799,14 @@ pub const ForkChoice = struct {
 
         // Detect reorg: if head changed and previous head is not an ancestor of new head
         if (!std.mem.eql(u8, &self.head.blockRoot, &previous_head.blockRoot)) {
-            const is_extension = self.isAncestorOf(previous_head.blockRoot, self.head.blockRoot);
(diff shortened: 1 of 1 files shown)


pkgs/node/src/forkchoice.zig
```diff
@@ -750,15 +750,89 @@ pub const ForkChoice = struct {
...
+                });
+            }
+        }
+
         return self.head;
     }
 
+    /// Checks if potential_ancestor is an ancestor of descendant by walking up parent chain.
+    fn isAncestorOf(self: *Self, potential_ancestor: types.Root, descendant: types.Root) bool {
+        var current_idx = self.protoArray.indices.get(descendant) orelse return false;
```
g11tech: if you can't find descndant shouldn't it be an error/panic condition because it should defintely be there

pkgs/node/src/forkchoice.zig
```diff
@@ -750,15 +750,89 @@ pub const ForkChoice = struct {
...
+
+        while (true) {
```
g11tech: while true are considered very bad loop conditions, you should do while(current_node.parent) directly for better logic

g11tech (approved):

Issue #496, opened 2026-01-20
cleanup zeam log where we are printing raw structures


g11tech:
<img width="1805" height="229" alt="Image" src="https://github.com/user-attachments/assets/4ddfe471-a7d4-4032-937f-afe260401197" />

print meaningful log instead of above wherever we are printing raw structures. it was used for debugging, but we can now print more meaningful log

cc @noopur23

PR #502, opened 2026-01-22
replace raw struct dumps with meaningful logs

bomanaps:
Cleans up debug logging by replacing raw structure dumps that printed memory addresses, allocators, and vtables with meaningful values like slot numbers, hex-formatted roots, and counts. 
Fixes https://github.com/blockblaz/zeam/issues/496

Screenshot 2026-01-22 at 03 09 02

Screenshot 2026-01-22 at 03 39 31


g11tech:
its better to define a format function for these structures (like ZeamArgs) not just eliminate them from the print so that we have meaningful log

PR #532, opened 2026-01-28
fix: lock sig maps to avoid race condition

### LINKED ISSUE #520: zeam crash on invalid mem access
Image

in one of the runs i came across this (so its random)

debug/figureout/fix

anshalshukla:
closes #520


g11tech (approved):
lgtm

Issue #497, opened 2026-01-20
Add global optional cache for root-to-slots mapping to optimize block processing performance


noopur23:
adding/maintaining/pruning  a global optional cache (because prover will not pass a cache and hence the cache will need to be constructed there) which gets passed to the process block since this mapping will stay same across blocks

_Originally posted by @g11tech in https://github.com/blockblaz/zeam/pull/490#discussion_r2707203865_

noopur23:
fixed in https://github.com/blockblaz/zeam/pull/540

PR #540, opened 2026-01-31
add root to slot caching

chetanyb:
addresses https://github.com/blockblaz/zeam/issues/497

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -220,7 +220,7 @@ pub const BeamChain = struct {
     /// processing queued blocks; the caller owns and must free the slice.
     pub fn processPendingBlocks(self: *Self) []types.Root {
         var all_missing_roots: std.ArrayListUnmanaged(types.Root) = .empty;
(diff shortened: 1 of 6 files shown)


pkgs/types/src/state.zig
```diff
@@ -375,9 +382,22 @@ pub const BeamState = struct {
...
+                const start_slot: usize = @intCast(finalized_slot + 1);
+                const historical_len_usize: usize = self.historical_block_hashes.len();
+                for (start_slot..historical_len_usize) |i| {
+                    const root = try self.historical_block_hashes.get(i);
+                    if (std.mem.eql(u8, &root, &utils.ZERO_HASH)) continue;
+                    try c.put(root, @intCast(i));
+                }
+            }
+        } else {
+            try self.fillRootToSlot(allocator, finalized_slot, &root_to_slot);
```
g11tech: we don't need to fill in the state each time, 
when we init the chain, the we need to fill it, and then every time we add block in the chain/forkchoice then we add it, that will always cover this because we never process a block who's parent we don't have in forkchoice

so on each block we only need to add once

PR #543, opened 2026-02-03
upgrade to zig 0.15.2

anshalshukla:


pkgs/api/src/event_broadcaster.zig
```diff
@@ -26,7 +26,10 @@ pub const SSEConnection = struct {
         self.mutex.lock();
         defer self.mutex.unlock();
 
-        try self.stream.writeAll(event_json);
+        var write_buf: [4096]u8 = undefined;
```
gballet: allocate (and free) memory, there's an allocator available in `self`.

anshalshukla: I've changed it to allocate, 512 bytes instead of 4096 since it is for sse events, I think it is better to do a stack allocation here instead of heap allocation. I've also added testcases to make sure that the write_buf is sufficiently large to handle the event json even if new events are added.

README.md
```diff
@@ -85,11 +87,11 @@ docker build -f Dockerfile.prebuilt \
 
 #### Prerequisites
 
- - Zeam requires zig version 0.14.1 to build.
- - Zeam requires Rust 1.85+ to be able to build the rust connections to various zkvms
- - Both [risc0](https://github.com/risc0/risc0) (v3.0.3) and OpenVM provers are supported:
-   - risc0 requires the toolchain to be [installed](https://dev.risczero.com/api/zkvm/install): `rzup install r0vm 3.0.3`
-   - OpenVM is self-contained (no external toolchain needed)
+- Zeam requires zig version 0.15.2 to build.
```
gballet: this is the only thing that needs to change in this PR, the rest isn't related to this change and should be reverted

build/risc0.zig
```diff
@@ -22,21 +22,23 @@ pub fn main() !void {
 
         const file = try std.fs.cwd().createFile(output_path, .{ .truncate = true });
         defer file.close();
-        const writer = file.writer();
 
-        // magic +  binary format
-        _ = try writer.write(magic);
-        try writer.writeInt(u32, BinaryFormatVersion, .little);
+        var write_buf: [4096]u8 = undefined;
```
gballet: that doesn't look correct, I don't think that the whole binary file can fit in 4K

pkgs/state-transition-runtime/src/powdr/io.zig
gballet: this change is frankly weird, was it needed or done by AI?

anshalshukla: It's needed because of the change in writer interface. I'll use an allocating writer here instead.

gballet (approved):
LGTM. We agreed with Anshal that this was breaking the risc0 prover, but I will investigate the topic as I am doing heavy editing for provers anyway.

PR #548, opened 2026-02-05
fix: add timeout to peer requests

anshalshukla:
- When sendBlocksByRootRequest() sends a request and the peer never responds (and never disconnects), roots stay in pending_block_roots forever. shouldRequestBlocksByRoot() returns false, permanently blocking re-fetch. This PR introduces timeout of 8 seconds and retries the same block again after timeout.
- getSyncStatus() only checks our_head_slot < max_peer_finalized_slot. After restart at head=1592 with peers at finalized=1568/head=1601, the guard reports .synced and allows block production on stale state. Our head is past finalization, but our own finalization (e.g. 1536) is behind the peer’s meaning we are on a divergent fork. This PR adds finalization comparison between peers and itsef.

### DIFF
--- a/pkgs/node/src/network.zig
+++ b/pkgs/node/src/network.zig
@@ -73,6 +73,7 @@ pub const Network = struct {
     pending_block_roots: PendingBlockRootMap,
     fetched_blocks: FetchedBlockMap,
     fetched_block_children: ChildrenMap,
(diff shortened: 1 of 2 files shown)


ch4r10t33r (approved):
Looks good.

PR #552, opened 2026-02-07
feat: implement devnet3 committee aggregation spec

GrapeBaBa:


pkgs/cli/src/main.zig
```diff
@@ -58,6 +58,7 @@ pub const NodeCommand = struct {
     @"network-dir": []const u8 = "./network",
     @"data-dir": []const u8 = constants.DEFAULT_DATA_DIR,
     @"checkpoint-sync-url": ?[]const u8 = null,
+    @"is-aggregator": bool = false,
```
g11tech: aggregator is based on a validator id being aggregator not the node
but probably what can be done is that if a node is indicated to be an aggregator, then one of the validator ids can be switched as aggregator

pkgs/types/src/utils.zig
```diff
@@ -31,6 +31,15 @@ pub const StateTransitionError = error{ InvalidParentRoot, InvalidPreState, Inva
 
 const json = std.json;
 
+pub const SubnetIdError = error{InvalidCommitteeCount};
+
+pub fn computeSubnetId(validator_id: ValidatorIndex, committee_count: u64) SubnetIdError!u32 {
+    if (committee_count == 0 or committee_count > std.math.maxInt(u32)) {
+        return error.InvalidCommitteeCount;
+    }
+    return @intCast(validator_id % committee_count);
```
g11tech: double check this @ch4r10t33r just for second pair of eyes

pkgs/node/src/validator_client.zig
```diff
@@ -212,9 +226,66 @@ pub const ValidatorClient = struct {
...
+                self.logger.warn("skipping aggregation production for slot={d}: behind peers (head_slot={d} < max_peer_finalized_slot={d})", .{
+                    slot,
+                    info.head_slot,
+                    info.max_peer_finalized_slot,
+                });
+                return null;
+            },
+        }
+
+        const aggregations = self.chain.aggregateCommitteeSignatures() catch |err| {
```
g11tech: ahh ok so a "node" is chosen as an aggregator, and this needs to work as peer-das custody to make sure their are enough aggregators, so based on validator ids, one would need to assign the subnet

g11tech: in that case do we even need this bit here? could just be handled purely at node oninterval

also question to be asked for spec:
how much redundancy is to be modeled for nodes aggregating on the same subnet cc @anshalshukla

pkgs/network/src/interface.zig
```diff
@@ -194,32 +198,83 @@ pub const LeanNetworkTopic = struct {
...
-pub const GossipTopic = enum {
+pub const GossipTopicKind = enum {
     block,
     attestation,
+    aggregation,
+};
 
-    pub fn encode(self: GossipTopic) []const u8 {
-        return std.enums.tagName(GossipTopic, self).?;
+pub const GossipTopic = struct {
```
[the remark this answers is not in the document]
GrapeBaBa: because `attestation` topic right now include subnet_id arg

pkgs/node/src/utils.zig
```diff
@@ -41,3 +42,7 @@ pub const OnIntervalCbWrapper = struct {
         return self.onIntervalCb(self.ptr, self.interval);
     }
 };
+
+pub fn computeSubnetId(validator_id: u64, committee_count: u64) error{InvalidCommitteeCount}!u32 {
```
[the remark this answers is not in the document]
anshalshukla: and u32 for committee_count, maybe we can create a type alias similar to ValidatorIndex if required to avoid u32/u64 confusion.

pkgs/network/src/interface.zig
```diff
@@ -194,32 +198,83 @@ pub const LeanNetworkTopic = struct {
...
+    kind: GossipTopicKind,
+    subnet_id: ?u32 = null,
+
+    pub fn encode(self: GossipTopic, allocator: Allocator) ![]u8 {
+        return switch (self.kind) {
+            .block => allocator.dupe(u8, "block"),
+            .aggregation => allocator.dupe(u8, "aggregation"),
+            .attestation => blk: {
+                const subnet_id = self.subnet_id orelse return error.MissingSubnetId;
+                break :blk std.fmt.allocPrint(allocator, "attestation_{d}", .{subnet_id});
```
anshalshukla: we can use `@tagName(self.kind)`

pkgs/network/src/interface.zig
```diff
@@ -194,32 +198,83 @@ pub const LeanNetworkTopic = struct {
...
-    pub fn encode(self: GossipTopic) []const u8 {
-        return std.enums.tagName(GossipTopic, self).?;
+pub const GossipTopic = struct {
+    kind: GossipTopicKind,
+    subnet_id: ?u32 = null,
+
+    pub fn encode(self: GossipTopic, allocator: Allocator) ![]u8 {
+        return switch (self.kind) {
+            .block => allocator.dupe(u8, "block"),
+            .aggregation => allocator.dupe(u8, "aggregation"),
```
anshalshukla: https://github.com/leanEthereum/leanSpec/blob/b39472e73f8a7d603cc13d14426eed14c6eff6f1/src/lean_spec/subspecs/networking/gossipsub/topic.py#L157

the encoding has been updated, please update it here as well

pkgs/node/src/chain.zig
```diff
@@ -1013,6 +1060,74 @@ pub const BeamChain = struct {
...
+    }
+
+    fn cachePendingAttestation(self: *Self, signed_attestation: networks.AttestationGossip) void {
+        self.pending_gossip_attestations.append(self.allocator, signed_attestation) catch |err| {
+            self.module_logger.warn("failed to cache pending attestation: {any}", .{err});
+        };
+    }
+
+    fn cachePendingAggregation(self: *Self, signed_aggregation: types.SignedAggregatedAttestation) void {
+        var cloned_proof: types.AggregatedSignatureProof = undefined;
```
g11tech: why sszing cloning inside here and not in cachePendingAttestation, (similarly for the pending blocks), unify the approach considering cloning could take time in which other threads can race and create conditions where this can be imported i.e. clone first check again and then finally add them if still not importable

pkgs/node/src/chain.zig
```diff
@@ -1013,6 +1060,74 @@ pub const BeamChain = struct {
...
+        self.pending_gossip_aggregations.append(self.allocator, .{
+            .data = signed_aggregation.data,
+            .proof = cloned_proof,
+        }) catch |err| {
+            self.module_logger.warn("failed to cache pending aggregation: {any}", .{err});
+            cloned_proof.deinit();
+        };
+    }
+
+    fn retryPendingAttestationGossip(self: *Self) void {
```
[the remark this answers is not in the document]
g11tech: so basically create a PR on top of this branch/PR and move retrying logic there

pkgs/node/src/validator_client.zig
```diff
@@ -80,8 +100,14 @@ pub const ValidatorClient = struct {
         switch (interval) {
             0 => return self.maybeDoProposal(slot),
             1 => return self.mayBeDoAttestation(slot),
-            2 => return null,
+            2 => {
```
g11tech: should happen in the node's oninterval directly

pkgs/node/src/forkchoice.zig
```diff
@@ -846,26 +887,22 @@ pub const ForkChoice = struct {
     // Internal unlocked version - assumes caller holds lock
     fn getProposalAttestationsUnlocked(self: *Self) ![]types.Attestation {
         var included_attestations: std.ArrayList(types.Attestation) = .empty;
+        if (self.latest_known_aggregated_payloads.count() == 0) {
+            return included_attestations.toOwnedSlice(self.allocator);
+        }
 
-        const latest_justified = self.fcStore.latest_justified;
+        var attestation_map = try self.extractAttestationsFromAggregatedPayloads(&self.latest_known_aggregated_payloads);
```
g11tech: undo, extractAttestationsFromAggregatedPayloads directly when importing, why should we delay importing these onto pur attestations tracker

pkgs/node/src/chain.zig
```diff
@@ -874,15 +911,15 @@ pub const BeamChain = struct {
...
-                        zeam_metrics.metrics.lean_attestations_invalid_total.incr(.{ .source = "block" }) catch {};
-                        self.module_logger.err("error processing block attestation={f} error={any}", .{ attestation, e });
-                        continue;
-                    };
                     zeam_metrics.metrics.lean_attestations_valid_total.incr(.{ .source = "block" }) catch {};
                 }
             }
 
+            // Store aggregated payloads from the block before updating head
+            self.storeAggregatedPayloads(&signedBlock) catch |err| {
```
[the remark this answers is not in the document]
g11tech: yes, matched the {} :+1:

PR #576, opened 2026-02-18
fix: initialize forkchoice finalized/justified checkpoints from anchor state

ch4r10t33r:
## Summary

- Fixes `ForkChoice.init` to use the anchor state's actual `latest_justified` and `latest_finalized` checkpoints instead of a synthetic checkpoint derived only from the anchor slot.
- When using `--checkpoint-sync-url`, the trusted finalized/justified state from the remote peer is now correctly propagated to the fork choice store.
- Adds logging at both the node and forkchoice layers to distinguish genesis initialization from checkpoint sync, including finalized/justified slot and root details.

## Test plan

- [ ] Start a node with `--checkpoint-sync-url` pointing at a synced peer and verify logs show correct finalized/justified slots from the remote state
- [ ] Start a node without `--checkpoint-sync-url` and verify logs show genesis initialization with slot 0
- [ ] Verify the checkpoint-synced node correctly tracks finalization from the synced slot forward

### DIFF
--- a/pkgs/node/src/forkchoice.zig
+++ b/pkgs/node/src/forkchoice.zig
@@ -326,24 +326,20 @@ pub const ForkChoice = struct {
             .confirmed = true,
         };
         const proto_array = try ProtoArray.init(allocator, anchor_block);
-        const is_genesis = opts.anchorState.slot == 0;
+        const anchorCP = types.Checkpoint{ .slot = opts.anchorState.slot, .root = anchor_block_root };
         const fc_store = ForkChoiceStore{
             .time = opts.anchorState.slot * constants.INTERVALS_PER_SLOT,
             .timeSlots = opts.anchorState.slot,
-            .latest_justified = opts.anchorState.latest_justified,
-            .latest_finalized = opts.anchorState.latest_finalized,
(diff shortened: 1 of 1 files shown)


pkgs/node/src/forkchoice.zig
```diff
@@ -326,13 +326,26 @@ pub const ForkChoice = struct {
...
         const proto_array = try ProtoArray.init(allocator, anchor_block);
-        const anchorCP = types.Checkpoint{ .slot = opts.anchorState.slot, .root = anchor_block_root };
+        const is_genesis = opts.anchorState.slot == 0;
         const fc_store = ForkChoiceStore{
             .time = opts.anchorState.slot * constants.INTERVALS_PER_SLOT,
             .timeSlots = opts.anchorState.slot,
-            .latest_justified = anchorCP,
-            .latest_finalized = anchorCP,
+            .latest_justified = opts.anchorState.latest_justified,
+            .latest_finalized = opts.anchorState.latest_finalized,
```
g11tech: this is wrong, anchorCP is the also the justified and finalized for the chain start

PR #585, opened 2026-02-20
fix: set finalized slot from checkpoint sync URL data

ch4r10t33r:
## Summary

- When using `--checkpoint-sync-url`, initializes the fork choice store's `latest_finalized` with the finalized slot from the downloaded state (`anchorState.latest_finalized.slot`) rather than the anchor slot
- `latest_justified` remains set to `anchorCP` (anchor slot/root) as before
- Plumbs `is_checkpoint_sync` flag through `NodeOpts` -> `ChainOpts` -> `ForkChoiceParams` so forkchoice init can distinguish checkpoint sync from genesis/database initialization
- Adds logging to distinguish genesis, checkpoint sync, and database initialization sources

## Test plan

- [x] All unit tests pass locally (`zig build test`)
- [x] Verify checkpoint sync node logs correct finalized slot from URL
- [x] Verify genesis/database init paths are unaffected (default `is_checkpoint_sync = false`)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -812,6 +812,7 @@ fn build_zkvm_targets(
                 .root_source_file = b.path("pkgs/state-transition-runtime/src/main.zig"),
                 .target = target,
                 .optimize = optimize,
+                .strip = true, // Strip debug info to avoid RISC-V relocation overflow
             }),
         });
         // addimport to root module is even required afer declaring it in mod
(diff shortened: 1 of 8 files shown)


ch4r10t33r:
Closing this PR as the functionality in main is already what is expected.

Issue #590, opened 2026-02-22
FutureSlot error in block processing in beam sim


noopur23:
error.FutureSlot in block processing in beam sim
<img width="1280" height="318" alt="Image" src="https://github.com/user-attachments/assets/5d4fefd3-3fdc-450b-8399-bf92b7c84ca3" />

PR #597, opened 2026-02-24
fix(node): retry cached descendants on FutureSlot

GrapeBaBa:
- Fixes the remaining `FutureSlot` race in cached descendant replay during sync.

  ### What changed
  - Treat `ForkChoiceError.FutureSlot` as retryable in cached-descendant processing (keep block cached, don’t evict).
  - Add interval-time retry for cached descendants whose parent is already in forkchoice.
  - Add regression test to ensure:
    1) block stays cached when replayed too early,
    2) block is processed once slot time catches up.


GrapeBaBa:
> explain the scenario where this is needed?

looked at the code again, use the approach which in the `produceBlock` to fix #590 , see #601

PR #620, opened 2026-02-28
node: align forkchoice tie-break with leanspec hash ordering

GrapeBaBa:
## Summary
- Align `ProtoArray.applyDeltasUnlocked` tie-break with leanSpec behavior: equal `weight` now resolves by lexicographically larger `blockRoot` only.
- Remove slot-based preference from tie-break path in `pkgs/node/src/forkchoice.zig`.
- Add regression test covering equal-weight siblings with different slots to ensure hash-order winner.

## Spec/Reference
- leanSpec reference: [code.](https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/forkchoice/store.py#L763)

### DIFF
--- a/pkgs/node/src/node.zig
+++ b/pkgs/node/src/node.zig
@@ -1880,6 +1880,10 @@ test "Node: publishBlock persists locally produced blocks for blocks-by-root syn
     defer node.deinit();
 
     const slot: usize = 4;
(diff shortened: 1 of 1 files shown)


g11tech:
why will we not pick the latest slot to tie break?

PR #631, opened 2026-03-04
feat: only subscribe attestation subnets at libp2p level for aggregators

zclawz:
## Summary

Non-aggregator nodes should publish attestations to subnet topics but **not receive** them. This is enforced at the libp2p level — if you don't subscribe, the Rust gossipsub node won't deliver incoming messages, but you can still publish to the topic.

### `pkgs/network/src/ethlibp2p.zig`
- Added `is_aggregator: bool = false` to `EthLibp2pParams`
- In `run()`: attestation subnet topics are only added to the Rust subscription list when `is_aggregator == true`

### `pkgs/cli/src/node.zig`
- Pass `is_aggregator` from node options to `EthLibp2p.init`

### `pkgs/cli/src/main.zig`
- Pass `is_aggregator` to all three `EthLibp2p.init` calls in the beam multi-node setup

## Behavior

| Node type | libp2p subscribe to attestation subnet | Can publish attestations |
|---|---|---|
| Non-aggregator with validators | ❌ (no receive) | ✅ |
| Aggregator with validators | ✅ | ✅ |
| Passive node (no validators) | N/A | ❌ |

No chain-level filtering needed — the fix is entirely at the network layer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,20 +32,18 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-zig-
 
-    - name: Fetch Zig dependencies with retry
+    - name: Fetch Zig dependencies
       run: |
-        max_attempts=5
-        attempt=1
-        while [ $attempt -le $max_attempts ]; do
+        max_attempts=3
+        for attempt in $(seq 1 $max_attempts); do
           if zig build --fetch; then
             echo "Successfully fetched dependencies on attempt $attempt"
             exit 0
           fi
           echo "Attempt $attempt/$max_attempts failed, retrying in 5 seconds..."
           sleep 5
-          attempt=$((attempt + 1))
         done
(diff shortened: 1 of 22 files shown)


GrapeBaBa (dismissed):

g11tech (commented):
need to figure out if flood publishing is good or should we subscribe and maintain mesh with ignore validator

zclawz:
Implemented the Subscribe+ignore approach as suggested. Changes pushed to this branch:

- Removed the `is_aggregator` guard on attestation subnet subscription — all nodes now subscribe to all attestation subnets unconditionally
- Non-aggregator nodes join the libp2p mesh for routing health but skip attestation message processing via an early return in `handleGossipFromRustBridge`
- Updated `is_aggregator` field comment to reflect the new semantics

This is more robust than flood-publish since the node is a real mesh participant and does not depend on peers being subscribed for message propagation to succeed.

Issue #559, opened 2026-02-10
Implement Devnet 3 Metrics for Committee Aggregation


bomanaps:
https://github.com/leanEthereum/leanMetrics/issues/22

PR #632, opened 2026-03-05
devnet 3 metrics

bomanaps:
Devnet 3 metrics https://github.com/blockblaz/zeam/issues/559 
 https://github.com/leanEthereum/leanMetrics/pull/23

### DIFF
--- a/pkgs/node/src/forkchoice.zig
+++ b/pkgs/node/src/forkchoice.zig
@@ -1282,42 +1282,6 @@ pub const ForkChoice = struct {
         var new_payloads_count: usize = 0;
         var gossip_sigs_count: usize = 0;
 
-        self.signatures_mutex.lock();
-        errdefer self.signatures_mutex.unlock();
-
-        var sig_it = self.gossip_signatures.iterator();
-        while (sig_it.next()) |entry| {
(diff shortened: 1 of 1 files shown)


bomanaps:
A node can have multiple validators, each assigned to a different subnet via validator_id % committee_count. Should lean_attestation_committee_subnet report just the first validator's subnet, or should we create multiple gauge entries with validator ID labels? cc @GrapeBaBa @g11tech

pkgs/cli/src/node.zig
```diff
@@ -268,6 +268,12 @@ pub const Node = struct {
             // Clean up metrics server if subsequent init operations fail
             errdefer if (self.metrics_server_handle) |handle| handle.stop();
 
+            // Set validator status gauges on node start
+            zeam_metrics.metrics.lean_is_aggregator.set(if (options.is_aggregator) 1 else 0);
+            // Committee subnet and count are placeholders until subnet logic is implemented
+            zeam_metrics.metrics.lean_attestation_committee_subnet.set(0);
+            zeam_metrics.metrics.lean_attestation_committee_count.set(0);
```
anshalshukla: We do have subnet logic and count as well, `compute_subnet_id` and config which mentions committee count

PR #643, opened 2026-03-08
fix: only assign one aggregator in simtest

### LINKED ISSUE #641: Make 3 nodes simulate test to just use one aggregator
Current when we run `zig build simtest`, it will set all three nodes as aggregator. Currently we should only have one aggregator for one subnet.

zclawz:
Closes #641

## Summary

When running `zig build simtest`, all three simulated nodes were initialized with `is_aggregator = beamcmd.@"is-aggregator"`. Since a subnet should have exactly one aggregator, nodes 1 and 2 (`beam_node_2` and `beam_node_3`) now always have `is_aggregator = false`; only node 0 (`beam_node_1`) inherits the `--is-aggregator` flag from the CLI argument.

## Testing
- [ ] `zig build test` passes

### DIFF
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -617,7 +617,8 @@ fn mainInner() !void {
                 .db = db_2,
                 .logger_config = &logger2_config,
                 .node_registry = registry_2,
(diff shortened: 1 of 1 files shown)


GrapeBaBa:
duplicate #642

PR #662, opened 2026-03-12
fix: consolidate key loading logic into key-manager

### LINKED ISSUE #661: Multiple load keys from file logic
_Originally addressed_ by @g11tech  https://github.com/blockblaz/zeam/pull/587#discussion_r2926043351

zclawz:
Closes #661

## Summary

There were three separate copies of the same "open pk/sk SSZ files → read bytes → KeyPair.fromSsz" pattern:

1. `pkgs/key-manager/src/lib.zig` – `loadPreGeneratedKey()`
2. `pkgs/cli/src/node.zig` – `loadHashSigKeys()` loop
3. `pkgs/cli/src/main.zig` – `testsig` command

This PR extracts a single shared helper `loadKeypairFromFiles(allocator, sk_path, pk_path)` in the key-manager package and updates all callers to use it.

Changes:
- `key-manager/src/lib.zig`: add public `loadKeypairFromFiles`, make `MAX_SK_SIZE`/`MAX_PK_SIZE` public, refactor `loadPreGeneratedKey` to delegate
- `cli/src/node.zig`: use `key_manager_lib.loadKeypairFromFiles`; remove now-unused `xmss` and `constants` imports
- `cli/src/main.zig`: use `key_manager_lib.loadKeypairFromFiles` in `testsig`; hoist `key_manager_lib` import to file scope

## Testing
- [x] `zig build` passes
- [x] `zig build test` passes (pre-existing IoUring/env failures unaffected)

### DIFF
--- a/pkgs/cli/src/node.zig
+++ b/pkgs/cli/src/node.zig
@@ -451,7 +451,8 @@ pub const Node = struct {
             defer self.allocator.free(pk_path);
 
             var keypair = key_manager_lib.loadKeypairFromFiles(self.allocator, sk_path, pk_path) catch |err| switch (err) {
-                error.FileNotFound => return error.HashSigSecretKeyMissing,
+                error.SecretKeyFileNotFound => return error.HashSigSecretKeyMissing,
+                error.PublicKeyFileNotFound => return error.HashSigPublicKeyMissing,
                 else => return err,
             };
             errdefer keypair.deinit();
--- a/pkgs/key-manager/src/lib.zig
+++ b/pkgs/key-manager/src/lib.zig
@@ -189,12 +189,18 @@ pub fn loadKeypairFromFiles(
     sk_path: []const u8,
     pk_path: []const u8,
 ) !xmss.KeyPair {
-    var sk_file = try std.fs.cwd().openFile(sk_path, .{});
+    var sk_file = std.fs.cwd().openFile(sk_path, .{}) catch |err| switch (err) {
+        error.FileNotFound => return error.SecretKeyFileNotFound,
+        else => return err,
+    };
     defer sk_file.close();
     const sk_data = try sk_file.readToEndAlloc(allocator, MAX_SK_SIZE);
     defer allocator.free(sk_data);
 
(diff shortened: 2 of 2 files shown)


pkgs/cli/src/node.zig
```diff
@@ -452,29 +450,10 @@ pub const Node = struct {
             const pk_path = try std.fmt.allocPrint(self.allocator, "{s}/{s}_pk.ssz", .{ hash_sig_key_dir, base });
             defer self.allocator.free(pk_path);
 
-            // Read secret key
-            var sk_file = std.fs.cwd().openFile(sk_path, .{}) catch |err| switch (err) {
+            var keypair = key_manager_lib.loadKeypairFromFiles(self.allocator, sk_path, pk_path) catch |err| switch (err) {
                 error.FileNotFound => return error.HashSigSecretKeyMissing,
```
anshalshukla: @zclawz we are mapping FileNotFound directly to HashSigSecretKeyMissing but earlier we had both HashSigSecretKeyMissing or HashSigPublicKeyMissing for secret key and public key. We should preserve that behavior as it is helpful for debugging

zclawz:
Fixed in 4d2601f — `loadKeypairFromFiles` now catches `FileNotFound` separately for sk and pk files, returning `error.SecretKeyFileNotFound` and `error.PublicKeyFileNotFound` respectively. `node.zig` then maps those back to `HashSigSecretKeyMissing` and `HashSigPublicKeyMissing` as the original behavior required.

anshalshukla (approved):

pkgs/cli/src/node.zig
zclawz: Fixed — `loadKeypairFromFiles` in key-manager maps `error.FileNotFound` for the secret key path to `error.SecretKeyFileNotFound` and for the public key path to `error.PublicKeyFileNotFound`. `node.zig` then maps those to `HashSigSecretKeyMissing` and `HashSigPublicKeyMissing` respectively, preserving the original distinction for debugging.

PR #685, opened 2026-03-20
fix: subnet subscription for attestations

### LINKED ISSUE #683: subnet subscription for attestations
currently we subscribe to all the subnets and process gossip attestations which we receive over them. But what we actually need to do is subscribe to the subnets for which we have registered validators and recieve & validate those gossip attestations. However we should only import them if we are an aggregator which is determined from the is-aggregator flag in the node command.

also it should be possible to also provide as a command line input  into the node command  in addition to is-aggregator flag as to which subnets to aggregate, which should also then be subscribe to recieve , process and import these gossip attestations.

make sure to this subscription happens at the rust libp2p level itself and we only check for is-aggregator while importing the attestation info forkchoice. the proposer attestation from the block should continue to be imported in the forkchoice as current implementation

zclawz:
Closes #683

## Summary

- Added **`--subnet-ids`** CLI flag (comma-separated, e.g. `0,1,2`) to explicitly configure which attestation subnets to subscribe and aggregate
- When `--subnet-ids` is provided, subscribe only to those subnets at the libp2p gossip level, overriding automatic computation from validator IDs
- **Gated forkchoice import** behind `--is-aggregator` flag: gossip attestations are only imported (into forkchoice via `onGossipAttestation`) when the node is an aggregator
- When `--subnet-ids` is also set alongside `--is-aggregator`, only attestations on those specific subnets are imported — all others are dropped at the chain layer
- Aggregator without `--subnet-ids` imports from all subscribed subnets (existing behaviour preserved for aggregator nodes)
- **Proposer attestations from blocks are unaffected** — they flow through the block processing path and are always imported regardless of aggregator flag

## Testing
- [ ] `zig build test` passes

### DIFF
--- a/pkgs/cli/src/node.zig
+++ b/pkgs/cli/src/node.zig
@@ -691,54 +691,36 @@ fn downloadCheckpointState(
 ) !types.BeamState {
     logger.info("downloading checkpoint state from: {s}", .{url});
 
-    // Parse URL using std.Uri
-    const uri = std.Uri.parse(url) catch return error.InvalidUrl;
-
-    // Initialize HTTP client
     var client = std.http.Client{ .allocator = allocator };
     defer client.deinit();
 
-    // Create HTTP request
-    var req = client.request(.GET, uri, .{}) catch |err| {
-        logger.err("failed to create HTTP request: {any}", .{err});
-        return error.ConnectionFailed;
-    };
-    defer req.deinit();
+    // Use an Allocating writer so client.fetch handles both Content-Length and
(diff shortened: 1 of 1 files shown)


g11tech:
@zclawz I have made changes as per the behavior that is required, can you fix the build

GrapeBaBa:
I checked the gossipsub spec, there is a built-in fanout mechanism which exactly match the only publish case. So I think for non aggregator it doesn't need subscribe the subnet at all. @g11tech @anshalshukla

PR #686, opened 2026-03-20
feat: implement recursive aggregation for devnet4

GrapeBaBa:


pkgs/spectest/src/runner/state_transition_runner.zig
```diff
@@ -414,8 +414,20 @@ fn parseValidators(
...
-                const pubkey = try expect.expectBytesField(FixtureError, types.Bytes52, validator_obj, &.{"pubkey"}, ctx, pubkey_label);
+
+                // Support both old format (pubkey) and new devnet4 format (attestationPubkey + proposalPubkey)
+                const attestation_pubkey = if (validator_obj.get("attestationPubkey")) |_| blk: {
+                    const att_label = std.fmt.bufPrint(&label_buf, "{s}.attestationPubkey", .{base_label}) catch "validator.attestationPubkey";
+                    break :blk try expect.expectBytesField(FixtureError, types.Bytes52, validator_obj, &.{"attestationPubkey"}, ctx, att_label);
+                } else blk: {
+                    const pubkey_label = std.fmt.bufPrint(&label_buf, "{s}.pubkey", .{base_label}) catch "validator.pubkey";
+                    break :blk try expect.expectBytesField(FixtureError, types.Bytes52, validator_obj, &.{"pubkey"}, ctx, pubkey_label);
+                };
```
anshalshukla: nit: imo we should just stay with the updated format and remove the old format support completely

pkgs/node/src/testing.zig
```diff
@@ -220,14 +227,18 @@ pub const NodeTestContext = struct {
 
                 const epoch: u32 = @intCast(aggregated_attestation.data.slot);
 
-                // Perform the actual aggregation
-                try xmss.aggregateSignatures(pub_keys, sig_ptrs, &message_hash, epoch, &signature_proof.proof_data);
+                // Perform the actual aggregation (no children in testing mode)
+                const empty_children_pks: [][]*const xmss.HashSigPublicKey = &.{};
+                const empty_children_proofs: []*const xmss.ByteListMiB = &.{};
+                try xmss.aggregateSignatures(pub_keys, sig_ptrs, empty_children_pks, empty_children_proofs, &message_hash, epoch, 2, &signature_proof.proof_data);
```
anshalshukla: maybe we can just pass 1 here, should slightly improve CI time

anshalshukla (changes requested):

pkgs/types/src/block.zig
```diff
@@ -458,6 +438,417 @@ pub fn createBlockSignatures(allocator: Allocator, num_aggregated_attestations:
...
+            .attestations = attestations_list,
+            .attestation_signatures = signatures_list,
+            .allocator = allocator,
+        };
+    }
+
+    /// Compute aggregated signatures using recursive aggregation:
+    /// Step 1: Collect individual gossip signatures from signatures_map
+    /// Step 2: Greedy child proof selection — new_payloads first, then known_payloads as helpers
+    /// Step 3: Recursive aggregate — combine selected children + remaining gossip sigs
```
anshalshukla: I think it is better to first greedily select from new_payloads and then select signatures and know_payloads, this will avoid aggregation of already aggregated signatures and this is how I've implemented it on the spec side as well

anshalshukla (approved):

pkgs/node/src/forkchoice.zig
```diff
@@ -997,6 +1002,9 @@ pub const ForkChoice = struct {
             const found_entries = sorted_entries.items.len > 0;
 
             for (sorted_entries.items) |map_entry| {
+                // Limit the number of distinct AttestationData entries per block (leanSpec #536).
+                if (processed_att_data.count() >= constants.MAX_ATTESTATIONS_DATA) break;
```
g11tech: add a todo: optimize this to have best entries in terms of moving the justification and finalization forward

g11tech (approved):

pkgs/node/src/forkchoice.zig
anshalshukla: I think it is the case right now, the entries are sorted by target, and entries whose source doesn't match will get rejected so it will maximally include attestation_data that that are at current justified  (finalisation candidate) and with closest (smallest, justification candidate) target slot

pkgs/node/src/constants.zig
```diff
@@ -5,6 +5,9 @@ const params = @import("@zeam/params");
 pub const INTERVALS_PER_SLOT = 5;
 pub const SECONDS_PER_INTERVAL_MS: isize = @divFloor(params.SECONDS_PER_SLOT * std.time.ms_per_s, INTERVALS_PER_SLOT);
 
+// Maximum number of distinct attestation data entries per block.
+pub const MAX_ATTESTATIONS_DATA = 8;
```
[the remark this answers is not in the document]
anshalshukla: @zclawz create a PR on top of main to address this comment

zclawz:
Created PR #711 to address the review comment: https://github.com/blockblaz/zeam/pull/711

Fixes both issues:
- Corrected value from 8 → 16 (per leanSpec #536)
- Moved from `constants.zig` to `ChainSpec` config, readable from `config.yaml` as `MAX_ATTESTATIONS_DATA`

PR #695, opened 2026-03-25
fix: batch pending parent root fetches to avoid 300+ sequential round-trips

ch4r10t33r:
## Summary

Batches pending parent-root fetches into a single `blocks_by_root` request
instead of firing one request per missing parent. Avoids 300+ sequential
round-trips when a peer walks a long parent chain.

Also fixes a missed `flushPendingParentFetches` in an early-return path of
`processBlockByRootChunk` so deferred roots aren't stranded on the queue.

## Scope note

This PR originally also included a `state_mutex` to serialise the Rust
networking thread against the libxev main thread. That has since been
addressed on `main` by #765 (`BeamNode.mutex`, narrower scope) after
#765 fixed the underlying SSZ-serialisation bug that was the real cause
of the \`integer overflow\` crash originally attributed to a data race.
The mutex commits have been dropped from this branch; only the batching
optimisation remains.

## Test plan

- [x] \`zig fmt --check .\`
- [x] \`zig build\`
- [x] \`zig build test --summary all\`
- [x] \`zig build simtest --summary all\`

### DIFF
--- a/pkgs/node/src/node.zig
+++ b/pkgs/node/src/node.zig
@@ -63,7 +63,7 @@ pub const BeamNode = struct {
     /// Maps block root → fetch depth. Collected during gossip/RPC processing
     /// and flushed as a single batched blocks_by_root request, avoiding the
     /// 300+ individual round-trips caused by sequential parent-chain walking.
-    pending_parent_roots: std.AutoHashMap(types.Root, u32),
+    batch_pending_parent_roots: std.AutoHashMap(types.Root, u32),
 
     const Self = @This();
 
--- a/pkgs/node/src/node.zig
+++ b/pkgs/node/src/node.zig
@@ -59,6 +59,11 @@ pub const BeamNode = struct {
     /// Serializes BeamNode work between the libxev main thread (onInterval) and
     /// the libp2p worker thread (onGossip / onReqRespResponse / onReqRespRequest).
     mutex: std.Thread.Mutex = .{},
+    /// Pending parent roots deferred for batched fetching.
+    /// Maps block root → fetch depth. Collected during gossip/RPC processing
+    /// and flushed as a single batched blocks_by_root request, avoiding the
+    /// 300+ individual round-trips caused by sequential parent-chain walking.
+    pending_parent_roots: std.AutoHashMap(types.Root, u32),
 
     const Self = @This();
 
@@ -116,6 +121,7 @@ pub const BeamNode = struct {
             .logger = opts.logger_config.logger(.node),
             .node_registry = opts.node_registry,
             .aggregation_subnet_ids = opts.aggregation_subnet_ids,
+            .pending_parent_roots = std.AutoHashMap(types.Root, u32).init(allocator),
         };
 
         chain.setPruneCachedBlocksCallback(self, pruneCachedBlocksCallback);
@@ -124,6 +130,7 @@ pub const BeamNode = struct {
     }
 
     pub fn deinit(self: *Self) void {
+        self.pending_parent_roots.deinit();
         self.network.deinit();
         self.chain.deinit();
         self.allocator.destroy(self.chain);
@@ -189,6 +196,8 @@ pub const BeamNode = struct {
                             });
                         }
                     }
+                    // Flush any pending parent root fetches accumulated during caching.
+                    self.flushPendingParentFetches();
                     // Return early - don't pass to chain until parent arrives
                     return;
                 }
@@ -346,6 +355,9 @@ pub const BeamNode = struct {
                 );
             };
         }
+
+        // Flush any parent roots accumulated during block/descendant processing.
+        self.flushPendingParentFetches();
     }
 
     fn pruneCachedBlocksCallback(ptr: *anyopaque, finalized: types.Checkpoint) usize {
@@ -510,7 +522,6 @@ pub const BeamNode = struct {
         AllocationFailed,
         CloneFailed,
         CachingFailed,
-        FetchFailed,
     };
 
     /// Cache a block and fetch its parent. Common logic used by both gossip and req-resp handlers.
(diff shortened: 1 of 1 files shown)


zclawz (commented):
Good optimization — clear problem, clean solution. A few observations:

**1. `roots.deinit(self.allocator)` — likely a bug**

`std.ArrayList` stores its allocator internally, so `deinit()` takes no arguments. This should be `roots.deinit()` (no parameter). With the extra argument this won't compile.

If you intended an unmanaged list, use `std.ArrayListUnmanaged` + `deinit(allocator)`. But since `self.allocator` is already captured at `initCapacity`, plain `ArrayList` + `deinit()` is the right choice here.

**2. `max_depth` across all batched roots**

Using the maximum depth across all pending roots for the single batched request is conservative but correct — the worst case is we request a few more ancestors than strictly needed for shallower roots. Fine as a starting point; could be refined later if needed.

**3. `FetchFailed` → `CachingFailed` consolidation**

Replacing the now-removed `FetchFailed` with `CachingFailed` (for the `put` allocation failure path) is reasonable — an allocation failure during enqueue is effectively a caching failure. Callers that handled `FetchFailed` may need updating if they had separate recovery logic, but since this is an internal error type it's unlikely to matter.

**4. Flush coverage looks complete**

The four flush sites (gossip early-return, `handleGossipProcessingResult`, `processBlockByRootChunk` end, and the RPC response handler) cover the natural exit points. Roots that accumulate on error paths will be flushed on the next message cycle, which is acceptable.

Fix the `deinit` call and this looks good to merge.

pkgs/node/src/node.zig
```diff
@@ -945,6 +964,38 @@ pub const BeamNode = struct {
...
+        if (count == 0) return;
+
+        var roots = std.ArrayList(types.Root).initCapacity(self.allocator, count) catch {
+            self.logger.warn("failed to allocate roots list for pending parent fetch flush", .{});
+            return;
+        };
+        defer roots.deinit(self.allocator);
+
+        var max_depth: u32 = 0;
+        var it = self.pending_parent_roots.iterator();
```
zclawz: **Bug:** `std.ArrayList.deinit` takes no arguments — the allocator was already captured at `initCapacity` time. This should be `defer roots.deinit();`. Using `ArrayListUnmanaged` is the pattern that requires an explicit allocator in `deinit`.

zclawz (changes requested):
## Review

Good direction — batching sequential parent-root fetches into one `blocks_by_root` request is the right fix for the 300+ round-trip problem. One compile error before merging.

### 🔴 Bug — `ArrayList.deinit` wrong signature

```zig
var roots = std.ArrayList(types.Root).initCapacity(self.allocator, count) catch { ... };
defer roots.deinit(self.allocator);   // ← compile error
```

`std.ArrayList.deinit` takes **no arguments** (the allocator is stored in the ArrayList at `initCapacity` time). `ArrayListUnmanaged` is the variant that requires an explicit allocator in `deinit`. Fix:

```zig
defer roots.deinit();
```

### ✅ Design notes (no action needed)

**`max_depth` for the batched request** — when multiple roots land in one flush, the highest depth is used for all of them. Conservative but correct: worst case it fetches more ancestors than strictly needed for some roots, but it never under-fetches. Fine.

**Same root enqueued twice** — `put` silently overwrites the depth. Because `flushPendingParentFetches` takes `max_depth` across all map entries anyway, the overwrite doesn't cause under-fetching. Fine.

**`FetchFailed` removal** — removing it from the enum is correct; failures are now non-fatal (logged as warn) and the caller no longer propagates a hard error on missing parent. ✅

**Flush sites** — gossip early-return (line 196), gossip end-of-handler (line 355), req-resp early-return (line 715), req-resp end-of-handler (line 761). Looks comprehensive. ✅

**Locking** — `pending_parent_roots` is accessed inside the existing `BeamNode.mutex` boundary (gossip and RPC handlers both hold it), so no additional synchronisation is needed. ✅

Fix the `deinit` call and this is good to merge.

pkgs/node/src/node.zig
```diff
@@ -59,6 +59,11 @@ pub const BeamNode = struct {
     /// Serializes BeamNode work between the libxev main thread (onInterval) and
     /// the libp2p worker thread (onGossip / onReqRespResponse / onReqRespRequest).
     mutex: std.Thread.Mutex = .{},
+    /// Pending parent roots deferred for batched fetching.
+    /// Maps block root → fetch depth. Collected during gossip/RPC processing
+    /// and flushed as a single batched blocks_by_root request, avoiding the
+    /// 300+ individual round-trips caused by sequential parent-chain walking.
+    pending_parent_roots: std.AutoHashMap(types.Root, u32),
```
g11tech:
```suggestion
    batch_pending_parent_roots: std.AutoHashMap(types.Root, u32),
```

g11tech (approved):

zclawz (changes requested):
Still not approved — the original bug is still present.

```zig
defer roots.deinit(self.allocator);   // ← compile error
```

`std.ArrayList.deinit` takes **no arguments** — the allocator is captured internally at `initCapacity` time. Passing `self.allocator` here is `ArrayListUnmanaged` API applied to `ArrayList`. Fix:

```zig
defer roots.deinit();
```

The field rename (`pending_parent_roots` → `batch_pending_parent_roots`) is fine. Everything else is unchanged and correct. One line to fix.

zclawz (approved):
Approved — and apologies for the false flag on `deinit`. In Zig 0.15.x `ArrayList.deinit` takes an explicit allocator argument (the API was updated to explicit allocator passing); my review was based on pre-0.14 semantics where the allocator was stored internally. CI passing confirmed it compiles fine.

All the substantive logic (batch map, `max_depth`, flush sites, `FetchFailed` removal) is correct. 🚢

Issue #564, opened 2026-02-11
Zeam node crashing in long running devnet


anshalshukla:
```
Feb-10 03:36:47.931 [debug] (zeam): [network] network-0:: onReqRespRequest=ReqRespRequest{ status } handlers=1 from peer=16Uiu2HAmPQhkD6Zg5Co2ee8ShshkiY4tDePKFARPpCS2oKSLj1E1(ream_0)
Feb-10 03:36:47.931 [debug] (zeam): [network] network-0:: serverStreamSendResponse ctx.method=status response.tag=status peer=16Uiu2HAmPQhkD6Zg5Co2ee8ShshkiY4tDePKFARPpCS2oKSLj1E1(ream_0)
Feb-10 03:36:47.931 [debug] (zeam): [network] network-0:: Streaming status response to peer=16Uiu2HAmPQhkD6Zg5Co2ee8ShshkiY4tDePKFARPpCS2oKSLj1E1(ream_0) channel=492
Feb-10 03:36:47.931 [info] (zeam): [network] rust-bridge: [reqresp] Sent response payload on channel 492 (peer: 16Uiu2HAmPQhkD6Zg5Co2ee8ShshkiY4tDePKFARPpCS2oKSLj1E1)
Feb-10 03:36:47.931 [info] (zeam): [network] rust-bridge: [reqresp] Sent end-of-stream on channel 492 (peer: 16Uiu2HAmPQhkD6Zg5Co2ee8ShshkiY4tDePKFARPpCS2oKSLj1E1)
Feb-10 03:36:47.972 [info] (zeam): [network] rust-bridge: [reqresp] Received request from 16Uiu2HAm7TYVs6qvDKnrovd9m4vvRikc4HPXm1WyLumKSe5fHxBv for protocol /leanconsensus/req/blocks_by_root/1/ssz_snappy (53 bytes)
Feb-10 03:36:47.972 [error] (zeam): [network] Error in deserializing the blocks_by_root RPC request from peer=16Uiu2HAm7TYVs6qvDKnrovd9m4vvRikc4HPXm1WyLumKSe5fHxBv(lantern_0): error.OutOfMemory
Feb-10 03:36:47.973 [warning] (zeam): [network] SSZ deserialization failed for blocks_by_root message - written 32 bytes to debug file: deserialization_dumps/failed_blocks_by_root_1770694607.bin
Feb-10 03:36:54.046 [info] (zeam): [database] initializing RocksDB
[...]

PR #705, opened 2026-04-08
add bound checks before decoding

anshalshukla:
Updates ssz lib which performs max, min length checks before decoding to avoid crashes due to arbitrarily big memory allocation


anshalshukla:
closes #564

PR #716, opened 2026-04-10
route attester/proposer keys by filename

bomanaps:


pkgs/cli/src/node.zig
```diff
@@ -512,31 +512,77 @@ pub const Node = struct {
 
         const hash_sig_key_dir = self.options.hash_sig_key_dir;
 
+        // First pass: group assignments by validator index, routing by filename.
+        // If the filename contains "attester" it goes to att_base; "proposer" to prop_base.
+        // A filename with neither is treated as a single key used for both roles.
```
anshalshukla: I think we should return an error when a single key is present

anshalshukla: @zclawz add a commit to resolve this comment

zclawz: Done — instead of silently treating ambiguous filenames as a single key for both roles, we now return `error.InvalidPrivkeyFileFormat`. Pushed in 611b5b9.

anshalshukla (approved):

GrapeBaBa (approved):

PR #718, opened 2026-04-12
fix: all nodes process attestations

shariqnaiyer:
This is a fix I had to make to make the zeam nodes process attestations as they were not aggregators.

I believe as per service.py in the spec all nodes call on_gossip_attestation.

There still may be some minor difference from spec on the is_aggregator logic.

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -744,25 +744,16 @@ pub const BeamChain = struct {
                     }
                 };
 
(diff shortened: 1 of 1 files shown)


g11tech (changes requested):
sorry a node who is not aggregator should not process/import attestations and even if they are aggregators, they are not supposed to import attestations from the subnets they are not listening to, ideally only use them for aggregation

PR #724, opened 2026-04-13
fix: catch xmss_setup_prover panic and return error code instead of aborting (#722)

### LINKED ISSUE #722: panic: xmss_setup_prover fails with NotFound in aggregator on multi-subnet devnet
## Bug

When running zeam as an aggregator in a multi-subnet devnet, the node panics during `aggregate` on the clock interval:

```
thread '<unnamed>' (1) panicked at .../leanmultisig/crates/rec_aggregation/src/compilation.rs:52:50:
called Result::unwrap() on an Err value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

## Stack trace (key frames)

```
19: xmss_setup_prover
20: aggregation.setupProver       pkgs/xmss/src/aggregation.zig:60
21: aggregation.aggregateSignatures   pkgs/xmss/src/aggregation.zig:90
22: aggregation.AggregatedSignatureProof.aggregate   pkgs/types/src/aggregation.zig:125
23: block.AggregatedAttestationsResult.computeAggregatedSignatures   pkgs/types/src/block.zig:701
24: forkchoice.ForkChoice.aggregateUnlocked   pkgs/node/src/forkchoice.zig:1461
25: forkchoice.ForkChoice.aggregate   pkgs/node/src/forkchoice.zig:1535
26: chain.BeamChain.aggregate   pkgs/node/src/chain.zig:1530
27: chain.BeamChain.maybeAggregateOnInterval   pkgs/node/src/chain.zig:1559
28: node.BeamNode.onInterval   pkgs/node/src/node.zig:1164
```

## Context

- Reproducible in multi-subnet devnet with aggregator role
- Prior to the panic, gossip attestations are being rejected with `error.VerificationFailed`
- `rec_aggregation/src/compilation.rs:52` calls `.unwrap()` on a file-open/read result — the file (likely a compiled prover circuit or key file) does not exist at the expected path
- `xmss_setup_prover` is called via FFI from `pkgs/xmss/src/aggregation.zig:60` (`setupProver`)

## Expected behaviour

`setupProver` should return an error instead of panicking, and the aggregator should handle the missing prover file gracefully (log + skip aggregation) rather than bringing down the node.

## Fix needed

1. In `leanmultisig` / `rec_aggregation`: replace `.unwrap()` at `compilation.rs:52` with proper error propagation (return `Result::Err`)
2. In `pkgs/xmss/src/aggregation.zig`: `setupProver` should surface the error to its caller
3. In `pkgs/node/src/forkchoice.zig` (`aggregateUnlocked`): catch the error, log a warning, and skip aggregation for this interval rather than crashing

## Logs

Reported by @ch4r10t33r during devnet4 testing.

zclawz:
Closes #722

## Root cause

`init_aggregation_bytecode()` in `rec_aggregation/compilation.rs:52` calls `.unwrap()` on a file-open. When the prover bytecode file is missing (ENOENT — e.g. on a fresh aggregator node), it panics inside `xmss_setup_prover()`. A Rust panic propagating through an `extern "C"` boundary is UB and triggers `panic_cannot_unwind`, aborting the process.

This fires on every `maybeAggregateOnInterval` tick, so the aggregator crashes repeatedly during devnet4 multi-subnet testing.

### `rust/multisig-glue/src/lib.rs`
- Remove `static Once` guards (replaced by `OnceLock<bool>`)
- Wrap init body in `std::panic::catch_unwind`; cache the `bool` result
- Change `xmss_setup_prover` / `xmss_setup_verifier` return type from `void` → `i32` (`0` = success, `-1` = failure)
- Subsequent calls return the cached result with no work

### `pkgs/xmss/src/aggregation.zig`
- Update `extern fn` declarations to `c_int` return type
- `setupProver()` / `setupVerifier()` now return `!void` and surface `error.ProverSetupFailed` / `error.VerifierSetupFailed`
- All three callsites changed to `try setupProver()` / `try setupVerifier()`
- `aggregateSignatures()` already returns `!void`, so the error propagates through `computeAggregatedSignatures` → `aggregateUnlocked` → `aggregate` → `maybeAggregateOnInterval`, which **already** catches and logs a warning — so no crash, just a skip

## Behaviour after fix

```
[warn] failed to aggregate attestation signatures for slot=64: error.ProverSetupFailed
```

Aggregation is skipped for that interval; the node keeps running.

## Testing
- `zig build` passes ✅

### DIFF
--- a/pkgs/xmss/src/aggregation.zig
+++ b/pkgs/xmss/src/aggregation.zig
@@ -13,8 +13,11 @@ pub const ByteListMiB = ssz.utils.List(u8, MAX_AGGREGATE_SIGNATURE_SIZE);
 pub const AggregatedXMSS = opaque {};
 
 // External C functions from multisig-glue (uses leanMultisig devnet4 with recursive aggregation)
-extern fn xmss_setup_prover() void;
-extern fn xmss_setup_verifier() void;
+/// Returns 0 on success, -1 if the prover bytecode file is missing or initialisation failed.
+/// Never panics — the Rust side wraps the body in catch_unwind (fix for #722).
+extern fn xmss_setup_prover() c_int;
+/// Returns 0 on success, -1 on failure.
+extern fn xmss_setup_verifier() c_int;
 
 extern fn xmss_aggregate(
     // Raw XMSS signatures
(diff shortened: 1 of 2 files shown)


ch4r10t33r (approved):
Looks good.

PR #726, opened 2026-04-13
fix: use-after-free in writeFailedBytes causing segfault in handleMsgFromRustBridge

### LINKED ISSUE #725: segfault in handleMsgFromRustBridge logger after error.TooLarge snappy decode
## Bug

After `error.TooLarge` is returned from snappy decoding a gossip block message, the node segfaults inside the logging path of `handleMsgFromRustBridge`:

```
Apr-13 21:10:13.503 [error] (zeam): [network] Error in snappyz decoding the message for topic=/leanconsensus/devnet0/block/ssz_snappy: error.TooLarge
Apr-13 21:10:13.509 [warning] (zeam): [network] SSZ deserialization failed for snappyz_decode message - written 9374750 bytes to debug file: deserialization_dumps/failed_snappyz_decode_1776114613.bin
Segmentation fault at address 0xe60625740000
/home/runner/work/zeam/zeam/pkgs/node/src/utils.zig:1007:25: log__anon_2304844 (zeam)
/home/runner/work/zeam/zeam/pkgs/utils/src/log.zig:306:19: handleMsgFromRustBridge (zeam)
```

## Stack trace

```
Segmentation fault at address 0xe60625740000
???:?:?: 0xe6062e061b10 in ??? (libc.so.6)
std/Io/Writer.zig:1007:25: 0x211890f in log__anon_2304844 (zeam)
pkgs/utils/src/log.zig:306:19: 0x210e40b in handleMsgFromRustBridge (zeam)
???:?:?: 0x277222f in ??? (???)
...
std/Thread.zig:509:13: 0x1bac4cf in entryFn (zeam)
```

## Analysis

- The segfault is in `log__anon_2304844` called from `log.zig:306` inside `handleMsgFromRustBridge`
- The crash address `0xe60625740000` looks like a dangling/invalid pointer — likely a slice pointing into a buffer that has already been freed or partially consumed after the `error.TooLarge` path
- The debug dump writes 9,374,750 bytes (9 MB) to disk before crashing — suggesting the oversized payload buffer is still alive long enough to be written, but then the slice used for logging is stale
- Likely cause: after `decodeWithMax` returns `error.TooLarge`, `handleMsgFromRustBridge` logs using a `[]u8` slice derived from the original Rust-bridge message buffer. If that buffer has been freed or reallocated on the Rust side before the Zig log call completes (cross-thread race), the log write segfaults

## Suspected location

`pkgs/utils/src/log.zig:306` — the warning log in the `error.TooLarge` / deserialization failure branch of `handleMsgFromRustBridge`

## Expected behaviour

The node should log the error and drop the oversized message cleanly without crashing.

## Environment

- Multi-subnet devnet4, aarch64 Linux runner
- Reported by @ch4r10t33r
- Related: PR #707 (added `decodeWithMax` — may have introduced the dangling-slice path)

## Note on related issues

The `error.VerificationFailed` attestations and `InsufficientPeers` publish errors in the same log session are separate — tracked in the broader devnet4 multi-subnet investigation.

zclawz:
Closes #725

## Root cause

writeFailedBytes allocated a filename with allocPrint, then freed it with defer allocator.free(filename) before returning it to the caller. Three callers then logged that freed pointer — classic use-after-free, segfaults inside the Zig logger.

Observed crash on devnet4: a 9 MB gossip block triggered error.TooLarge, writeFailedBytes wrote the dump and freed the filename, then handleMsgFromRustBridge tried to log the freed pointer — segfault at 0xe60625740000.

## Fix

Change writeFailedBytes return type from ?[]const u8 to bool. Move all logging (including the file path) inside the function while the allocation is still live. Three callsites updated: handleMsgFromRustBridge, deserializeGossipMessage, handleRPCRequestFromRustBridge.

- zig build passes
- zig fmt clean

### DIFF
--- a/pkgs/network/src/ethlibp2p.zig
+++ b/pkgs/network/src/ethlibp2p.zig
@@ -236,16 +236,20 @@ fn serverStreamIsFinished(ptr: *anyopaque) bool {
     return ctx.finished;
 }
 
-/// Writes failed deserialization bytes to disk for debugging purposes
-/// Returns the filename if the file was successfully created, null otherwise
-/// If timestamp is null, generates a new timestamp automatically
-fn writeFailedBytes(message_bytes: []const u8, message_type: []const u8, allocator: Allocator, timestamp: ?i64, logger: zeam_utils.ModuleLogger) ?[]const u8 {
+/// Writes failed deserialization bytes to disk for debugging purposes.
+/// Logs the outcome (success or failure) itself; returns true on success.
+///
+/// Previously returned `?[]const u8` (the allocated filename) while also doing
+/// `defer allocator.free(filename)` — so callers received a dangling pointer and
+/// segfaulted when logging it.  The fix: log from inside this function and return
+/// a plain bool; callers no longer touch the filename string at all (#725).
+fn writeFailedBytes(message_bytes: []const u8, message_type: []const u8, allocator: Allocator, timestamp: ?i64, logger: zeam_utils.ModuleLogger) bool {
     // Create dumps directory if it doesn't exist
     std.fs.cwd().makeDir("deserialization_dumps") catch |e| switch (e) {
         error.PathAlreadyExists => {}, // Directory already exists, continue
         else => {
             logger.err("Failed to create deserialization dumps directory: {any}", .{e});
-            return null;
+            return false;
         },
     };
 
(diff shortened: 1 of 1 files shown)


ch4r10t33r (approved):
Looks good.

PR #729, opened 2026-04-13
docker, build: use x86-64-v3 baseline for portable Docker images

ch4r10t33r:
## Summary

- The Rust libs were compiled with `-Ctarget-cpu=native`, which enables AVX512 when the CI builder has it. Servers without AVX512 then crash with "illegal instruction".
- `build.zig`: add `-Drust-target-cpu` option (defaults to `native` for local builds). The Dockerfile passes `x86-64-v3` (AVX2, no AVX512).
- `Dockerfile`: for amd64, pass `-Dcpu=x86_64_v3` (Zig) and `-Drust-target-cpu=x86-64-v3` (Rust) so the image runs on any x86_64 server with AVX2 support (Haswell / 2013+).

## Root cause

`build_rust_project()` in `build.zig` hardcodes `CARGO_ENCODED_RUSTFLAGS="-Ctarget-cpu=native"` on x86_64. When CI builds on a machine with AVX512 (e.g. Xeon Scalable), the Rust libraries (leanMultisig, hashsig-glue, libp2p-glue) are compiled with AVX512 instructions. Zig code also defaults to native CPU features. Servers without AVX512 hit SIGILL ("illegal instruction") at runtime.

## Fix

**`build.zig`**: Replace the hardcoded `target-cpu=native` with a configurable build option:

```
-Drust-target-cpu=<value>   (default: native)
```

Local builds remain unchanged (native). Docker/CI builds pass `x86-64-v3`.

**`Dockerfile`**: For amd64, pass both Zig and Rust CPU targets:

```
zig build ... -Dcpu=x86_64_v3 -Drust-target-cpu=x86-64-v3
```

`x86-64-v3` enables AVX2/FMA/BMI but not AVX512 — compatible with all x86_64 servers from ~2013 onwards.

## Test plan

- [ ] Docker image builds successfully for amd64
- [ ] Image runs without "illegal instruction" on servers without AVX512
- [ ] Local `zig build` still uses native CPU features (no regression)
- [ ] leanMultisig SIMD codepaths still activate (AVX2 enabled via x86-64-v3)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,10 +80,15 @@ COPY .git/refs .git/refs
 
 # Get git commit hash and build the project with optimizations
 # Use cache mount for Zig build cache as well
-# Set RUSTFLAGS for ARM64 to use generic CPU (QEMU-compatible)
+# CPU targeting per architecture:
+#   amd64: x86-64-v3 (AVX2, no AVX512) — portable across modern x86_64 servers
+#   arm64: generic CPU — QEMU-compatible when cross-building
 ARG TARGETARCH
 RUN --mount=type=cache,target=/root/.cache/zig \
-    if [ "$TARGETARCH" = "arm64" ]; then \
+    EXTRA_ZIG_FLAGS="" && \
+    if [ "$TARGETARCH" = "amd64" ]; then \
+        EXTRA_ZIG_FLAGS="-Dcpu=x86_64_v3 -Drust-target-cpu=x86-64-v3"; \
+    elif [ "$TARGETARCH" = "arm64" ]; then \
         export RUSTFLAGS="-C target-cpu=generic"; \
     fi && \
     GIT_VERSION=$(cat .git/HEAD | grep -o '[0-9a-f]\{40\}' || echo "unknown") && \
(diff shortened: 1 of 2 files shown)


zclawz (approved):
LGTM — clean fix for a real portability problem.

**What it does:** Separates "build for the machine I am on" (`native`) from "build for any modern server" (`x86-64-v3`). Local builds stay fast/optimal; Docker images become portable across any x86_64 host with AVX2 (Haswell 2013+, so effectively all current cloud VMs).

**Changes reviewed:**

`build.zig`
- New `-Drust-target-cpu` option with `"native"` default — no behaviour change for local builds, zero friction ✅
- `b.fmt` allocation is correct (build-step lifetime, no leak issue) ✅
- Comment updated to explain the new `x86-64-v3` use-case clearly ✅

`Dockerfile`
- `EXTRA_ZIG_FLAGS` shell variable correctly handles all three cases: amd64 (AVX2 flags), arm64 (RUSTFLAGS generic, already present), other (no flags) ✅
- Zig flag `-Dcpu=x86_64_v3` and Rust flag `-Drust-target-cpu=x86-64-v3` applied together — consistent ✅
- arm64 `RUSTFLAGS` export preserved, not broken by the restructuring ✅

**One minor note:** `EXTRA_ZIG_FLAGS` is unquoted in the final `zig build` invocation (`$EXTRA_ZIG_FLAGS`). This is intentional — the flags need word-splitting — but worth a comment in case someone adds a flag with spaces later. Not a blocker.

Approved. 🚢

PR #732, opened 2026-04-14
multisig-glue: wrap xmss_aggregate and xmss_verify in catch_unwind

ch4r10t33r:
## Summary

- `rec_xmss_aggregate` and `xmss_verify_aggregation` can panic inside the `rec_aggregation` SIMD backend. A Rust panic through an `extern "C"` boundary is UB — on Linux it manifests as a **General Protection Exception** with no address, crashing the process.
- Wraps both FFI entry points in `catch_unwind` (same pattern as #724 for the setup functions). On panic, `xmss_aggregate` returns null and `xmss_verify_aggregated` returns false, letting the Zig caller handle the failure gracefully.

**Observed in CI:** risc0 prover test crashes with GPE at `aggregation.zig:139` during `genMockChain`.

Note: This is a defensive fix that prevents the crash from being UB. The root cause of the panic inside `rec_xmss_aggregate` still needs investigation — it could be an assertion failure in the SIMD arithmetic backend, missing/invalid prover state, or a target-cpu mismatch causing illegal instructions.

## Test plan

- [ ] CI risc0 prover test should no longer crash with GPE (will get `AggregationFailed` error instead)
- [ ] Existing aggregation tests pass
- [ ] Investigate root cause of panic in `rec_xmss_aggregate`

### DIFF
--- a/rust/multisig-glue/src/lib.rs
+++ b/rust/multisig-glue/src/lib.rs
@@ -182,16 +182,22 @@ pub unsafe extern "C" fn xmss_aggregate(
         .map(|(pks, proof)| (pks.as_slice(), proof))
         .collect();
 
-    // Call rec_aggregation
-    let (_pub_keys, agg_sig) = rec_xmss_aggregate(
-        &children_with_keys,
-        raw_xmss,
-        message_hash,
-        slot,
-        log_inv_rate,
-    );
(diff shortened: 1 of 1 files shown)


ch4r10t33r:
Superseded by #733 which includes both the catch_unwind fix and the CI x86-64-v3 target-cpu fix.

PR #735, opened 2026-04-14
build, auto-release: fix RUSTFLAGS precedence silently dropping target-cpu

ch4r10t33r:
## Summary

Fixes the root cause of "Illegal instruction" errors in released AMD64 Docker images. PR #729 added `-Drust-target-cpu=x86-64-v3` support, but the flag was silently ignored in CI and release builds.

**Root cause**: `build.zig` set `CARGO_ENCODED_RUSTFLAGS` to pass `-Ctarget-cpu` to Cargo. Per [Cargo docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html), `CARGO_ENCODED_RUSTFLAGS` is **completely ignored when `RUSTFLAGS` is set**. The `actions-rust-lang/setup-rust-toolchain` GitHub Action sets `RUSTFLAGS=-D warnings` in every CI/release job, silently dropping the target-cpu flag.

**Changes:**

- **`build.zig`**: Switch from `CARGO_ENCODED_RUSTFLAGS` to `RUSTFLAGS` directly so the target-cpu flag overrides any inherited environment value
- **`auto-release.yml`**: Add `-Dcpu=x86_64_v3 -Drust-target-cpu=x86-64-v3` to the amd64 native build step (was using `-Dcpu=baseline` with no Rust target-cpu)

## What was broken

| Build path | Before this PR | After |
|---|---|---|
| `Dockerfile` direct | Works (no RUSTFLAGS in Docker env) | Works |
| `auto-release.yml` → published images | **Broken** (RUSTFLAGS override + missing flag) | Fixed |
| CI `docker-build` native step | **Broken** (RUSTFLAGS override) | Fixed |

## Test plan

- [ ] CI builds pass on both ubuntu and macos runners
- [ ] Verify with `objdump` that released amd64 binary contains AVX2 but no AVX-512 instructions

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -184,7 +184,7 @@ jobs:
     - name: Build zeam natively
       run: |
         if [ "${{ matrix.arch }}" = "amd64" ]; then
-          zig build -Doptimize=ReleaseSafe -Dtarget=x86_64-linux-gnu -Dcpu=baseline -Dgit_version="$(git rev-parse --short HEAD)"
+          zig build -Doptimize=ReleaseSafe -Dtarget=x86_64-linux-gnu -Dcpu=x86_64_v3 -Drust-target-cpu=x86-64-v3 -Dgit_version="$(git rev-parse --short HEAD)"
         else
           zig build -Doptimize=ReleaseSafe -Dtarget=aarch64-linux-gnu -Dcpu=baseline -Dgit_version="$(git rev-parse --short HEAD)"
         fi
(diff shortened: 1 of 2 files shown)


zclawz (approved):
LGTM — this correctly fixes the silent no-op introduced by #729.

**Root cause is accurately identified:** Per Cargo docs, `CARGO_ENCODED_RUSTFLAGS` is completely ignored when `RUSTFLAGS` is already set. `actions-rust-lang/setup-rust-toolchain` sets `RUSTFLAGS=-Dwarnings` in every job, so the `-Ctarget-cpu` flag from #729 was being silently dropped — meaning release images were still compiled with `native` (potentially AVX512) despite the new option existing.

**`build.zig` change:**
- Switches to `RUSTFLAGS` directly, value `"-Ctarget-cpu={s} -Dwarnings"` — correctly re-includes `-Dwarnings` so CI lint behaviour is preserved ✅
- Trade-off: any additional `RUSTFLAGS` set by other tooling would be clobbered, but given the current CI setup this is the right call and the comment explains the reasoning clearly ✅

**`auto-release.yml` change:**
- amd64 release builds now pass `-Dcpu=x86_64_v3 -Drust-target-cpu=x86-64-v3` — consistent with Dockerfile from #729 ✅
- arm64 path unchanged ✅

Without this, #729 was effectively dead code in CI/release. This makes the portability guarantee real. Approved. 🚢

g11tech (approved):

PR #742, opened 2026-04-16
fix(validator): deduplicate validator indices from assignments

GrapeBaBa:
## Summary
- `annotated_validators.yaml` contains two entries per validator (attester key + proposer key), both with the same `index`
- `getValidatorIndices()` returned duplicates (e.g. `[2, 2]`), causing `mayBeDoAttestation()` to construct and publish the same attestation twice per slot
- The second publish was rejected by gossipsub with `Publish error: Duplicate`
- Fix: deduplicate indices using a hash set

## Reproduction
Run 4-node local devnet (or 2 zeam + 2 nlean interop). Every slot at interval 1:
```
[error] [network] rust-bridge: Publish error: Duplicate
```

## Test plan
- [ ] Run 4-zeam local devnet, verify zero `Publish error: Duplicate` in logs
- [ ] Run 2 zeam + 2 nlean interop, verify zero duplicate errors

### DIFF
--- a/pkgs/cli/src/node.zig
+++ b/pkgs/cli/src/node.zig
@@ -107,11 +107,20 @@ pub const NodeOptions = struct {
     }
 
     pub fn getValidatorIndices(self: *const NodeOptions, allocator: std.mem.Allocator) ![]usize {
-        var indices = try allocator.alloc(usize, self.validator_assignments.len);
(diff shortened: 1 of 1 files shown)


ch4r10t33r (approved):
Looks good to me.

anshalshukla (approved):

g11tech (approved):

PR #745, opened 2026-04-16
ci: add Hive integration workflow (daily + run-hive label)

ch4r10t33r:
## Summary

Implements Hive CI integration per issue #740.

- **Daily scheduled run** at `03:00 UTC` against `main`.
- **Label-triggered PR run**: only executes when the `run-hive` label is applied to a PR, mirroring the existing `stf` label pattern used by `risc0.yml`.
- **Manual trigger** via `workflow_dispatch` with overridable simulator, devnet profile, hive_repository, hive_version inputs.

## Where failures surface

| Trigger | Report | Location |
|---|---|---|
| `schedule` | Auto-opened GitHub issue (labels: `ci-failure`, `hive`), updated in place across consecutive failing days | [issues](https://github.com/blockblaz/zeam/issues?q=is%3Aissue+is%3Aopen+label%3Aci-failure+label%3Ahive) |
| `pull_request` (label) | Comment posted on the PR with run status + artifact reference | The PR itself |
| `workflow_dispatch` | Artifacts on the workflow run | Actions tab |

The scheduled-failure issue is opened/updated by `JasonEtco/create-an-issue@v2` using the template at `.github/ISSUE_TEMPLATE/hive-scheduled-failure.md`. Close the issue once a subsequent daily run goes green.

## Upstream integration

The required upstream pieces already exist in `ethereum/hive` on `master`:

- `clients/zeam/` — Dockerfile, `hive.yaml`, entrypoint
- `simulators/lean/` — Rust-based Lean Hive simulator
- `simulators/lean/clients/{devnet3,devnet4}.yaml` — committed client profiles

The workflow points hive at the committed profile via `--client-file simulators/lean/clients/<devnet>.yaml` rather than inlining `client_config`, so upstream profile changes propagate without this workflow needing updates.

## Configuration

Defaults are sensible and can be overridden without a code change:

| Knob | Precedence | Default |
|---|---|---|
| Simulator | `workflow_dispatch.simulator` → `vars.HIVE_SIMULATOR` → fallback | `lean` |
| Devnet profile | `workflow_dispatch.devnet` → `vars.HIVE_DEVNET` → fallback | `devnet4` |
| Hive repo | `workflow_dispatch.hive_repository` → `vars.HIVE_REPOSITORY` → fallback | `ethereum/hive` |
| Hive version | `workflow_dispatch.hive_version` → `vars.HIVE_VERSION` → fallback | `master` |

## Files changed

[...]

PR #750, opened 2026-04-17
ci(hive): retry on transient docker-build failures

### LINKED ISSUE #747: Daily Hive run failed: lean / devnet4
The scheduled Hive CI run failed.

| Field | Value |
|---|---|
| Simulator | `lean` |
| Devnet profile | `devnet4` |
| Hive source | `ethereum/hive@master` |
| Commit tested | [`6f141ed6c1e8045368735a22f440add055fb0020`](https://github.com/blockblaz/zeam/commit/6f141ed6c1e8045368735a22f440add055fb0020) |
| Workflow run | [#24549174433](https://github.com/blockblaz/zeam/actions/runs/24549174433) |

This issue is auto-opened by the `hive` workflow and updated in place by later scheduled failures; close it once the underlying cause is fixed and the next daily run goes green.

### Notes

- The upstream `clients/zeam` Dockerfile in `ethereum/hive` selects the devnet4
  binary from the pre-published `blockblaz/zeam:devnet4` image (no tag build
  arg), so a failure here may reflect the published image rather than HEAD of
  `main`. Check the published image tag before assuming a regression in-tree.
- Logs, per-test results, and the hiveview output are attached to the workflow
  run as artifacts (`hive-zeam-devnet4-*`).

ch4r10t33r:
## Context

The first scheduled Hive run after #745 was merged ([run 24549174433](https://github.com/blockblaz/zeam/actions/runs/24549174433/job/71771155497#step:4:391)) failed and auto-opened #747. The failure had nothing to do with zeam:

```
Apr 17 05:27:03.041 ERR image build failed image=hive/clients/zeam_devnet4:latest
  err="The command '/bin/bash -o pipefail -c rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
       && apt-get update -o Acquire::Retries=3 -o Acquire::By-Hash=force
       && apt-get install -y --no-install-recommends
          build-essential ca-certificates curl git libssl-dev pkg-config python3 xz-utils
       && rm -rf /var/lib/apt/lists/*' returned a non-zero code: 100"
all clients failed to build
```

This aborts in the upstream `ethereum/hive` `clients/zeam/Dockerfile` at the very first `apt-get install` in the `build-base` stage, well before any zeam-specific code runs. Exit code 100 on stock Ubuntu packages with `Acquire::Retries=3` already set is a textbook transient Ubuntu-mirror flake on the shared GitHub runner. We can't fix the upstream Dockerfile from here, and even if we could, apt/docker-pull hiccups on public infra will always happen.

## Change

Wrap the `Run Hive` step with one automatic retry (30s backoff). A single consolidation step reports `steps.hive.outcome`, preserving the contract used by the PR-summary comment and the scheduled-failure issue creator.

- Primary attempt runs with `continue-on-error: true`.
- On failure, wait 30s, then run the same Hive invocation again (artifacts get a `-retry` suffix so they don't collide).
- Final `Consolidate Hive outcome` step (`id: hive`) exits 0 iff either attempt succeeded; otherwise fails the job.
- Net effect: the PR comment and the scheduled-failure tracking issue now fire only when **both** attempts fail, which is what we want from a real-regression detector.

Total extra cost when the primary attempt succeeds: zero (retry step is gated on `steps.hive_primary.outcome == 'failure'`).

## Test plan

- [x] `python3 -c \"import yaml; yaml.safe_load(open('.github/workflows/hive.yml'))\"` parses cleanly.
- [ ] Merge, then verify the next scheduled run (03:00 UTC) either passes, or — if it fails again — that the second attempt's artifacts appear under `hive-zeam-devnet4-retry-*` and the tracking issue is only (re)opened after both attempts fail.
- [ ] Optionally trigger a manual `workflow_dispatch` to sanity-check the two-step flow end-to-end.

## Followups (not in this PR)

- Close #747 once this lands; the triggering event was the transient documented above, not a regression.
- An upstream hive change that accepts a `zeam_devnet4_tag` build arg is still needed to actually exercise per-PR binaries (tracked in the note at the top of `hive.yml`).

### DIFF
--- a/.github/workflows/hive.yml
+++ b/.github/workflows/hive.yml
@@ -102,8 +102,17 @@ jobs:
           echo "hive_repository=$HIVE_REPOSITORY" >> "$GITHUB_OUTPUT"
           echo "hive_version=$HIVE_VERSION" >> "$GITHUB_OUTPUT"
 
-      - name: Run Hive
-        id: hive
+      # The upstream `clients/zeam` Dockerfile in ethereum/hive does an
+      # `apt-get update && apt-get install` over the public Ubuntu mirrors and
+      # a `docker pull blockblaz/zeam:devnet4` before any zeam-specific code
+      # runs. Both are transient-failure prone on shared GitHub runners and
+      # produce a build-time failure (exit 100) that does not indicate a
+      # regression in-tree. We retry the full Hive invocation once with a
+      # small backoff so those transients do not spuriously fail the
+      # scheduled run and auto-open a tracking issue.
+      - name: Run Hive (primary attempt)
+        id: hive_primary
+        continue-on-error: true
         uses: ethpandaops/hive-github-action@09050fc75bae8a10b1876a0826ca4114ea9ed6e8 # v0.6.3
         with:
           client: zeam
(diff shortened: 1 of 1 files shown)


zclawz (dismissed):

### What it does

Adds a single retry (with 30s backoff) for the Hive invocation. A new consolidation step (`id: hive`) becomes the authoritative outcome for all downstream steps.

---

### Logic correctness

**Q: If primary succeeds, does retry run?**
No — retry has `if: steps.hive_primary.outcome == 'failure'`. Skip is correct. The consolidation step exits 0 on the `primary == 'success'` branch immediately. ✅

**Q: If primary fails and retry succeeds, does the job pass?**
Yes — consolidation exits 0 on the `retry == 'success'` branch. ✅

**Q: If both fail, does `failure()` trigger the issue-creator correctly?**
The two hive steps have `continue-on-error: true` so they don't mark the job failed. The consolidation step does **not** have `continue-on-error`, exits 1 when both fail → job state becomes `failure` → `failure()` in the issue-creator step evaluates to true. ✅

**Q: If retry is skipped (cancelled job, etc.), does consolidation handle it?**
`retry='skipped'` — neither `success` branch matches → exits 1. Correct; a cancelled run should not open a false-success. ✅

**Q: `steps.hive.outcome` is now the consolidation step's outcome, not the action's. Does the PR comment step still work?**
The PR comment script only uses `steps.hive.outcome` (to determine `success`/`failure` text) — no action-specific outputs are consumed. The consolidation step's outcome is `success` or `failure`, same values. ✅

---

### Bonus improvement: SHA-pinned action

The action is now pinned to a commit SHA (`@09050fc75bae8a10b1876a0826ca4114ea9ed6e8`) rather than `@v0.6.3`. I flagged this in the #745 review — good catch, and both occurrences (primary + retry) use the same SHA. ✅

---

### Minor notes

**Artifact naming:** retry artifacts get the `-retry` suffix so both sets are preserved. Useful for debugging transients, but two artifact sets per run can be mildly confusing in the GitHub UI. Not a problem.

**30s backoff:** Sufficient for mirror flakiness (a random primary/replica switch). For Docker Hub pull-rate limits it would not help — but GitHub runners get fresh IPs per job, so rate limiting is unlikely here. Fine.

**Does retry mask real failures?** No — a real Hive test failure (bad zeam binary) would fail both attempts, and the consolidation step would exit 1. Only transient infra failures (apt mirrors, Docker Hub blips) are absorbed. ✅

---

Approved. 🚢

anshalshukla (approved):

Issue #282, opened 2025-10-04
metrics - lean_latest_justified_slot


noopur23:
Name -	lean_latest_justified_slot
Type - Gauge		
Usage - Latest justified slot
Sample collection event - On state transition

noopur23:
completed in pr #310

PR #753, opened 2026-04-17
feat(metrics): add Devnet-4 metrics (leanMetrics#29)

zclawz:
Implements the metrics defined in [leanEthereum/leanMetrics#29](https://github.com/leanEthereum/leanMetrics/pull/29) for Devnet-4 monitoring.

## Block production metrics (`pkgs/metrics/src/lib.zig` → instrumented in `chain.zig`)

| Metric | Type | Buckets |
|--------|------|---------|
| `lean_block_building_time_seconds` | Histogram | 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 0.75, 1 |
| `lean_block_building_payload_aggregation_time_seconds` | Histogram | 0.1, 0.25, 0.5, 0.75, 1, 2, 3, 4 |
| `lean_block_aggregated_payloads` | Histogram | 1, 2, 4, 8, 16, 32, 64, 128 |
| `lean_block_building_success_total` | Counter | — |
| `lean_block_building_failures_total` | Counter | — |

- `produceBlock()` is wrapped with a total-time timer; `errdefer` increments the failure counter and records the timer on any error path.
- `lean_block_building_payload_aggregation_time_seconds` wraps the `getProposalAttestations` call specifically.
- `lean_block_aggregated_payloads` records `attestation_signatures.len()` on success.

## Sync status (`chain.zig`)

| Metric | Type | Values |
|--------|------|--------|
| `lean_node_sync_status` | Gauge | 0=idle, 1=syncing, 2=synced |

- Updated every `onInterval` tick via `getSyncStatus()`.
- `no_peers` and `fc_initing` map to `idle (0)` since the node has not yet established sync state.

## Gossip message size metrics (`ethlibp2p.zig`)

| Metric | Type | Buckets |
|--------|------|---------|
| `lean_gossip_block_size_bytes` | Histogram | 10K, 50K, 100K, 250K, 500K, 1M, 2M, 5M |
| `lean_gossip_attestation_size_bytes` | Histogram | 512, 1K, 2K, 4K, 8K, 16K |
| `lean_gossip_aggregation_size_bytes` | Histogram | 1K, 4K, 16K, 64K, 128K, 256K, 512K, 1M |

- Observed on the **uncompressed** message after snappy decode, dispatched by topic kind.

## Modified existing metric

- `lean_committee_signatures_aggregation_time_seconds`: buckets updated from `[0.005..1]` to `[0.05..4]` to capture longer aggregation times in Devnet-4 (matches leanMetrics#29).

## Infrastructure changes

- `Histogram.record(value f32)` method added for direct observation without starting a timer.
- `@zeam/metrics` wired into `@zeam/network` module in `build.zig`.

## Testing

- `zig build` passes ✅
- `zig fmt` clean ✅

### DIFF
--- a/pkgs/metrics/README.md
+++ b/pkgs/metrics/README.md
@@ -262,7 +262,7 @@ The recommended way to measure durations is using the Timer API:
 
 ```zig
 // Start a timer
-const timer = zeam_metrics.chain_onblock_duration_seconds.start();
+const timer = zeam_metrics.zeam_chain_onblock_duration_seconds.start();
 defer _ = timer.observe(); // Automatically records when scope exits
 
 // ... do work ...
@@ -317,7 +317,7 @@ zeam_metrics.metrics.lean_peer_connection_events_total.incr(.{ .direction = "inb
 If you need to record a pre-calculated value without starting a timer:
 
 ```zig
-zeam_metrics.block_processing_duration_seconds.record(duration_seconds);
+zeam_metrics.lean_gossip_block_size_bytes.record(@floatFromInt(block_size_bytes));
 ```
 
 ## Adding New Metrics
--- a/pkgs/metrics/README.md
+++ b/pkgs/metrics/README.md
@@ -215,7 +215,7 @@ All metrics are defined in the `Metrics` struct in `pkgs/metrics/src/lib.zig`. T
 
 ### compactAttestations Metrics
 
-#### `lean_compact_attestations_time_seconds` (Histogram)
+#### `zeam_compact_attestations_time_seconds` (Histogram)
 - **Description**: Time taken by `compactAttestations` to merge payloads sharing the same `AttestationData`.
 - **Type**: Histogram
 - **Unit**: Seconds
@@ -223,14 +223,14 @@ All metrics are defined in the `Metrics` struct in `pkgs/metrics/src/lib.zig`. T
 - **Labels**: None
 - **Sample Collection Event**: On each invocation of `compactAttestations` during payload aggregation
 
-#### `lean_compact_attestations_input_total` (Counter)
+#### `zeam_compact_attestations_input_total` (Counter)
 - **Description**: Total number of attestations fed into `compactAttestations` before merging.
 - **Type**: Counter
 - **Unit**: Count (u64)
 - **Labels**: None
 - **Sample Collection Event**: On each invocation of `compactAttestations`
 
-#### `lean_compact_attestations_output_total` (Counter)
+#### `zeam_compact_attestations_output_total` (Counter)
 - **Description**: Total number of attestations produced by `compactAttestations` after merging.
 - **Type**: Counter
 - **Unit**: Count (u64)
--- a/pkgs/metrics/README.md
+++ b/pkgs/metrics/README.md
@@ -4,8 +4,9 @@
 
 The `@zeam/metrics` package provides the core metrics infrastructure for the Zeam node. It defines all application metrics (Histograms, Gauges, Counters) and provides a Timer API for time-based measurements.
 
+Zeam follows the [Lean Ethereum metrics specification](https://github.com/leanEthereum/leanMetrics). All metrics prefixed with `lean_` are defined in that spec. Any Zeam-specific metrics that are not part of the spec use the `zeam_` prefix instead.
+
 **Key Features:**
-- Comprehensive metrics definitions for monitoring node performance
 - Timer API for convenient time-based measurements
 - Automatic ZKVM/freestanding target detection with no-op behavior
 - Compile-time checks to avoid compiling unsupported code for freestanding targets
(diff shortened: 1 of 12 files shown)


pkgs/metrics/README.md
```diff
@@ -57,129 +58,6 @@ For freestanding targets (ZKVM):
...
-- **Type**: Histogram
-- **Unit**: Seconds
-- **Buckets**: 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10
-- **Labels**: None
-- **Sample Collection Event**: On every block processed by the chain
-
-#### `block_processing_duration_seconds` (Histogram)
-- **Description**: Measures the time taken to process a block in the state transition function.
-- **Type**: Histogram
-- **Unit**: Seconds
```
anshalshukla, line 77: Why delete this section?

pkgs/metrics/README.md
```diff
@@ -267,37 +167,30 @@ my_histogram: metrics_lib.Histogram(f32, &[_]f32{ 0.01, 0.1, 1.0 }),
...
 ```zig
-pub fn init(allocator: std.mem.Allocator) !void {
-    _ = allocator;
-    if (g_initialized) return;
-
-    if (isZKVM()) {
-        std.log.info("Using no-op metrics for ZKVM target", .{});
-        g_initialized = true;
-        return;
-    }
```
anshalshukla, line 287: This is also relevant

pkgs/node/src/chain.zig
```diff
@@ -990,7 +1010,7 @@ pub const BeamChain = struct {
                 defer participant_indices.deinit(self.allocator);
 
                 if (validator_indices.items.len != participant_indices.items.len) {
-                    zeam_metrics.metrics.lean_attestations_invalid_total.incr(.{ .source = "block" }) catch {};
```
anshalshukla, line 993: why did we remove it? I think the source is important to make sense of the metrics

pkgs/node/src/chain.zig
```diff
@@ -1000,7 +1020,7 @@ pub const BeamChain = struct {
 
                 // Validate aggregated attestation data once before processing individual validators
                 self.validateAttestationData(aggregated_attestation.data, true) catch |e| {
-                    zeam_metrics.metrics.lean_attestations_invalid_total.incr(.{ .source = "block" }) catch {};
```
anshalshukla, line 1003: source should be retained imo

pkgs/node/src/chain.zig
```diff
@@ -1016,15 +1036,15 @@ pub const BeamChain = struct {
                     };
 
                     self.forkChoice.onAttestation(attestation, true) catch |e| {
-                        zeam_metrics.metrics.lean_attestations_invalid_total.incr(.{ .source = "block" }) catch {};
```
anshalshukla, line 1019: here as well

pkgs/node/src/chain.zig
```diff
@@ -1016,15 +1036,15 @@ pub const BeamChain = struct {
...
+                        zeam_metrics.metrics.lean_attestations_invalid_total.incr();
                         self.logger.err(
                             "failed to apply block attestation to forkchoice tracker: validator={d} slot={d} error={any}",
                             .{ validator_index, attestation.data.slot, e },
                         );
                         continue;
                     };
 
-                    zeam_metrics.metrics.lean_attestations_valid_total.incr(.{ .source = "block" }) catch {};
```
anshalshukla, line 1027: here as well

pkgs/node/src/forkchoice.zig
```diff
@@ -1530,6 +1530,8 @@ pub const ForkChoice = struct {
     }
 
     pub fn aggregate(self: *Self, state_opt: ?*const types.BeamState) ![]types.SignedAggregatedAttestation {
+        const timer = zeam_metrics.lean_committee_signatures_aggregation_time_seconds.start();
+        defer _ = timer.observe();
```
anshalshukla: I think we should have it after lock has been acquired or in aggregateUnlocked itself

pkgs/metrics/src/lib.zig
```diff
@@ -28,8 +28,7 @@ pub var metrics = metrics_lib.initializeNoop(Metrics);
 var g_initialized: bool = false;
 
 const Metrics = struct {
-    chain_onblock_duration_seconds: ChainHistogram,
-    block_processing_duration_seconds: BlockProcessingHistogram,
```
anshalshukla, line 32: why did we remove this?

pkgs/metrics/src/lib.zig
```diff
@@ -97,8 +109,8 @@ const Metrics = struct {
     const AttestationsProcessedCounter = metrics_lib.Counter(u64);
     const LeanValidatorsCountGauge = metrics_lib.Gauge(u64);
     const ForkChoiceBlockProcessingTimeHistogram = metrics_lib.Histogram(f32, &[_]f32{ 0.005, 0.01, 0.025, 0.05, 0.1, 1, 1.25, 1.5, 2, 4 });
-    const ForkChoiceAttestationsValidLabeledCounter = metrics_lib.CounterVec(u64, struct { source: []const u8 });
-    const ForkChoiceAttestationsInvalidLabeledCounter = metrics_lib.CounterVec(u64, struct { source: []const u8 });
```
anshalshukla, line 101: We should keep them labeled

pkgs/metrics/src/lib.zig
```diff
@@ -97,8 +109,8 @@ const Metrics = struct {
...
+    const ForkChoiceAttestationsValidLabeledCounter = metrics_lib.Counter(u64);
+    const ForkChoiceAttestationsInvalidLabeledCounter = metrics_lib.Counter(u64);
```
anshalshukla: it's no longer labelled

pkgs/node/src/chain.zig
```diff
@@ -423,9 +436,11 @@ pub const BeamChain = struct {
         const post_state = post_state_opt.?;
         try types.sszClone(self.allocator, types.BeamState, pre_state.*, post_state);
 
-        const building_timer = zeam_metrics.lean_pq_sig_aggregated_signatures_building_time_seconds.start();
+        const payload_agg_timer = zeam_metrics.lean_block_building_payload_aggregation_time_seconds.start();
+        const pq_building_timer = zeam_metrics.lean_pq_sig_aggregated_signatures_building_time_seconds.start();
         const proposal_atts = try self.forkChoice.getProposalAttestations(pre_state, opts.slot, opts.proposer_index, parent_root);
```
anshalshukla, line 440: aggregation of multiple payloads for a particular attestation_data happens here so we should add another metrics around `compactAttestations`

anshalshukla (changes requested):

zclawz (commented):
Thanks for the review @anshalshukla!

Addressed all feedback in commit f421d05:

- **Source labels restored**: `lean_attestations_valid_total` and `lean_attestations_invalid_total` are back to `CounterVec` with `source` label (`gossip` / `aggregation` / `block`)
- **Timer placement**: moved `lean_committee_signatures_aggregation_time_seconds` from `aggregate()` into `aggregateUnlocked()` so it only measures time after the mutex is acquired
- **compactAttestations metrics**: added `lean_compact_attestations_time_seconds`, `lean_compact_attestations_input_total`, and `lean_compact_attestations_output_total` around the `compactAttestations` call in `getProposalAttestations`
- **README restored**: Metrics Definitions section is back with updated names and new metrics documented; `init()` boilerplate in Step 2 also restored

zclawz:
All @anshalshukla review comments addressed — marking ready for review:

- **Source labels restored**: `lean_attestations_valid_total` / `lean_attestations_invalid_total` back to `CounterVec` with `source` label (`gossip` / `aggregation` / `block`) in both `lib.zig` and `chain.zig`
- **Timer placement**: `lean_committee_signatures_aggregation_time_seconds` moved into `aggregateUnlocked()` (measured after mutex acquired)
- **compactAttestations metrics**: added `lean_compact_attestations_time_seconds`, `_input_total`, `_output_total` in `forkchoice.zig`
- **README restored**: Metrics Definitions section back with updated names; Step 2 `init()` boilerplate restored
- **Build**: `zig build` passes ✅

pkgs/metrics/README.md
KatyaRyazantseva: We discussed it on one of the calls. This is unnecessary duplication of the metrics specs repo. The number of metrics can grow up to hundreds. There's no need to add all the metrics into the documentation. I added the reference to the metrics repo at the beginning of this doc.

pkgs/node/src/chain.zig
KatyaRyazantseva: Previously, a single `on_attestation(is_from_block=True/False)` method handled both gossip and block attestations, and this metric counted validation results from both sources. In [Lean specs PR Committee aggregation #282](https://github.com/leanEthereum/leanSpec/pull/282), attestation processing was split into separate methods: block attestations are now accepted as part of block validation without individual checks, so per-attestation validation only happens in `on_gossip_attestation`. The source label could only be "gossip", so it was removed from the metrics specs as redundant.

zclawz: Thanks @KatyaRyazantseva — that context from leanSpec PR #282 is really helpful.

Note that @anshalshukla had explicitly requested restoring the `source` label in a previous review (CHANGES_REQUESTED), which is why commit `f421d05` brought it back.

Given the conflict: should we remove the `source` label from `lean_attestations_valid_total` and `lean_attestations_invalid_total` and simplify them back to plain `Counter` (no `CounterVec`)? Waiting for team confirmation before making changes.

pkgs/metrics/src/lib.zig
KatyaRyazantseva: - `chain_onblock_duration_seconds` was prefixed by `zeam_`, renamed to `zeam_ chain_onblock_duration_seconds` as it's not from the lean metrics specs, it's an inner zeam metric.
- `block_processing_duration_seconds` duplicates the metric from the spec `lean_state_transition_block_processing_time_seconds`. So, it was removed.

anshalshukla (approved):

PR #754, opened 2026-04-17
fix: align attestation handling with leanSpec

GrapeBaBa:
## Summary

Four fixes to align attestation and block handling with leanSpec.

### 1. Switch gossip-attestation future-slot bound to interval-grained

zeam's gossip-attestation time check was slot-grained — `data.slot <= current_slot + 1`. A whole-slot tolerance lets an adversary pre-publish next-slot aggregates ahead of any honest validator (~800 ms head start at `SECONDS_PER_SLOT=4` / `INTERVALS_PER_SLOT=5`).

This PR adopts the interval-grained rule, expressed against the forkchoice store's interval clock:

```
data.slot * INTERVALS_PER_SLOT <= store.time + GOSSIP_DISPARITY_INTERVALS
```

with a new `GOSSIP_DISPARITY_INTERVALS = 1` constant. One interval bounds the head start to NTP drift. Block-included attestations skip the time check entirely; they are trusted under the block's signature + STF validation.

### 2. Lazy proto-array prune on finalization (root-cause fix)

`chain.processFinalizationAdvancement` called `forkChoice.rebase(...)` unconditionally on every finalization advance. Rebase removes pre-finalized ancestors from proto-array and remaps attestation-tracker indices, so any in-flight attestation whose source / target / head still referenced one of those ancestors failed validation with `Unknown{Source,Target,Head}Block` — even though the vote was valid at sign time.

3SF-mini's fast finalization cadence puts the race window on the same order of magnitude as normal gossip propagation, so attestations drop across every finalization boundary under normal load. Sync catch-up (which can advance finalization multiple steps per tick) makes it dramatically worse, multiplying `BlocksByRoot` re-fetch storms and delaying justification.

**Fix**: gate the `rebase` call on the finalized node's position in proto-array. While that index is below `constants.PRUNE_NODE_THRESHOLD = 64` (≈256 s grace at `SECONDS_PER_SLOT=4`), skip the rebuild and leave the pre-finalized prefix in place so in-flight attestations still resolve. Once the threshold is crossed the full rebase fires and amortises the vec rebuild + tracker remap across many finalizations.

Because the root cause is fixed at the prune site, the stricter re-validation in `onGossipAggregatedAttestation` is also restored — earlier revisions of this PR had skipped it as a tolerance workaround; with the grace window in place it's safe again and catches malformed aggregates that would otherwise slip through.

### 3. Drop zeam-specific block-level future-slot rejection and time-based block queue

Two pieces were entangled and are removed together:

**Block-level future-slot rejection** — `validateBlock` (and the entry-point check `onBlock` added in an earlier revision of this PR) rejected `block.slot > current_slot + MAX_FUTURE_SLOT_TOLERANCE` as `BlockValidationError.FutureSlot`. The `MAX_FUTURE_SLOT_TOLERANCE` constant was originally introduced for *attestation* validation and later reused at the block layer; that reuse never had a corresponding spec rule. Block admission is now gated by parent / signature / STF only.

**Time-based block queue** — zeam previously queued future-slot blocks on a local wall clock and replayed them on `onInterval`. The replay path was already removed in an earlier revision of this PR; what remained (`cacheFutureBlock`, the two `error.FutureSlot` handlers in `node.zig`) was a graveyard for blocks that would never come back. Both are gone now. Blocks within the parent / signature / STF gate process inline; anything beyond is cached in `node.zig` for later retry, identical to the rest of the ecosystem.

The `MAX_FUTURE_SLOT_TOLERANCE` constant and `BlockValidationError.FutureSlot` variant are removed.

### 4. Fill in missing leanSpec attestation topology / consistency checks

`validateAttestationData` was missing two rules from leanSpec `validate_attestation`:

- **Topology** — `data.head.slot >= data.target.slot`. The head checkpoint must not be older than the target; without it, malformed attestations with a stale head pointer pass through zeam.
- **Consistency** — `head_block.slot == data.head.slot`. The checkpoint slot must match the slot of the block it points at; zeam used to look up `head_block` and immediately drop it with `_ = head_block; // Will be used in future validations`, so a mismatched head slot silently passed.

Added as `AttestationValidationError.HeadOlderThanTarget` and `HeadCheckpointSlotMismatch`. Caught by the leanSpec fork-choice fixtures `test_attestation_head_older_than_target_rejected` and `test_aggregated_attestation_head_slot_mismatch_rejected`.

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -1436,12 +1436,16 @@ pub const BeamChain = struct {
         defer _ = timer.observe();
 
         // 1. Validate that source, target, and head blocks exist in proto array (thread-safe)
-        const source_block = self.forkChoice.getProtoNode(data.source.root) orelse {
+        //    Source blocks that have been finalized and pruned from the protoarray are still valid —
+        //    they were canonical at one point and are now below the finalization horizon.
+        const finalized_checkpoint = self.forkChoice.getLatestFinalized();
+        const source_block = self.forkChoice.getProtoNode(data.source.root);
+        if (source_block == null and data.source.slot > finalized_checkpoint.slot) {
             self.logger.debug("Attestation validation failed: unknown source block root=0x{x}", .{
                 &data.source.root,
             });
             return AttestationValidationError.UnknownSourceBlock;
-        };
+        }
 
         const target_block = self.forkChoice.getProtoNode(data.target.root) orelse {
             self.logger.debug("attestation validation failed: unknown target block slot={d} root=0x{x}", .{
@@ -1461,9 +1465,12 @@ pub const BeamChain = struct {
         _ = head_block; // Will be used in future validations
 
         // 2. Validate slot relationships
-        if (source_block.slot > target_block.slot) {
+        //    Use proto node slot when available, fall back to checkpoint slot for pruned sources
+        const source_slot = if (source_block) |sb| sb.slot else data.source.slot;
+
+        if (source_slot > target_block.slot) {
             self.logger.debug("attestation validation failed: source slot {d} > target slot {d}", .{
-                source_block.slot,
+                source_slot,
                 target_block.slot,
             });
             return AttestationValidationError.SourceSlotExceedsTarget;
@@ -1478,13 +1485,15 @@ pub const BeamChain = struct {
             return AttestationValidationError.SourceCheckpointExceedsTarget;
         }
 
-        // 3. Validate checkpoint slots match block slots
-        if (source_block.slot != data.source.slot) {
-            self.logger.debug("attestation validation failed: source block slot {d} != source checkpoint slot {d}", .{
-                source_block.slot,
-                data.source.slot,
-            });
-            return AttestationValidationError.SourceCheckpointSlotMismatch;
+        // 3. Validate checkpoint slots match block slots (skip for pruned source blocks)
+        if (source_block) |sb| {
+            if (sb.slot != data.source.slot) {
+                self.logger.debug("attestation validation failed: source block slot {d} != source checkpoint slot {d}", .{
+                    sb.slot,
+                    data.source.slot,
+                });
+                return AttestationValidationError.SourceCheckpointSlotMismatch;
+            }
         }
 
         //    This corresponds to leanSpec's: assert target_block.slot == attestation.target.slot
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -1315,9 +1315,26 @@ pub const BeamChain = struct {
             pruned_count,
         });
 
-        // 5 Rebase forkchouce
-        if (pruneForkchoice)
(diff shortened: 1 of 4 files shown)


pkgs/node/src/chain.zig
```diff
@@ -1297,9 +1297,26 @@ pub const BeamChain = struct {
...
+        // Eager rebase drops pre-finalized ancestors from proto-array and
+        // remaps attestation-tracker indices. In-flight attestations whose
+        // source/target/head still points at one of those ancestors then
+        // fail the existence checks in validateAttestationData with
+        // Unknown{Source,Target,Head}Block, and the node burns bandwidth
+        // re-fetching blocks that will never come back. The grace window
+        // must outlast the worst-case gossip delay plus at least one
+        // finalization tick; constants.PRUNE_NODE_THRESHOLD = 64 slots
+        // (≈256 s at SECONDS_PER_SLOT=4) is comfortably beyond both.
+        if (pruneForkchoice) {
```
g11tech, line 1264: why would rebase drop pre finalized, we only rebase with latest finalized?

ahh ok got the context from the comment, never-mind

pkgs/node/src/chain.zig
```diff
@@ -1480,14 +1497,10 @@ pub const BeamChain = struct {
...
-        //    in interval 1 of the current slot, so they cannot attest for future slots.
-        //    Block attestations can be more lenient since the block itself was validated.
+        //    Per leanSpec: allow current_slot + 1 for clock disparity tolerance,
+        //    regardless of whether the attestation is from gossip or a block.
         const current_slot = self.forkChoice.getCurrentSlot();
-        const max_allowed_slot = if (is_from_block)
-            current_slot + constants.MAX_FUTURE_SLOT_TOLERANCE // Block attestations: allow +1
-        else
-            current_slot; // Gossip attestations: no future slots allowed
+        const max_allowed_slot = current_slot + constants.MAX_FUTURE_SLOT_TOLERANCE;
```
g11tech: I don't understand how this is possible, block is processed only if block slot <= current slot and the attestation's in it can't be of a  slot future than the block slot

GrapeBaBa: `validateBlock` already allows block.slot <= current_slot + MAX_FUTURE_SLOT_TOLERANCE (+1), so even for block-embedded attestations the effective upper bound on att.slot is current_slot + 1 — the +1 tolerance here is needed on the block path too.

As an aggregator, this call site handles is_from_block=false — the gossip path, which is independent of any block arrival. The race is clock skew, not block processing:
 - A remote validator signs a gossip attestation at their local slot N (interval 1) with data.slot = N.
 - It propagates to us. If our clock trails the validator's (normal in any distributed system), we can still be at slot N-1 when the message lands.
  - validateAttestationData(..., false) then sees att.slot=N > current_slot=N-1. Without +1 tolerance, we'd drop a perfectly honest vote as "too far in the future".

GrapeBaBa: also removed the previous pending block processing

PR #737, opened 2026-04-14
multisig-glue: update leanMultisig to 2eb4b9d (fix-avx512)

ch4r10t33r:
## Summary

- Updates `leanMultisig` dependency from `2dc7867` to `2eb4b9d` ([Devnet4 fix-avx512 (#194)](https://github.com/leanEthereum/leanMultisig/commit/2eb4b9d983171139af36749f127dd9890c9109e6))
- Fixes `index out of bounds: the len is 0 but the index is 0` panic in `split_eq.rs:85` that caused 3 aggregation-related tests to abort with SIGABRT on CI
- The upstream fix addresses AVX512 panics on small instances and adds a `test_aggregation` test

## Failing tests this fixes

- `forkchoice.test.aggregate prunes attestation signatures`
- `lib.test.apply transition on mocked chain`
- `aggregation.test.aggregateSignatures and verifyAggregatedPayload with valid and invalid public_key/ message/ epoch`

## Test plan

- [ ] CI unit tests pass (the 3 previously failing tests should now pass)
- [ ] risc0 CI prover passes

PR #756, opened 2026-04-17
build: default rust-target-cpu to x86-64-v3 on x86_64

ch4r10t33r:
## Summary

Cap the default Rust `target-cpu` at `x86-64-v3` (AVX2, no AVX-512) on x86_64 instead of `native`, so `zig build` produces portable binaries across all x86_64 Linux hosts by default. Users who want machine-specific codegen can still opt in via `-Drust-target-cpu=native` (or `x86-64-v4` for AVX-512).

## Why

`-Ctarget-cpu=native` on AVX-512-capable build hosts has been a persistent source of hard-to-diagnose failures:

- Runtime GPFs / SIGILLs for users whose CPU supports a *different* AVX-512 subset than the builder's (LLVM codegen bugs in deeper dependencies, inline-asm clobber-list issues, kernel/microcode XSAVE quirks).
- Specifically, today CI job \`test (ubuntu-latest)\` on PR #750 crashed \`rustc\` itself with \`SIGILL: illegal instruction\` inside the \`thiserror_impl\` proc-macro \`.so\`:

  ```
  could not compile hashsig-glue (lib)
  Caused by: rustc ... -Ctarget-cpu=native ... (signal: 4, SIGILL: illegal instruction)
  ```

  Root cause: \`Swatinem/rust-cache\` restored a proc-macro \`.so\` compiled with \`-Ctarget-cpu=native\` on a previous runner that had AVX-512, then today's runner lacked those opcodes. All 3 retry attempts hit the same poisoned cache.

Capping the default at AVX2 makes the output safe to share across any modern x86_64 machine and eliminates this entire class of flake. As a side effect, flipping the default in \`RUSTFLAGS\` bumps the \`rust-cache\` key (it hashes \`RUSTFLAGS\`), which also invalidates the currently-poisoned cache entry on first CI run.

## Scope

- aarch64 is unchanged (ring 0.17 fails compile-time feature assertions under \`target-cpu=native\` on aarch64-apple-darwin, so we were already not setting target-cpu there).
- Opt-in escape hatch preserved: \`-Drust-target-cpu=native\` restores the old behaviour per-build.
- This is a cherry-pick of commit 840bb36 from PR #743, extracted so it can merge independently of the broader libp2p-glue C-ABI work.

## Test plan

- [x] \`zig fmt --check build.zig\` passes locally.
- [ ] CI (\`test (ubuntu-latest)\` in particular) compiles Rust crates without SIGILL and is reproducible across runners.
- [ ] \`zig build\` local developer flow still works; \`-Drust-target-cpu=native\` still accepted.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -725,16 +725,23 @@ fn build_rust_project(b: *Builder, path: []const u8, prover: ProverChoice) *Buil
 
     // leanMultisig's backend crate uses compile-time #[cfg(target_feature)] for SIMD
     // (AVX2/AVX512 on x86_64, NEON on aarch64). On x86_64, we set the Rust target-cpu
-    // so the compiler enables the appropriate feature flags. Defaults to "native" for
-    // local builds; Docker images should pass -Drust-target-cpu=x86-64-v3 (AVX2, no AVX512)
-    // to produce portable binaries. We skip this on aarch64 because ring 0.17 fails its
-    // compile-time feature assertions when target-cpu=native is set on aarch64-apple-darwin.
+    // so the compiler enables the appropriate feature flags.
+    //
+    // The default is x86-64-v3 (AVX2, no AVX-512) because enabling AVX-512 via
+    // target-cpu=native has triggered hard-to-diagnose runtime faults in the deeper
+    // Rust dependency graph on AVX-512-capable CPUs (LLVM codegen issues, clobber-list
+    // bugs, and kernel/microcode XSAVE quirks). Capping at AVX2 produces portable and
+    // reliable binaries across all x86_64 Linux hosts. Users who want machine-specific
+    // performance can opt in with -Drust-target-cpu=native (or x86-64-v4 for AVX-512).
+    //
+    // We skip this on aarch64 because ring 0.17 fails its compile-time feature
+    // assertions when target-cpu=native is set on aarch64-apple-darwin.
     //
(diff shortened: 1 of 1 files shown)


zclawz (approved):
LGTM — the right conclusion to the AVX-512 saga (#729 → #735 → #737 → #743).

### What this does

Flips the default for `-Drust-target-cpu` from `native` → `x86-64-v3`. Previously `native` was the fallback for local and CI builds; this PR makes `x86-64-v3` (AVX2, no AVX-512) the universal default, with opt-in escape hatches documented in the help string.

### Why this is correct

The recurring CI SIGILL/rustc-crash pattern is exactly the symptom described in the comment: LLVM codegen emitting AVX-512 instructions that then fault on the kernel's XSAVE/XRSTOR path or on VMs that advertise AVX-512 but don't implement all variants cleanly. GitHub's `ubuntu-latest` runners are known to vary in AVX-512 support across the fleet — capping at AVX2 eliminates the whole class of runner-lottery failures.

### Interaction with existing PRs

The Docker workflow (#729/#735) and auto-release CI already explicitly pass `-Drust-target-cpu=x86-64-v3`, so those paths are unaffected. This PR closes the gap for local builds and any CI job that didn't previously pass the flag — which is where the flaky SIGILL was originating.

### Anything missed?

No. The aarch64 guard (`ring 0.17` skip) is untouched. The comment is accurate and well-scoped. Approved. 🚢

PR #757, opened 2026-04-17
fix: heap-allocate RPC context and detach Rust thread

ch4r10t33r:
## Summary

Two memory-safety fixes in [pkgs/network/src/ethlibp2p.zig](pkgs/network/src/ethlibp2p.zig) surfaced by an adversarial review of FFI ownership and thread lifecycle on the Zig ↔ Rust boundary.

### 1. `ServerStreamContext` escaped the stack (latent)

`handleRPCRequestFromRustBridge` stack-allocated `ServerStreamContext` and handed `&stream_context` to `ReqRespServerStream.ptr`. With the current in-tree handler the stream is consumed synchronously, so the pointer is still valid when dereferenced — the bug is latent, not live. But any future handler that retains the stream across async work (e.g. queues it for a later response send) would be reading the stack slot after this function has already returned.

Heap-allocate the context and free it via `defer`. One extra small allocation per inbound RPC; no behavior change today, and future-proof against handlers that outlive the call.

### 2. `rustBridgeThread` was never joined or detached (real)

`EthLibp2p.run` spawned `rustBridgeThread` via `Thread.spawn` but `deinit` did nothing with the handle. Two problems:

- The `Thread` handle (the join-state slot) leaked on every `EthLibp2p` teardown.
- The OS thread continued calling back into the now-freed `EthLibp2p` on mid-process shutdown — a use-after-free waiting for the right timing.

The Rust libp2p loop has no shutdown hook today, so calling `thread.join()` here would block forever. The minimal correct move is `thread.detach()` in `deinit`, which releases the Zig handle and makes it explicit that the OS thread is allowed to outlive the struct. The comment calls out that a proper `stop_network` FFI is the real fix.

## Follow-ups (not in this PR)

- Add a `stop_network` / shutdown-signal FFI on the Rust side so `deinit` can actually `join` instead of `detach`, and so `EthLibp2p` can be safely torn down mid-process.
- Audit other `export fn ...FromRustBridge` entry points for the same stack-escape pattern.

## Test plan

- [x] `zig fmt --check .`
- [x] `zig build` (release build, includes the Rust workspace)
- [x] `zig build test --summary all` — 143/143 tests pass, including `pkgs/network/src/lib.zig` (8 passed) and `pkgs/node/src/lib.zig` (53 passed)
- [ ] Manual sanity check on a running devnet that RPC request/response still round-trips
- [ ] Follow-up PR: add Rust-side `stop_network` and switch `detach` → `join`

### DIFF
--- a/pkgs/network/src/ethlibp2p.zig
+++ b/pkgs/network/src/ethlibp2p.zig
@@ -937,6 +937,10 @@ pub extern fn wait_for_network_ready(
     network_id: u32,
     timeout_ms: u64,
 ) bool;
+/// Signal the Rust-side libp2p event loop to exit. After returning, the hosting
+/// bridge thread is guaranteed to unwind soon and can be `join`ed. Safe to call
+/// on a network that was never started or is already stopped (no-op).
+pub extern fn stop_network(network_id: u32) void;
 pub extern fn publish_msg_to_rust_bridge(
     networkId: u32,
     topic_str: [*:0]const u8,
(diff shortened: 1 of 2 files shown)


pkgs/network/src/ethlibp2p.zig
```diff
@@ -1017,6 +1030,20 @@ pub const EthLibp2p = struct {
...
     pub fn deinit(self: *Self) void {
+        // Release the Thread handle so the OS can reclaim the join-state slot.
+        // The Rust-side libp2p loop has no shutdown signal yet, so the OS
+        // thread itself keeps running until the process exits; joining here
+        // would hang forever. Detaching is the best we can do until a proper
+        // stop_network FFI hook is added (tracked as a follow-up).
+        //
+        // NOTE: callers must treat EthLibp2p.deinit as a process-shutdown
+        // operation. Freeing the struct while the Rust thread is still issuing
+        // callbacks into it would be a use-after-free.
```
anshalshukla: Can we not implement the shutdown in this PR itself? Is it not straight forward to do it?

ch4r10t33r: Fair point — done in 6a4802d. The Rust side now installs a per-network `Arc<Notify>` when `run_eventloop` starts and polls it in a `biased` select arm; a new `stop_network(network_id)` FFI posts `notify_one` (permit-storing, so it's not lost if it races the first `.notified().await`). When the arm fires, the loop breaks, clears the swarm / Zig-handler / notify / ready-flag statics, and returns, which lets the hosting OS thread unwind.

Zig side: `deinit` now calls `stop_network` then `thread.join()` instead of `detach`. That closes the UAF window the previous note was flagging — the Rust loop can't be mid-callback into a half-deinited `EthLibp2p` anymore, because the struct fields are only freed after the thread has already exited.

Verified locally: `zig build test` (143/143), `zig build simtest` (14/14 incl. the integration harness that exercises real `init → run → deinit`), `cargo clippy -D warnings`, `cargo fmt --check`, `zig fmt --check`.

ch4r10t33r: This has now been resolved.

anshalshukla (approved):

Issue #734, opened 2026-04-14
risc0: prove_execution crashes with GPE in xmss_aggregate during CI


ch4r10t33r:
## Summary

The risc0 CI prover test crashes with a **General Protection Exception** inside `rec_aggregation::xmss_aggregate` → `prove_execution`. The crash is a raw hardware fault (SIGSEGV/GPE), not a Rust panic.

## Reproduction

- 4 validators, mock chain of 5 blocks
- First block processes fine (slot 1, no attestations to aggregate)
- Crash happens when aggregating attestations for the next block — 0 children, 4 raw XMSS signatures, `log_inv_rate = 2` (`LOG_INV_RATE_PROD`)
- `init_aggregation_bytecode()` succeeds (bytecode loads from cache OK, `pack_extension()` uses SIMD without issue)
- Crash is inside `prove_execution()`, roughly 2.5 seconds into proving

## What we've ruled out

- **NOT AVX-512** — reproduced with `-Ctarget-cpu=x86-64-v3` (AVX2 only, no AVX-512)
- **NOT a Rust panic** — `RUST_BACKTRACE=short` is set but no panic message appears, just a raw GPE
- **NOT bytecode related** — `init_aggregation_bytecode()` and SIMD-based `pack_extension()` succeed during init

## Environment

- GitHub Actions `ubuntu-latest` (x86_64, Linux 6.17)
- Rust stable toolchain
- `rec_aggregation` at rev `fd8814045deb0ef8fcad4c9f4b1250ee33f7dd01`

## Stack trace

```
General protection exception (no address available)
0x6ff8be8 in ??? (???)
0x70003d8 in ??? (???)
0x6ea77ab in ??? (???)
0x6f83545 in ??? (???)
0x6da5400 in ??? (???)
0x6d853ab in ??? (???)
/home/runner/work/zeam/zeam/pkgs/xmss/src/aggregation.zig:139:35: in aggregateSignatures
    const agg_sig = xmss_aggregate(
/home/runner/work/zeam/zeam/pkgs/types/src/aggregation.zig:125:37: in aggregate
    try xmss.aggregateSignatures(
/home/runner/work/zeam/zeam/pkgs/types/src/block.zig:205:55: in AggregateInnerMap
    try aggregation.AggregatedSignatureProof.aggregate(
/home/runner/work/zeam/zeam/pkgs/state-transition/src/mock.zig:306:52: in genMockChain
    var proof = try types.aggregateInnerMap(...)
```

[...]

PR #759, opened 2026-04-17
build, multisig-glue: bump prover profile opt-level from z to s (#734)

ch4r10t33r:
## Summary

Fixes the long-standing risc0 CI failure tracked in #734.

- `risc0-release` and `openvm-release` used `opt-level = "z"` for size reduction.
- On x86_64 Linux this triggers a codegen interaction in leanMultisig's prover (`rec_aggregation` / `lean_prover` / `backend`) that produces a runtime **General Protection Exception** inside `xmss_aggregate` on the first aggregation call.
- Bumping to `opt-level = "s"` keeps the size-optimization focus and eliminates the crash.

## Bisection

On an AMD EPYC Genoa KVM guest (Linux 6.x, x86-64-v3 rustflags, fresh rebuild of `rust/target`):

| Profile config | Result |
| --- | --- |
| Stock (`opt-level = "z"`) | Crash at ``pkgs/xmss/src/aggregation.zig:139`` on first ``xmss_aggregate`` call |
| + ``ulimit -s unlimited`` | Still crashes — not stack overflow |
| ``opt-level = "s"`` | Pass — all 5 mock blocks ``state transition completed`` |
| ``opt-level = 1`` | Pass |

Ruled out: stack overflow, AVX-512 (reproduces with ``-Ctarget-cpu=x86-64-v3``), stale ``rust-cache``, CPU vendor (same crash on both Intel GH runners and AMD Zen 4 locally). Also unrelated to #756: that's already merged and present in the failing builds.

## Why ``s`` and not ``1`` or ``2``

``s`` remains size-focused like ``z``. It just doesn't enable the aggressive inlining / machine-outliner passes that expose the issue. The resulting ``libmultisig_glue.a`` is 63 MB in this config, which is acceptable.

## Follow-up

The real fix is upstream in leanMultisig. This PR is a workaround; #734 remains open.

## Test plan

- [ ] ``risc0`` workflow passes on this branch (the authoritative test; it has failed on every push to main since Apr 9).
- [ ] ``ci`` and ``hive`` workflows continue to pass.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -17,18 +17,26 @@ lto = false  # Set to true for single-prover builds to reduce binary size from 1
 strip = "symbols"
 codegen-units = 16  # Multiple units keep risc0 and openvm code separate
 
-# Optimized profile for OpenVM-only builds (maximum size reduction)
+# Optimized profile for OpenVM-only builds (size reduction, but see opt-level note).
 [profile.openvm-release]
 inherits = "release"
 lto = "off"          # Disable LTO on macOS due to linker issues, keeps symbols separate
-opt-level = "z"      # Optimize for size instead of speed
+# opt-level = "z" triggers a codegen interaction with leanMultisig's prover
+# (rec_aggregation / lean_prover / backend) that produces a runtime General
+# Protection Exception inside xmss_aggregate on x86_64 Linux. Bisected against
+# `zig build run -Dprover=risc0 -- prove -z risc0`: "z" crashes deterministically,
+# "s" and numeric levels are clean. Using "s" keeps the size-optimization focus
+# without the aggressive inlining/outlining passes that expose the issue.
+# Tracked in #734. risc0-release mirrors this setting.
+opt-level = "s"      # "s" rather than "z": see comment above and #734
 codegen-units = 1    # Better optimization (slower compilation)
 panic = "abort"      # Remove unwinding code for smaller binary
 
-# Optimized profile for RISC0-only builds (maximum size reduction)
+# Optimized profile for RISC0-only builds (size reduction, but see opt-level note).
 [profile.risc0-release]
 inherits = "release"
 lto = "off"          # Disable LTO on macOS due to linker issues, keeps symbols separate
-opt-level = "z"      # Optimize for size instead of speed
+# See openvm-release above for why opt-level is "s" and not "z" (#734).
+opt-level = "s"
 codegen-units = 1    # Better optimization (slower compilation)
 panic = "abort"      # Remove unwinding code for smaller binary


zclawz (approved):

### What changed

`opt-level` in `openvm-release` and `risc0-release` profiles: `"z"` → `"s"`. Two characters, but well worth understanding.

---

### Is the root cause correctly identified?

Yes. The key difference between `-Oz` and `-Os`:

- `-Oz` enables more aggressive loop outlining, disables vectorisation, and makes different inlining trade-offs vs `-Os` — specifically, it prefers code-size over almost everything else, including predictable codegen for target-feature-decorated functions.
- `leanMultisig`'s `rec_aggregation / backend` crate uses `#[cfg(target_feature)]` for AVX2/AVX-512 SIMD dispatch. Under `-Oz`, LLVM's outliner can split a target-feature-attributed function body across outlined stubs that do NOT carry the function's `#[target_feature(enable = "avx512f")]` annotation. The stub then executes an AVX-512 instruction without the CPU context having been set up for it (or without a `vzeroupper` transition), producing a GPF on certain Linux kernel + microcode combinations.
- `-Os` still optimises for size but uses conservative outlining — this class of split doesn't occur.

The symptom (deterministic crash in `xmss_aggregate` on x86_64 Linux, `"z"` crashes, `"s"` and numeric levels are clean) is textbook for this LLVM outliner + target-feature interaction.

### Scope check — are there other `opt-level = "z"` sites?

Checked. The main `[profile.release]` does **not** use `"z"` — it uses the Rust default (`opt-level = 3`). Only `openvm-release` and `risc0-release` had `"z"`, and both are updated consistently. ✅

### Does PR #756 (`-Ctarget-cpu=x86-64-v3` default) make this redundant?

No — these are orthogonal. `target-cpu=x86-64-v3` prevents the CPU from seeing AVX-512 feature flags in `RUSTFLAGS`, but the outliner issue happens at the *LLVM IR* level during codegen of the prover binary, before any runtime feature detection. The crash occurs because the optimizer emits an AVX-512 instruction in a code path that shouldn't have it, regardless of what `target-cpu` says about the host. Both fixes are needed.

### Binary size impact

`"z"` vs `"s"` typically saves 5–15% depending on the codebase. For prover binaries that are already stripped and use `panic = "abort"` / `codegen-units = 1`, the delta will be in the low single-digit percent range — acceptable given the stability gain.

### Is this a workaround or a fix?

A workaround — the underlying LLVM outliner + target-feature bug remains. The PR correctly acknowledges this by referencing tracking issue #734 in both the comment and title. If LLVM or Rust upstream fixes the outliner regression, reverting to `"z"` would be safe. ✅

### Title vs diff

Title says "multisig-glue" but the diff is only `Cargo.toml`. Not wrong — `xmss_aggregate` lives in `multisig-glue` and is the crash site — but the change itself is purely a profile flag. Worth noting, not a blocker.

---

Well-reasoned workaround with correct analysis and thorough documentation. Approved. 🚢

gballet (approved):
That is consistent with my analysis of the bug. Ideally, we'd get rid of as much rust as possible, so that we don't face this kind of problems.

Issue #636, opened 2026-03-06
Update store


anshalshukla:
Implement changes in https://github.com/leanEthereum/leanSpec/pull/436

PR #764, opened 2026-04-18
feat: add runtime aggregator role toggle API

ch4r10t33r:
## Summary

Implements the runtime aggregator role toggle from [leanSpec PR #636](https://github.com/leanEthereum/leanSpec/pull/636).

- `BeamChain.is_aggregator_enabled` is now `std.atomic.Value(bool)` with `isAggregator()` / `setAggregator()` helpers; gossip import and tick-driven aggregation paths pick up the new value on their next read.
- New `/lean/v0/admin/aggregator` endpoint on the API server:
  - `GET` returns `{"is_aggregator": <bool>}`
  - `POST` accepts `{"enabled": <bool>}`, flips the flag, updates the `lean_is_aggregator` metric, returns `{"is_aggregator": <new>, "previous": <old>}`
  - Bounded body (`ADMIN_BODY_MAX_BYTES = 1024`), strict `Content-Length` handling, `503` when chain isn't yet wired, `405` with `Allow: GET, POST` for other methods, `400` on missing/invalid JSON or missing/non-bool `enabled`, `413` if body too large
- `--is-aggregator` CLI flag is now a startup seed; help text clarifies the role can be flipped at runtime via the new endpoint.
- Network subnet subscriptions remain fixed at startup (hot-standby model per the spec PR) — no ENR or gossip subscription changes at runtime.

## Test plan

- [x] `zig fmt --check .`
- [x] `zig build simtest --summary all` — all 15 tests pass, including new `admin aggregator endpoint - GET returns seed, POST toggles at runtime` covering: startup polling through the 503 window, GET seed, POST off/on toggles with `previous` field, idempotent repeats, and `400` paths for missing `enabled` and non-boolean `enabled`.


zclawz (dismissed):
## Review — LGTM ✅

### Atomic design — correct

`is_aggregator_enabled: std.atomic.Value(bool)` with `.load(.acquire)` for all readers and `.swap(.acq_rel)` in `setAggregator` is the right memory-ordering choice. The acquire/release pair ensures the new value is visible to all threads (gossip handler, tick loop) after `setAggregator` returns. All 4 pre-PR direct field reads are correctly migrated: 2 in `chain.zig` use `.load(.acquire)`, 3 in `node.zig` use the new `isAggregator()` method. ✅

### `handleAggregatorPost` — safe request handling

- Requires `content-length`, rejects 0-length and oversized bodies ✅
- Body is heap-allocated to `content_length` bytes (not `ADMIN_BODY_MAX_BYTES`) — avoids over-reading ✅
- Read loop correctly handles partial reads via `readSliceShort`, then checks `read_total != len` for short-body 400 ✅
- `std.json.parseFromSlice` used correctly with `defer parsed.deinit()` ✅
- Stack-allocated response buffers are sized generously (`[128]u8` for a ≤40-byte response body) ✅

### Things to be aware of (neither are blocking)

**Metrics update is not atomic with `setAggregator`:**
```zig
const previous = chain.setAggregator(enabled);  // atomic
zeam_metrics.metrics.lean_is_aggregator.set(...)  // separate op
```
There's a tiny window where the runtime flag and the metric disagree. Fine for Prometheus — scrapes are not transactional — but worth documenting.

**No authentication on `/lean/v0/admin/aggregator`:**
The endpoint is on the same port as the regular API and requires no credentials. If the API port is publicly reachable, anyone can toggle the aggregator role. For a testnet/devnet this is acceptable; worth an auth layer before production. The hot-standby model comment in `node.zig` is good documentation. ✅

### Integration test — thorough

Covers: GET initial state, POST off + verify response + verify GET, POST on + verify idempotent POST, 400 for missing field, 400 for wrong type. The poll-for-chain-ready loop with a 60s deadline is the right pattern for this test environment. ✅

### `boolToJson` helper

`inline fn boolToJson` — small, correct, saves repetition. ✅

Approved. 🚢

anshalshukla:
I find this a bit confusing so we need to start the validator as aggregator if we want to toggle the switch later and start aggregating? Why not subscribe to the subnets when the API request is received? Or have is_aggregator as an enum value with aggregating, staged, none. So if the node is started as aggregation/staged it can subscribed to the subnets. 

Current steps to setup would look very counterintuitive and will be prone to errors

pkgs/cli/test/integration.zig
```diff
@@ -199,25 +209,47 @@ const ZeamRequest = struct {
...
+        var connection = try net.tcpConnectToAddress(address);
+        defer connection.close();
+
+        var request_buffer: [4096]u8 = undefined;
+        const request = try std.fmt.bufPrint(&request_buffer, "POST {s} HTTP/1.1\r\n" ++
+            "Host: {s}:{d}\r\n" ++
+            "Content-Type: application/json\r\n" ++
+            "Content-Length: {d}\r\n" ++
+            "Connection: close\r\n" ++
+            "\r\n" ++
```
anshalshukla: can't we use http client here? I think we should avoid manual formation of headers

ch4r10t33r: Switched `getAggregator`/`postAggregator` to `std.http.Client.fetch` in 1caab37. Helpers now return a typed `HttpResponse { status: std.http.Status, body: []u8 }` and the tests `expectEqual` on `std.http.Status.ok` / `.bad_request` instead of grepping `HTTP/1.1 200`. `getMetrics` / `getHealth` still use the raw TCP writer — they predate this PR and felt out of scope.

Simtest (incl. the admin aggregator endpoint test) still passes: 15/15.

ch4r10t33r:
> subnets when the API request is received? Or have is_aggregator as an enum value with aggregating, staged, none. So if the node is started as aggregation/staged it can subscribed to the subnets.
> 
> Current steps to setup would look very counterintuitive and will be prone to errors

This feature is intended for debugging purposes only. In a subnet scenario, if the sole aggregator crashes or stops aggregating, another validator can be manually switched to act as the aggregator. This helps preserve network finality. It is an interim solution until in-protocol redundancy is implemented.

zclawz (approved):
Re-approved ✅

The only change vs the previous version is the integration test rework. Using `std.http.Client` via `fetchAdmin` / `HttpResponse` is a clear improvement over the hand-rolled TCP approach:

- Status assertions use `expectEqual(std.http.Status.ok, ...)` instead of substring-matching `"HTTP/1.1 200"` in raw wire bytes ✅
- Poll loop checks `get_before.status != .ok` — typed, unambiguous ✅
- `HttpResponse.deinit()` with `defer` — clean memory management with no leaks on error paths ✅
- `std.http.Client` handles Content-Length and request framing, removing boilerplate and the risk of hand-rolling headers incorrectly ✅

All prior review points still apply. Approved. 🚢

PR #765, opened 2026-04-18
fix: segfault during cached-block catch-up sync

ch4r10t33r:
## Summary

Fixes two distinct segfaults observed during catch-up sync on x86_64:

1. **`ssz.serialize` corrupting live cached `SignedBlock`s.** Stack bottoms out in
   `onBlock` iterating `aggregated_attestations` (or later in `verifySignatures`)
   with a dangling slice into a cached block's heap buffer. Root cause:
   `ssz.serialize` on a live cached `SignedBlock` (directly, or via `sszClone`)
   mutates nested `List` / `Bitlist` fields (`aggregation_bits`, `proof_data`),
   so the next read of the same cached block crashes.

2. **`fetched_blocks` hashmap rehash race** (folded in from #741). The map in the
   network layer is shared between the libxev main thread (`onInterval`) and the
   libp2p worker thread (`onGossip` / `onReqRespResponse` / `onReqRespRequest`).
   When the worker inserts and triggers a rehash, the main thread's in-flight
   lookups/iterations become invalid and panic.

### ssz-corruption fix (cherry-picked from `fix/libp2p-glue-c-abi-x86_64`)

- `1aecb39 chain, database: persist pre-forkchoice signed block SSZ for rocksdb`
- `1fd70ea node, types: capture block SSZ bytes at cache time, never re-serialize cached blocks`

Highlights:

- `types/utils`: new `sszCloneAndGetBytes` — one serialize pass returning both a
  deep clone and the SSZ bytes (caller owns), avoiding a second serialize on the
  same value.
- `network`: new `fetched_block_ssz` map (root → `[]u8`) with lifecycle tied to
  `cacheFetchedBlock` / `removeFetchedBlock` / `deinit`; `getFetchedBlockSsz` /
  `storeFetchedBlockSsz` helpers.
- `node.cacheBlockAndFetchParent`: uses `sszCloneAndGetBytes` and stores the
  bytes alongside the cached block.
- `chain.CachedProcessedBlockInfo`: new `sszBytes: ?[]const u8` field.
- `node.processCachedDescendants`: retrieves stored bytes and passes them via
  `blockInfo.sszBytes`.
- `chain.onBlock` / `updateBlockDb`: uses `blockInfo.sszBytes` when present,
  falls back to a disposable-clone serialize only for locally produced blocks.

### Hashmap rehash race fix (cherry-picked from #741)

- `71280e3 fix race condition leading to seg faults` (by @anshalshukla)

Adds a `std.Thread.Mutex` to `BeamNode` that serializes work between the
libxev main thread and the libp2p worker thread. Locks are taken around:

- `onGossip`
- `onReqRespResponse`
- `onReqRespRequest` (`.blocks_by_root` branch)
- `onInterval` (chain tick, pending replay, RPC sweep, `processReadyCachedBlocks`)

## Test plan

- [x] `zig fmt --check .`
- [x] `zig build simtest --summary all`
- [ ] Soak under catch-up-sync scenario on x86_64 hardware to confirm both
      original crashes are resolved

### DIFF
--- a/pkgs/database/src/rocksdb.zig
+++ b/pkgs/database/src/rocksdb.zig
@@ -238,6 +238,26 @@ pub fn RocksDB(comptime column_namespaces: []const ColumnNamespace) type {
                 );
             }
 
+            /// Same as `putBlock` but stores already-serialized SSZ (must match `types.SignedBlock` encoding).
+            pub fn putBlockSerialized(
+                self: *WriteBatch,
+                comptime cn: ColumnNamespace,
+                block_root: types.Root,
+                serialized_block: []const u8,
+            ) void {
+                const key = interface.formatBlockKey(self.allocator, &block_root) catch |err| {
+                    self.logger.err("failed to format block key for putBlockSerialized: {any}", .{err});
+                    return;
+                };
+                defer self.allocator.free(key);
+
+                self.put(cn, key, serialized_block);
+                self.logger.debug("added pre-serialized block to batch: root=0x{x} len={d}", .{
+                    &block_root,
+                    serialized_block.len,
+                });
+            }
+
             /// Put a state to this write batch
             pub fn putState(
                 self: *WriteBatch,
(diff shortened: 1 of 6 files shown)


noopur23:
the pr resolved the seg fault issue.

noopur23 (approved):

PR #766, opened 2026-04-19
ci(hive): hoist simulators/lean Cargo.lock to hive root before docker build

### LINKED ISSUE #763: Daily Hive run failed: lean / devnet4
The scheduled Hive CI run failed.

| Field | Value |
|---|---|
| Simulator | `lean` |
| Devnet profile | `devnet4` |
| Hive source | `ethereum/hive@master` |
| Commit tested | [`64f36899b9fb112723ec21ff23378ba5dc6b5524`](https://github.com/blockblaz/zeam/commit/64f36899b9fb112723ec21ff23378ba5dc6b5524) |
| Workflow run | [#24621770168](https://github.com/blockblaz/zeam/actions/runs/24621770168) |

This issue is auto-opened by the `hive` workflow and updated in place by later scheduled failures; close it once the underlying cause is fixed and the next daily run goes green.

### Notes

- The upstream `clients/zeam` Dockerfile in `ethereum/hive` selects the devnet4
  binary from the pre-published `blockblaz/zeam:devnet4` image (no tag build
  arg), so a failure here may reflect the published image rather than HEAD of
  `main`. Check the published image tag before assuming a regression in-tree.
- Logs, per-test results, and the hiveview output are attached to the workflow
  run as artifacts (`hive-zeam-devnet4-*`).

ch4r10t33r:
## Summary

Closes #763.

All three scheduled Hive runs on `main` since the workflow landed (`#745`) have failed with the same upstream build error:

```
Apr 19 05:44:47.060 ERR image build failed image=hive/simulators/lean:latest
  dir=. err=\"COPY failed: file not found in build context or excluded by
  .dockerignore: stat Cargo.lock: file does not exist\"
```

([run 24621770168](https://github.com/blockblaz/zeam/actions/runs/24621770168/job/71993751023))

## Root cause (in upstream ethereum/hive, not zeam)

- [`ethereum/hive@684d4ad` (\"Add Lean RPC compatibility tests\", #1422, 2026-04-14)](https://github.com/ethereum/hive/commit/684d4ad) rewrote `simulators/lean/Dockerfile` as a multi-stage build that does `COPY ./Cargo.toml`, `COPY ./Cargo.lock`, `COPY ./hivesim-rs`, `COPY ./simulators/lean` from the build context root.
- The same commit added `simulators/lean/hive_context.txt` containing `../..`, which [`internal/libdocker/builder.go::BuildSimulatorImage`](https://github.com/ethereum/hive/blob/master/internal/libdocker/builder.go) uses to override the docker build context to the **hive repo root**.
- The hive repo root has `Cargo.toml` committed (workspace spanning `hivesim-rs` and `simulators/lean`), but `Cargo.lock` is **gitignored** at the root (see `/Cargo.lock` in [`ethereum/hive/.gitignore`](https://github.com/ethereum/hive/blob/master/.gitignore)). Nothing in the `./hive` build pipeline generates that file before the docker build runs.
- Result: every `./hive --sim lean` run fails at the first `COPY ./Cargo.lock` because the file literally isn't in the build context.

### Verification the images/inputs exist

Checked via `gh api` against `ethereum/hive` at `master`:

- Root `Cargo.toml`: present (workspace definition).
- Root `Cargo.lock`: **missing** (gitignored by `/Cargo.lock`).
- `simulators/lean/Cargo.toml`: present.
- `simulators/lean/Cargo.lock`: **present and committed** (resolves the same workspace).
- `simulators/lean/hive_context.txt`: `../..` (pins build context to hive root).
- `clients/zeam/Dockerfile`: present.
- `simulators/lean/clients/devnet4.yaml`: present.

So the failure is purely the missing root `Cargo.lock`, not a missing client image or missing devnet profile. Reproduced locally with a minimal Dockerfile against a fresh `ethereum/hive` checkout: `COPY ./Cargo.lock ./Cargo.lock` fails before the fix and succeeds after hoisting `simulators/lean/Cargo.lock` to the root.

## Fix

Replace the single `ethpandaops/hive-github-action@v0.6.3` step with its inlined equivalent (same pinned SHAs for `actions/setup-go`, `docker/setup-docker-action`, `actions/checkout`, `actions/upload-artifact`) so we can insert a workaround step between `Checkout hive` and `Build hive`:

```yaml
- name: Workaround upstream missing Cargo.lock at hive root
  working-directory: ./src
  shell: bash
  run: |
    set -euo pipefail
    if [ -f Cargo.lock ]; then
      echo \"Cargo.lock already present at hive root; upstream fix landed, nothing to do.\"
      exit 0
    fi
    if [ ! -f simulators/lean/Cargo.lock ]; then
      echo \"simulators/lean/Cargo.lock missing; cannot work around missing root Cargo.lock.\" >&2
      exit 1
    fi
    cp simulators/lean/Cargo.lock Cargo.lock
```

Why hoisting is safe: the hive root `Cargo.toml` defines a Cargo workspace whose members are `hivesim-rs` and `simulators/lean`, and `simulators/lean/Cargo.lock` is that workspace's committed lockfile. Copying it to the root satisfies the Dockerfile's `COPY ./Cargo.lock` without changing what `cargo build --locked` will actually resolve. The step is self-deactivating the day upstream lands a real fix.

Why inlining vs. sticking with `hive-github-action`: the action checks out `ethereum/hive` inside itself, and `actions/checkout@v6` defaults to `clean: true` (`git clean -ffdx`) which would wipe any pre-seeded `Cargo.lock` even though the file is gitignored. There is no way to interpose our workaround between its internal checkout and its internal `./hive` invocation without inlining. Pinned SHAs and behavior otherwise match v0.6.3 exactly.

## How PR runs trigger hive

Per `.github/workflows/hive.yml`:

```yaml
on:
  pull_request:
    types: [opened, synchronize, reopened, labeled, unlabeled]
# ...
if: >-
  github.event_name != 'pull_request' ||
  contains(github.event.pull_request.labels.*.name, 'run-hive')
```

So Hive runs on a PR only when the `run-hive` label is applied (or re-applied on synchronize). This PR intentionally requests that label so CI exercises the fix end-to-end; once the first labeled run goes green we can mark it ready for review.

## Follow-ups (not part of this PR)

- Open an upstream PR against `ethereum/hive` that either commits a root `Cargo.lock` (removing the `/Cargo.lock` line from `.gitignore` and checking in the generated file) or changes `simulators/lean/Dockerfile` to `COPY ./simulators/lean/Cargo.lock ./Cargo.lock`. The workaround step here is explicitly idempotent against that landing.
- When the upstream fix lands, drop the workaround step and (optionally) revert to `ethpandaops/hive-github-action`.

## Test plan

- [x] Root cause confirmed by inspecting `ethereum/hive@master` contents via `gh api` (listed above) and by reading the two failed scheduled runs.
- [x] Fix validated locally: a minimal docker build with `COPY ./Cargo.toml` + `COPY ./Cargo.lock` against a fresh `ethereum/hive` clone fails as in CI, and succeeds after `cp simulators/lean/Cargo.lock Cargo.lock`.
- [x] `actionlint .github/workflows/hive.yml` reports no new issues (the pre-existing SC2129 style warning on the `Resolve inputs` step is on `main` already and not touched here).
- [ ] Add the `run-hive` label on this PR so hive CI runs against this branch and we can confirm the upstream build now gets past the COPY step.

### DIFF
--- a/.github/scripts/hive-summary.sh
+++ b/.github/scripts/hive-summary.sh
@@ -1,0 +1,196 @@
+#!/usr/bin/env bash
+#
+# Emit a markdown summary of a hive results directory.
+#
+# Reads every `<unix>-<hex>.json` suite file hive writes under RESULTS_DIR,
+# aggregates pass/fail/timeout counts across suites, and for each failure
+# pulls the relevant slice out of the suite's testDetailsLog (via the
+# {begin,end} byte offsets hive records on each failing test) so the
+# summary can show why the test failed without forcing the reader to
+# download the artifact.
+#
+# Usage: hive-summary.sh <results-dir> [output-file]
+#
+#   <results-dir>   directory hive wrote with `--results-root`
+#   [output-file]   file to append the markdown summary to; defaults to
+#                   stdout. Intended to be pointed at $GITHUB_STEP_SUMMARY
+#                   and/or a PR-comment fragment.
+#
+# Output is markdown. If the directory has no suite files (e.g. hive
+# failed before running any tests) the script still emits a summary
+# block noting that no results were produced.
+
+set -euo pipefail
+
+results_dir="${1:-}"
+output="${2:-/dev/stdout}"
+
+if [ -z "$results_dir" ]; then
+  echo "usage: $0 <results-dir> [output-file]" >&2
+  exit 2
+fi
+
+if [ ! -d "$results_dir" ]; then
+  echo "results directory not found: $results_dir" >&2
+  exit 1
+fi
+
+# Emit markdown to a temp buffer first so a partial failure doesn't
+# truncate an already-open $GITHUB_STEP_SUMMARY.
+tmp="$(mktemp)"
+trap 'rm -f "$tmp"' EXIT
+
+shopt -s nullglob
+# Suite files are named `<unix>-<hex>.json` (see libhive/testmanager.go:
+# writeSuiteFile). `hive.json` is the top-level run metadata and must be
+# excluded.
+suites=()
+for f in "$results_dir"/*.json; do
+  base="$(basename "$f")"
+  [ "$base" = "hive.json" ] && continue
+  suites+=("$f")
+done
+
+if [ "${#suites[@]}" -eq 0 ]; then
+  {
+    echo "### Hive test results"
+    echo
+    echo "No suite result files were produced in \`$results_dir\`. This usually means hive failed before any test suite finished (e.g. during client or simulator image build)."
+  } >> "$tmp"
+  cat "$tmp" >> "$output"
+  exit 0
+fi
+
+# Aggregate counters and a compact per-failure record in a single jq pass
+# per suite. Timeouts are counted both as failures and as a separate
+# bucket, matching how hiveview displays them.
+total=0
+passed=0
+failed=0
+timeouts=0
+# Failures list: one record per line, tab-separated fields:
+#   <suite>\t<test>\t<timeout?>\t<details_log>\t<begin>\t<end>\t<inline_details>
+failures_tsv=""
+
+for suite in "${suites[@]}"; do
+  suite_name="$(jq -r '.name // "(unnamed)"' "$suite")"
+  details_log_rel="$(jq -r '.testDetailsLog // ""' "$suite")"
+  details_log_abs=""
+  if [ -n "$details_log_rel" ] && [ -f "$results_dir/$details_log_rel" ]; then
+    details_log_abs="$results_dir/$details_log_rel"
+  fi
+
+  counts="$(jq -r '
+    [.testCases[].summaryResult] as $r
+    | [($r|length),
+       ([$r[] | select(.pass==true)] | length),
+       ([$r[] | select(.pass==false)] | length),
+       ([$r[] | select(.timeout==true)] | length)]
+    | @tsv
+  ' "$suite")"
+  IFS=$'\t' read -r t p f to <<<"$counts"
+  total=$((total + t))
+  passed=$((passed + p))
+  failed=$((failed + f))
+  timeouts=$((timeouts + to))
+
+  suite_failures="$(jq -r --arg suite "$suite_name" --arg log "$details_log_abs" '
+    .testCases
+    | to_entries
+    | map(select(.value.summaryResult.pass == false))
+    | .[]
+    | [$suite,
+       .value.name,
+       (.value.summaryResult.timeout // false | tostring),
+       $log,
+       (.value.summaryResult.log.begin // ""),
+       (.value.summaryResult.log.end   // ""),
+       (.value.summaryResult.details   // "")]
+    | @tsv
+  ' "$suite")"
+  if [ -n "$suite_failures" ]; then
+    failures_tsv+="${suite_failures}"$'\n'
+  fi
+done
+
+# Devnet label is carried through by the caller via HIVE_DEVNET_LABEL; we
+# intentionally don't re-parse it from the results (hive's JSON doesn't
+# record the client-file path).
+devnet_label="${HIVE_DEVNET_LABEL:-unknown}"
+simulator_label="${HIVE_SIMULATOR_LABEL:-unknown}"
+
+{
+  echo "### Hive test results"
+  echo
+  echo "| Simulator | Devnet | Suites | Tests | Passed | Failed | Timeouts |"
+  echo "|---|---|---|---|---|---|---|"
+  echo "| \`$simulator_label\` | \`$devnet_label\` | ${#suites[@]} | $total | $passed | $failed | $timeouts |"
+  echo
+} >> "$tmp"
+
+if [ "$failed" -eq 0 ]; then
+  echo "All $total tests passed." >> "$tmp"
+  cat "$tmp" >> "$output"
+  exit 0
+fi
+
+# Snippet budget per failure. GitHub step summary caps at 1 MiB total
+# and PR comments at 65536 chars, so keep each excerpt bounded and
+# collapse via <details> so the comment stays scannable.
+snippet_bytes=1024
+
+{
+  echo "<details open>"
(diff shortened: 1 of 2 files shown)


zclawz (dismissed):
## Review — LGTM with notes ✅

Sizable PR but coherent in purpose: replaces the black-box `hive-github-action` with inlined steps to (a) work around the upstream missing `Cargo.lock` bug and (b) emit rich pass/fail summaries directly in the workflow.

---

### Upstream bug workaround — correct

The root cause analysis in the comment is accurate:
- `simulators/lean/Dockerfile` uses `hive_context.txt = ../../` so docker build sees the hive repo root as its context, but `Cargo.lock` is gitignored there → `COPY failed: stat Cargo.lock: file does not exist`.
- The committed `simulators/lean/Cargo.lock` predates the workspace restructure, so naively hoisting it would fail `cargo build --locked`.

The fix — `cargo generate-lockfile --manifest-path Cargo.toml` — generates a workspace-consistent lockfile that satisfies `--locked`. The self-healing guard (`if [ -f Cargo.lock ]`) removes the workaround automatically once upstream commits a root `Cargo.lock`. ✅

---

### Action SHA verification

| Action | Comment tag | SHA in PR | Verified |
|--------|-------------|-----------|----------|
| `actions/setup-go` | v6.1.0 | `4dc6199c7b1a...` | not checked (flag for author) |
| `docker/setup-docker-action` | v4.6.0 | `e61617a16c40...` | not checked |
| `actions/checkout` | v6.0.0 | `1af3b93b6815...` | ✅ confirmed via GitHub API |
| `actions/upload-artifact` | v5.0.0 | `330a01c490ac...` | not checked |

`actions/checkout@v6.0.0` SHA is verified. Authors should also verify the other three SHAs match their tagged releases.

---

### 🔴 Fragile success detection

```bash
(./hive ... 2>&1 || true) | tee hive.log
if tail -n 10 hive.log | grep -q "simulation .* finished"; then
```

Two problems:
1. `|| true` swallows hive's exit code entirely. `tee` exits 0 regardless. So even if hive crashes with a fatal error, the `tail | grep` is the only gate.
2. `tail -n 10` is brittle — if hive emits lines after the completion message (e.g. a final summary or newline), `grep` will still find it. If the message moves earlier in the log, `tail -n 10` might miss it.

More robust alternative using `PIPESTATUS`:
```bash
./hive ... --docker.output 2>&1 | tee hive.log
hive_exit=${PIPESTATUS[0]}
if [ "$hive_exit" -eq 0 ]; then exit 0; else exit 1; fi
```
This trusts hive's own exit code, which is the canonical signal of success/failure. Not a blocker since the workaround has been observed to work, but worth a follow-up.

---

### 🟡 Dead variable: `RUN_SUCCESS`

`echo "RUN_SUCCESS=..." >> "$GITHUB_ENV"` is set but never referenced by any downstream step — the job outcome is determined by the step's exit code alone. Can be removed.

---

### 🟡 `hive-summary.sh`: jq `@tsv` + embedded newlines in `details`

`jq @tsv` escapes embedded tabs as `\t` but does **not** escape newlines. If a hive test's `summaryResult.details` field contains a newline, the TSV line will split and the `while IFS=$'\t' read -r suite name ...` loop will misparse it, producing a garbled entry. For the current lean simulator's test output this is probably fine, but it's worth documenting as a known limitation or filtering with `@json` → inline base64 for safety.

---

### 🟡 `dd bs=1` byte-seeking performance

`dd if="$log_path" bs=1 skip="$begin" count="$count"` is byte-exact but very slow for large `$begin` offsets (one syscall per byte). If hive log files are large, this could add noticeable latency in the summary step. `tail -c +$((begin+1))` then `head -c $count` would be faster and equally correct on GNU coreutils. Not a blocker for CI.

---

### Things that are correct

- Temp-file pattern with `trap rm` prevents partial writes to `$GITHUB_STEP_SUMMARY` on failure ✅
- `shopt -s nullglob` + explicit exclusion of `hive.json` correctly identifies suite files ✅  
- `snippet_bytes=1024` cap on inline excerpts prevents blowing the PR comment or step summary size limits ✅
- `--docker.output` flag enables streaming build logs — critical for diagnosing `cargo build` failures that previously only showed a one-liner ✅
- The Cargo.lock workaround step exits loudly (`exit 1`) if `Cargo.toml` is missing, rather than silently handing docker a bad context ✅
- PR comment markdown now includes pass/fail table + failing test excerpts — big improvement over the previous generic summary ✅

---

Approved. The two 🔴/🟡 items are worth follow-ups but don't block the fix landing. 🚢

zclawz (dismissed):
All four findings from the previous review have been addressed:

- **🔴 Fragile success detection** → `${PIPESTATUS[0]}` now captures hive's exit code directly; grep-on-tail is gone. Comment explains the intent clearly. ✅
- **🟡 Dead variable** → `RUN_SUCCESS` removed. ✅
- **🟡 TSV + newlines in `details`** → `details` is no longer carried through `@tsv`; instead `suite_file` and `test_key` are passed through the TSV (safe: single-line identifiers), and `details` is re-read on demand with a targeted `jq --arg k "$test_key" .testCases[$k].summaryResult.details` against the source JSON. Embedded newlines in the details field now survive intact. ✅
- **🟡 `dd bs=1` performance** → replaced with `tail -c "+$((begin+1))" | head -c "$count"` — buffered I/O, same semantics, much cheaper for large offsets. ✅

Approved. 🚢

zclawz (approved):
Re-approved ✅

The change adds an optional `[counts-file]` third argument and `emit_counts()` function to `hive-summary.sh` — clean, backward-compatible addition.

**`emit_counts()` design:**
- No-op when `counts_file` is unset (existing callers unaffected) ✅
- Always emits all 5 keys (`suites`, `total`, `passed`, `failed`, `timeouts`) to prevent partial output when `cat`'d into `$GITHUB_OUTPUT` ✅
- Called at no-suites early exit (zeros), all-pass early exit (real counts), and (by pattern) at end of failure path ✅

All previously approved items remain intact. Approved. 🚢

Issue #43, opened 2025-08-14
track and dealloc the structures in network/src/mock.zig


noopur23:
file: pkgs/network/src/mock.zig
pub fn onGossip
TODO: track and dealloc the structures

_Originally posted by @gballet in https://github.com/blockblaz/zeam/pull/36#discussion_r2250142093_

markolazic01:
Hello there, I'd like to try solving this.

I noticed that this fix can be directly added into the flow of a gossip handler inside the `interface.zig`, is it alright to manage it there or should it be strictly applied to the `mock.zig`?

PR #770, opened 2026-04-21
fix: enhancement in dealloc of in-flight gossip structures

markolazic01:
## Track and dealloc in-flight gossip structures

Enhancement, related to #43

### Problem

`GenericGossipHandler.onGossip` allocates a `MessagePublishWrapper` and an `xev.Completion` per handler when scheduling on the event loop. These are freed inside the timer callback after dispatch, but if `deinit()` is called before the timers fire, the allocations leak. This edge-case is not very likely to happen, but this change might be useful to have.

### Solution

- Added an `inflightWrappers` tracking set (`AutoHashMapUnmanaged`) to `GenericGossipHandler`
- Added `completion` and `gossipHandler` fields to `MessagePublishWrapper` so it can self-untrack
- In `onGossip`: register each wrapper after creation
- In the timer callback: remove the wrapper from tracking before freeing
- In `deinit`: free any remaining in-flight wrappers and their completions

Also reordered the allocation in `onGossip` so that the completion is created before the wrapper (since the wrapper now takes it as a parameter), with an `errdefer` to clean up on failure.

I am interested to hear if this helps with #43 and, if not, what is actually needed.


pkgs/network/src/interface.zig
```diff
@@ -894,14 +907,18 @@ pub const GenericGossipHandler = struct {
...
-
                 // Create a separate completion object for each handler to avoid conflicts
                 const completion = try self.allocator.create(xev.Completion);
                 completion.* = undefined;
 
+                errdefer self.allocator.destroy(completion);
+
+                const publishWrapper = try MessagePublishWrapper.init(self.allocator, handler, data, sender_peer_id, self.networkId, self.logger, completion, self);
+
+                try self.inflightWrappers.put(self.allocator, publishWrapper, {});
```
anshalshukla, line 951: If `inflightWrappers.put` fails (OOM), `publishWrapper` leaks — there is no `errdefer` between `MessagePublishWrapper.init` on line 916 and this `put`. The completion is cleaned up by the prior `errdefer`, but the wrapper allocation (including `cloned_data` and `sender_peer_id_copy` allocated inside `init`) is orphaned. Suggest adding `errdefer publishWrapper.deinit();` after line 916.

Reason: bug — introduces a new leak path in a PR whose purpose is eliminating leaks.

anshalshukla:
@markolazic01 thanks for the PR. I'm considering a few changes in the design and remove xev from the p2p and run it in different thread without scheduling option. Give me a few days and I'll come back with an update on this PR. The `errdefer` comment above is good to have but let me first decide upon if mock can be better refactored without the scheduling option.

pkgs/network/src/interface.zig
markolazic01: Great catch, thank you! Fixed in [3f38e6c](https://github.com/blockblaz/zeam/pull/770/commits/3f38e6cc6112bf84f6c51901f1957a11d0097d2c)  (the latest commit).

markolazic01:
Thanks @anshalshukla for taking time to review this PR. I added the requested line now but I will remove it later if you find it unnecessary. Take your time.

markolazic01:
Hey @anshalshukla, just a kind reminder on this one.

Issue #749, opened 2026-04-17
node: fetched-block cache fills during finality stall and never drains — node permanently stuck after ~1024 slots without finalization


ch4r10t33r:
## Summary

When finality stalls on the network for a prolonged period (observed: ~8 hours on devnet-4), a zeam node's fetched-block cache silently fills to its 1024-entry cap and then **permanently** rejects all new blocks, including the ones it would need to catch up. There is no self-recovery: the only way out is restarting with `--checkpoint-sync-url`.

Because the cache is **pruned only on finalization events**, any stall longer than `MAX_CACHED_BLOCKS` slots (~68 min at 4s/slot) is sufficient to wedge the node. If that node is the configured aggregator, the wedge removes aggregations from the network and can amplify the original stall.

## Observed behavior (devnet-4, 2026-04-17)

- All 8 clients across the network reported `finalized=3797, justified=3802` and stopped advancing. (Initial cause of the stall is not the subject of this issue — see caveat below.)
- Non-zeam clients continued to produce blocks; network head advanced to ~slot 12085.
- zeam's head was frozen at slot 3806. Its logs filled with:
  ```
  [warning] Cache full (1024 blocks), rejecting future block 0x.. at slot 12078
  [warning] Cache full (1024 blocks), rejecting block 0x.. at slot 12080
  [error]   error.UnknownSourceBlock producing aggregation for slot=3806
  ```
- curl `/lean/v0/checkpoints/justified` on zeam returned `slot=3800` while peers returned `slot=3802`.
- Recovery required \`spin-node.sh --restart-client zeam_0 --checkpoint-sync-url https://leanpoint.leanroadmap.org/lean/v0/states/finalized\`.

## Root cause analysis

Cache admission in \`pkgs/node/src/node.zig\`:

\`\`\`zig
// If cache is full, reject - proactive pruning on finalization keeps the cache bounded
if (self.network.fetched_blocks.count() >= constants.MAX_CACHED_BLOCKS) {
    self.logger.warn(\"Cache full ({d} blocks), rejecting block ...\", ...);
    return CacheBlockError.CachingFailed;
}
\`\`\`
(also mirrored in \`cacheFutureBlock\` at ~line 593)

Pruning is wired via \`prune_cached_blocks_fn\` and invoked only when finalization advances (\`chain.zig:1102\`). \`constants.zig\` defines:

\`\`\`zig
pub const MAX_CACHED_BLOCKS = 1024;
\`\`\`

[...]

PR #772, opened 2026-04-22
fix: treat block_cache miss as stale in justifications cleanup

### LINKED ISSUE #771: types/state: STF aborts with InvalidJustificationRoot when a justifications root was just pruned from block_cache, wedging the node permanently
## Summary

When a block causes the STF to advance finalization, the post-finalization prune loop in `processAttestations` looks up each pending-justifications root in the shared `block_cache` to learn its slot. If any root is not in `block_cache`, the STF returns `StateTransitionError.InvalidJustificationRoot` and the whole block is rejected. This is not a transient error: subsequent slot ticks retry the same state transition on the same state and hit the same missing lookup forever, so the node is wedged and cannot recover without a checkpoint-sync restart.

Every other client in the network (grandine, gean, ethlambda, lantern, qlean) imported the same block without issue.

## Observed behavior (devnet-4, 2026-04-22)

- At `Apr-22 06:56:29.535` zeam_0 applied a jumbo finalization jump as part of processing block 268:
  ```
  [chain] finalization canonicality analysis
          (previousFinalized slot=171 to latestFinalized slot=252):
          newly finalized=55, orphaned/missing=26, non finalized descendants=12 & finalized non canonical=4
  ```
  This triggered the cached-block prune callback and block_cache was pruned up to slot 252.

- At `Apr-22 06:56:32.155`, block 269 (`0xb4bd9e5f…`, proposer 5 / lantern_0, parent `0x92a7e754…` slot 268) was rejected:
  ```
  [chain]   error processing block for slot=269 root=0xb4bd9e5f...: error.InvalidJustificationRoot
  [network] network-0:: onGossip handler error=error.InvalidJustificationRoot
  ```

- zeam retried the block via RPC twice, same result:
  ```
  [warning] failed to import block fetched via RPC 0xb4bd9e5f... from peer qlean_0: error.InvalidJustificationRoot
  ```

- From slot 272 interval 0 onward, the slot-tick path itself hit the same error on every tick:
  ```
  [s=272 i=0] [error] error ticking validator to time(intervals)=1360 err=error.InvalidJustificationRoot
  [s=272 i=0] [error] error ticking validator to time(intervals)=1360 err=error.InvalidJustificationRoot
  ... (repeats indefinitely)
  ```

- zeam_0's head stayed frozen at slot 268, attestation and aggregation production stopped. All peers were healthy and advanced head to ~slot 770; because zeam_0 was the only `isAggregator:true` node, aggregations ceased network-wide and finality froze at slot 252 across all 8 clients until we moved the aggregator role off zeam via `POST /lean/v0/admin/aggregator`.

## Root cause

`pkgs/types/src/state.zig:505-528` (inside `processAttestations`, after `can_target_finalize` flips to true and `delta > 0`):

```zig
if (delta > 0) {
    try self.shiftJustifiedSlots(delta, allocator);

    var roots_to_remove: std.ArrayList(Root) = .empty;
    defer roots_to_remove.deinit(allocator);
    var iter = justifications.iterator();
    while (iter.next()) |entry| {
        const root = entry.key_ptr.*;
        const slot = block_cache.get(root) orelse return StateTransitionError.InvalidJustificationRoot;
        if (slot <= finalized_slot) {
            try roots_to_remove.append(allocator, root);
        }
    }
    for (roots_to_remove.items) |root| {
        if (justifications.fetchRemove(root)) |kv| {
            allocator.free(kv.value);
        }
    }
}
```

`block_cache` is the shared `utils.RootToSlotCache` owned by `BeamChain`. Its prune policy (invoked from `pkgs/node/src/chain.zig:1132-1138` on every finalization advance, implemented in `pkgs/node/src/node.zig:pruneCachedBlocksCallback`) is: **remove every cached block at slot ≤ new `latest_finalized.slot`**.

The loop above is logically a cleanup of `state.justifications` entries that are now pre-finalized. But because the cleanup is gated on `block_cache.get(root)`, and `block_cache` has already been pruned past the finalization boundary on a previous transition, any pre-finalized root still present in `state.justifications` will miss in the cache — which is precisely the case this loop was written to handle.

So the invariant the code assumes — *"every root in `state.justifications` is always in `block_cache`"* — is false in exactly the scenario the code is trying to clean up. A miss should mean "already stale, drop it", not "hard fail".

### Why this also wedges slot ticks, not just gossip

The error surface is not limited to the inbound gossip path. Slot ticking (`time(intervals)=1360`) also runs the STF on the current state; since the state continues to have the same `justifications` map and the cache keeps the same pruned view, the next tick fails identically. The validator never advances past `s=272 i=0` and never produces another attestation or aggregation. A restart with `--checkpoint-sync-url` is currently the only recovery.

## Suggested directions (non-prescriptive)

Any of the following would have avoided this stall. They are not mutually exclusive.

1. **Treat cache miss as stale.** In `state.zig:519`, a missing root should be dropped from `state.justifications` the same way an ≤ `finalized_slot` root is. Rationale: `block_cache`'s only prune rule drops exactly those blocks. If the slot info is gone, the root is either pre-finalization (safe to drop) or was never seen (shouldn't be in the map to begin with — invariant to assert separately).
2. **Don't lose the slot info.** Store `(root, slot)` pairs in `state.justifications_roots` / the `justifications` working map directly. The STF would then be self-contained and independent of `block_cache` pruning races. This matches the spec's intent — a justification is a checkpoint `(root, slot)`, not just a root.
3. **Preserve `block_cache` entries referenced by `state.justifications_roots`.** Teach the prune callback to retain any root currently in `state.justifications_roots` regardless of slot, dropping it only after the STF removes it from `justifications`. Narrower fix but keeps the two pieces coupled rather than decoupling them.
4. **Don't wedge the slot tick on STF errors.** Orthogonal to the STF bug: a hard STF error on a block should reject the block, not freeze forkchoice/time. The tick should continue so the node can keep attesting on its existing head and re-evaluate when new blocks/attestations arrive.

Fix (1) alone would have recovered this incident without any operator action. Fix (4) is worth doing regardless because it's the second time an STF error has wedged the entire tick loop on this node (previous incident: `error.UnknownSourceBlock` in aggregator production path).

## Reproduction sketch

Any scenario where two finalization-advancing blocks arrive close together and the gap between them is large enough that the first advance prunes `block_cache` past roots still referenced by `state.justifications`. On devnet-4 the trigger was the 171→252 jump at slot 268 followed by block 269 advancing finalization again.

Unit-test repro: construct a pre-state whose `justifications_roots` include a root at slot S, feed a block whose attestations finalize past S, then feed a second block whose attestations finalize further. With the current `block_cache` prune policy the second block's STF will fail.

## References

- `pkgs/types/src/state.zig` L505-528 (prune loop with `block_cache.get(root) orelse return InvalidJustificationRoot`)
- `pkgs/types/src/state.zig` L381-388 (cache plumbing: `owned_cache` vs. chain-owned `block_cache`)
- `pkgs/node/src/node.zig` `pruneCachedBlocksCallback` (finalization-gated prune: removes slot ≤ finalized)
- `pkgs/node/src/chain.zig` L1132-1138 (prune invocation on finalization advance)
- Related (different symptom, same "STF error wedges slot tick" pattern): #749

ch4r10t33r:
Fixes #771.

The post-finalization prune loop in `processAttestations` looked each pending-justification root up in `block_cache` and hard-errored with `StateTransitionError.InvalidJustificationRoot` on a miss. Once the **chain-owned** cache was pruned past any root still sitting in a cached per-state `justifications_roots` map, the miss became permanent: the same pre-state re-failed on every block retry and every slot-tick-driven STF run — wedging the node until checkpoint-sync restart.

## Why the miss is actually reachable

See [#771 follow-up comment](https://github.com/blockblaz/zeam/issues/771#issuecomment-4296405621) for the full timeline. Short version: the global `root_to_slot_cache` is pruned against **chain-wide** `latestFinalized.slot`, but cached `BeamState.justifications_roots` are pinned to the **per-state** `latest_finalized.slot`, which is frozen at import time. When a cross-fork reorg advances chain-wide finality past a reachable cached state's own finalized slot, that state's `justifications_roots` can reference roots the global cache has just evicted.

On devnet-4 this happened at `06:56:29.535`: importing a late-arriving competing block at slot 267 advanced `latestFinalized` 171→252 and triggered `prune(252)`. Fork-choice then swung back to `92a7(268)` (whose post-state still had `latest_finalized=171`), and the next block on top of it (`b4bd9e5f(269)`) hit this loop with a root in `(171, 252]` that was no longer in the cache.

## Fix (this PR — narrow hotfix)

`RootToSlotCache.prune(F)` removes exactly the entries with `slot <= F`. So a miss either means the root was once in the cache and has since been pruned (⇒ pre-finalization ⇒ same outcome as the explicit `<=` branch — drop it) or it was never populated for a root that nonetheless lives in `state.historical_block_hashes` (no slot info to finalize against — drop is still the only safe action). In both cases: treat the miss identically to the `<=` branch, drop the root, continue.

This is option (1) from the issue — minimal, stays inside `state.zig`, no changes to state shape, SSZ types, or the chain-layer prune policy. Sufficient to unwedge this class of stall.

## Out of scope (follow-ups)

Captured in the #771 follow-up comment, not addressed here:

- **Structural coherence fix (preferred follow-up).** Evict cached states whose `latest_finalized.slot < chain latestFinalized.slot` when `onFinalization` fires. Makes the missed-lookup path unreachable in practice and tightens memory. Cleaner than widening the prune floor.
- **Slot-tick resilience.** A hard STF error on block import should reject the block, not freeze forkchoice/time (option 4 in the issue, same "STF error wedges tick" pattern as #749).
- **Observability.** The pre-fix error site didn't log the offending root or either finalized_slot. This PR adds a `logger.debug` in the drop path; an `err`-level log with `root`/`state_finalized`/`new_finalized` on any future hard-failure branches would make repeats trivial to diagnose.

## Scope

Deliberately narrow:

- no changes to `state.justifications_roots` / `justifications_validators` shape
- no changes to `pruneCachedBlocksCallback` or `RootToSlotCache`
- no changes to the forkchoice/tick loop

## Test plan

- [x] `zig fmt --check .`
- [x] `zig build test --summary all` — 19 tests pass in `pkgs/types/src/lib.zig` (was 18, +1 regression)
- [x] `zig build simtest --summary all`
- [x] New regression test `process_attestations drops pending justifications missing from block_cache` fails against the unpatched cleanup loop with `InvalidJustificationRoot` (verified by locally reverting just the src change), passes with the fix.

### DIFF
--- a/pkgs/types/src/state.zig
+++ b/pkgs/types/src/state.zig
@@ -516,8 +516,22 @@ pub const BeamState = struct {
                         var iter = justifications.iterator();
                         while (iter.next()) |entry| {
                             const root = entry.key_ptr.*;
-                            const slot = block_cache.get(root) orelse return StateTransitionError.InvalidJustificationRoot;
-                            if (slot <= finalized_slot) {
+                            // A miss in `block_cache` for a root that's still in
+                            // `state.justifications` is not an error — it's the
+                            // expected outcome once finalization has advanced past
+                            // the root's slot. `RootToSlotCache.prune` removes
+                            // entries with `slot <= (some prior finalized_slot)`
+                            // and finalization is monotonic, so a miss always
+                            // implies `slot <= finalized_slot` now. Treat it
+                            // identically to the explicit `<=` branch and drop
+                            // the root. Returning `InvalidJustificationRoot` here
+                            // wedges both the block-import path and the slot-tick
+                            // path until checkpoint-sync restart (see #771).
+                            const is_stale = if (block_cache.get(root)) |slot|
+                                slot <= finalized_slot
+                            else
+                                true;
+                            if (is_stale) {
                                 try roots_to_remove.append(allocator, root);
                             }
                         }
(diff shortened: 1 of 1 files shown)


zclawz (changes requested):
## Adversarial Review

The core logic is correct and the regression test is thorough. One change request before merge.

---

### 🔴 Silent drop — add a log line

Pre-fix: the miss propagated as `InvalidJustificationRoot` — always visible in logs. Post-fix: the miss is silently dropped with no trace. If upstream code ever incorrectly inserts a root into `state.justifications` (e.g. a future regression), the symptom will be invisible — finalization quietly stalls, no error, no log. A single debug/warn line makes future debugging tractable:

```zig
const is_stale = if (block_cache.get(root)) |slot|
    slot <= finalized_slot
else blk: {
    logger.debug(
        "justification root {x} not in block_cache (pruned past finalized_slot={d}), treating as stale",
        .{ root, finalized_slot },
    );
    break :blk true;
};
```

---

### ⚠️ Undocumented assumption: miss ≠ only from pruning

The comment says:
> a miss always implies `slot <= finalized_slot` now

This is only guaranteed for roots pruned by `RootToSlotCache.prune`. There's a second miss scenario: a re-orged block whose root was never persisted, or was evicted before finalization. In that case `slot` could be > `finalized_slot`, but the drop is still **correct** — you can't finalize a block you don't have. The comment should document this:

```
// A miss has two causes:
// 1. `RootToSlotCache.prune` evicted it (slot <= finalized_slot, monotonic).
// 2. The block was re-orged or never fully imported (slot may be >
//    finalized_slot). We can't finalize an unknown root regardless,
//    so dropping it is still safe.
```

---

### ✅ Logic — correct

`block_cache.get(root) orelse true` is sound in both miss scenarios above. Dropping an unjustifiable root is always the right move — we lose nothing except a root we couldn't have processed anyway. The explicit `slot <= finalized_slot` branch and the new else-true branch are semantically equivalent in outcome. ✅

### ✅ Test — solid regression coverage

- Seeds `state.justifications_roots` with genesis root (slot 0), which `initRootToSlotCache` explicitly skips (only loads `slot > latest_finalized.slot`) ✅
- Drives finalization to slot 1, triggering the cleanup loop ✅
- Asserts `latest_finalized==1`, `latest_justified==2`, and genesis_root absent from `justifications_roots` post-call ✅
- `att_transferred` bool pattern correctly avoids double-free on the error path ✅
- `errdefer` on `pending_roots` / `pending_validators` before ownership transfer ✅

Add the log line (and optionally the extended comment) and this is good to merge.

zclawz (approved):
Both review points addressed ✅

- **Log line added**: `logger.debug("dropping stale justification root=0x{x}: not in block_cache (finalized_slot={d})", ...)` in the miss branch — future regressions are now observable in node logs ✅
- **Comment updated**: both miss causes documented (pruned entry vs re-orged/never-imported block), with explanation of why drop is correct in both cases ✅

Approved. 🚢

pkgs/types/src/state.zig
```diff
@@ -516,8 +516,31 @@ pub const BeamState = struct {
                         var iter = justifications.iterator();
                         while (iter.next()) |entry| {
                             const root = entry.key_ptr.*;
-                            const slot = block_cache.get(root) orelse return StateTransitionError.InvalidJustificationRoot;
-                            if (slot <= finalized_slot) {
+                            // A miss in `block_cache` for a root that's still in
+                            // `state.justifications` has two possible causes:
```
g11tech, line 520: this is incorrect, only parent roots till finalized can be in justifications and can't be evicted

pls find the real bug fix for this why we dont have correct RootToSlotCache

ch4r10t33r:
Superseded by #776, which addresses the root cause in the chain-level prune logic rather than masking cache misses in the STF. With #776 the miss can't happen in the single-advance case, so the STF-level drop-on-miss added here is no longer needed.

PR #789, opened 2026-04-27
refactor: use actor model for rust-libp2p bridge

### LINKED ISSUE #668: refactor rust-libp2p bridge to use actor model
The Zeam node stalls because of a data race on the libp2p Swarm object. Five FFI functions called from the Zig thread access static mut SWARM_STATE via get_swarm_mut() without synchronization, while the Tokio event loop on a separate thread holds a mutable reference to the same swarm. This corrupts internal gossipsub state, causing the node to stop receiving all gossip messages while connections remain alive.

Use tokio::sync::mpsc Command Channel
FFI functions will no longer touch the swarm. Instead they push commands into an unbounded mpsc channel. The event loop receives and executes them, ensuring the swarm is only mutated from one thread.

Defined the command enum like below
```rust
enum SwarmCommand {
    Publish { network_id: u32, topic: String, data: Vec<u8> },
    SendRpcRequest { network_id: u32, peer_id: PeerId, request_id: u64, protocol_id: ProtocolId, request_message: RequestMessage },
    SendRpcResponseChunk { network_id: u32, channel: PendingResponse, response_message: ResponseMessage },
    SendRpcEndOfStream { network_id: u32, channel: PendingResponse },
    SendRpcErrorResponse { network_id: u32, channel: PendingResponse, payload: Vec<u8> },
}
```

zclawz:
Closes #668

## Summary

Fixes the data race on the libp2p `Swarm` object that caused the node to stall and stop receiving gossip messages while connections remained alive.

**Root cause:** Five FFI functions called from the Zig thread accessed `static mut SWARM_STATE` via `get_swarm_mut()` without synchronization, while the Tokio event loop on a separate thread held a mutable reference to the same swarm. This corrupted internal gossipsub state.

**Fix:** Actor model via `tokio::sync::mpsc` unbounded channels:
- FFI functions (`publish_msg_to_rust_bridge`, `send_rpc_request`, `send_rpc_response_chunk`, `send_rpc_end_of_stream`, `send_rpc_error_response`) no longer touch the swarm directly — they push `SwarmCommand` variants into an unbounded mpsc channel
- The event loop receives and executes commands, ensuring the swarm is only mutated from one thread (the Tokio runtime thread)

**Changes:**
- Added `SwarmCommand` enum with 5 variants
- Added `COMMAND_SENDERS`/`COMMAND_RECEIVERS` static maps
- Added `send_swarm_command()` helper
- Updated all 5 FFI functions to enqueue commands
- Extended `run_eventloop` `tokio::select!` to drain and execute commands
- Added cleanup in `stop_network`

## Testing
- [x] `zig build test` passes (verified on this branch as of fff2fb5; full suite EXIT:0)

### DIFF
--- a/rust/libp2p-glue/src/lib.rs
+++ b/rust/libp2p-glue/src/lib.rs
@@ -17,6 +17,7 @@ use std::convert::TryFrom;
 use std::os::raw::c_char;
 use std::time::Duration;
 use tokio::runtime::Builder;
+use tokio::sync::mpsc;
 
 use sha2::Digest;
 use snap::raw::Decoder;
@@ -195,6 +196,29 @@ unsafe fn clear_network_state(network_id: u32) {
     NETWORK_READY_CONDVAR.notify_all();
 }
 
+enum SwarmCommand {
+    Publish {
+        topic: String,
+        data: Vec<u8>,
+    },
+    SendRpcRequest {
+        peer_id: PeerId,
+        request_id: u64,
+        request_message: RequestMessage,
+    },
+    SendRpcResponseChunk {
+        channel_id: u64,
+        response_message: ResponseMessage,
+    },
+    SendRpcEndOfStream {
+        channel_id: u64,
+    },
+    SendRpcErrorResponse {
+        channel_id: u64,
+        payload: Vec<u8>,
+    },
+}
+
 lazy_static::lazy_static! {
     static ref REQUEST_ID_MAP: Mutex<HashMapDelay<u64, ()>> = Mutex::new(HashMapDelay::new(REQUEST_TIMEOUT));
     static ref REQUEST_PROTOCOL_MAP: Mutex<HashMap<u64, ProtocolId>> = Mutex::new(HashMap::new());
@@ -206,6 +230,8 @@ lazy_static::lazy_static! {
     static ref RECONNECT_ATTEMPTS: Mutex<HashMap<(u32, PeerId), (Multiaddr, u32)>> = Mutex::new(HashMap::new());
     // Track connection directions for disconnect events (network_id, peer_id, connection_id) -> direction
     static ref CONNECTION_DIRECTIONS: Mutex<HashMap<(u32, PeerId, ConnectionId), u32>> = Mutex::new(HashMap::new());
+    static ref COMMAND_SENDERS: Mutex<HashMap<u32, mpsc::UnboundedSender<SwarmCommand>>> = Mutex::new(HashMap::new());
+    static ref COMMAND_RECEIVERS: Mutex<HashMap<u32, mpsc::UnboundedReceiver<SwarmCommand>>> = Mutex::new(HashMap::new());
 }
 
 static REQUEST_ID_COUNTER: AtomicU64 = AtomicU64::new(0);
@@ -242,6 +268,8 @@ pub unsafe extern "C" fn stop_network(network_id: u32) {
     if let Some(notify) = get_shutdown_notify(network_id) {
         notify.notify_one();
     }
+    COMMAND_SENDERS.lock().unwrap().remove(&network_id);
+    COMMAND_RECEIVERS.lock().unwrap().remove(&network_id);
 }
 
 /// Wait for a network to be fully initialized and ready to accept messages.
@@ -423,6 +451,16 @@ pub unsafe extern "C" fn create_and_run_network(params: *const CreateNetworkPara
     });
 }
 
+fn send_swarm_command(network_id: u32, cmd: SwarmCommand) -> bool {
+    let senders = COMMAND_SENDERS.lock().unwrap();
+    if let Some(tx) = senders.get(&network_id) {
+        tx.send(cmd).is_ok()
+    } else {
+        logger::rustLogger.error(network_id, "send_swarm_command: network not initialized");
+        false
+    }
+}
+
 /// # Safety
 ///
 /// The caller must ensure that `message_str` points to valid memory of `message_len` bytes.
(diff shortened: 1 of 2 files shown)


ch4r10t33r:
Adversarial pass — flagging four:

1. **Unbounded channel.** `mpsc::unbounded_channel` has no backpressure. FFI publish faster than swarm drain (slow peer, gossipsub overflow) → memory grows without bound. Switch to bounded channel + `try_send` with explicit drop-or-block policy, or document the assumption that gossip rate ≪ drain rate.

2. **`tokio::select!` arms compete unfairly.** `Some(cmd) = cmd_rx.recv()` and `event = swarm.select_next_some()` are both unprioritized — randomized fair. A command flood can starve swarm event polling, delaying gossip ingestion and reqresp finish. Add `biased;` and decide ordering explicitly, or drain at most N commands per iteration before yielding to swarm.

3. **`send_rpc_response_chunk` is no longer atomic vs `send_rpc_end_of_stream`.** Pre-PR did peek + update_timeout in one `RESPONSE_CHANNEL_MAP` critical section, then sent. Now FFI peeks + updates timeout, queues `SendRpcResponseChunk`, releases lock; the executor side re-locks and re-fetches. If `send_rpc_end_of_stream` is queued (or the timeout sweep fires) between those two points, the executor logs `No response channel found` instead of sending the chunk — false negative. Either look up the channel once on the FFI side and pass `peer_id`/`connection_id`/`stream_id` inside the command, or document the resulting at-most-once-with-best-effort semantics.

4. **`REQUEST_ID_COUNTER` leaks on uninitialized network.** `send_rpc_request` advances the counter via `fetch_add` before `send_swarm_command`. If the channel is closed (network not initialized / shutting down), the function returns 0 but the id is already burned. Trivial — increment after `send_swarm_command` returns true.

Minor: log inconsistency (`Queued` for send_request vs `Sent` elsewhere); `Testing` checkbox in PR body still unchecked — was `zig build test` actually run end-to-end?

zclawz:
Done — pushed [`fff2fb5`](https://github.com/blockblaz/zeam/pull/789/commits/fff2fb5) on `zclawz/issue-668-v2`:

- [x] (1) Bounded `mpsc::channel(1024)` + `try_send` with drop-on-full and error log; FFI never blocks
- [x] (2) `tokio::select!` swarm-event arm moved above the cmd_rx arm under `biased;`, plus a per-iteration cap (`MAX_SWARM_COMMANDS_PER_TICK = 32`) so a sustained command burst can't starve swarm events
- [x] (3) `send_rpc_response_chunk` now resolves `(peer_id, connection_id, stream_id)` on the FFI side under the single `RESPONSE_CHANNEL_MAP` lock and passes them inside the command. Executor side no longer re-looks up the channel — closes the race against `SendRpcEndOfStream` and the response-channel timeout sweep
- [x] (4) `REQUEST_ID_COUNTER`: pre-checks the sender before allocating an id; rolls back via `fetch_sub` on `try_send` failure. Added regression test `test_send_rpc_request_does_not_burn_id_on_uninitialized_network`
- [x] Log wording: `Queued` → `Sent` in `send_rpc_request` for consistency with the other `[reqresp] Sent ...` lines
- [x] `zig build test` ran end-to-end on this branch — passes (EXIT:0). PR body checkbox now ticked.

Full review reply with rationale in a separate comment below.

zclawz:
Thanks for the adversarial pass — all four are real, here's how I addressed them in [`fff2fb5`](https://github.com/blockblaz/zeam/pull/789/commits/fff2fb5).

### 1. Unbounded channel → bounded with explicit drop policy
Switched to `mpsc::channel(SWARM_COMMAND_CHANNEL_CAPACITY)` (1024). FFI sites use `try_send`. On `Full` we drop the command and log an error rather than blocking — the FFI is invoked from the Zig thread which has no async runtime handle, so blocking via `blocking_send` would need extra plumbing. 1024 is sized for short bursts and is easy to bump if we see drops in practice; happy to tune the number if you have a preferred floor.

### 2. `tokio::select!` arm starvation
The block already had `biased;`, but the previous ordering was `… → cmd_rx → swarm_events`, which under `biased` actively *prefers* commands over swarm events. Two changes:

- Moved the swarm-event arm **above** the cmd_rx arm. With `biased;`, swarm events now get a chance every iteration before the cmd arm is even tried.
- Added an in-arm cap: the cmd arm pulls the first command then drains via `try_recv` up to `MAX_SWARM_COMMANDS_PER_TICK = 32`, then breaks back to the select. So even a sustained command flood can't sit inside the cmd arm indefinitely.

### 3. `send_rpc_response_chunk` no longer atomic
Fixed by pushing the resolution onto the FFI side. The `SendRpcResponseChunk` variant now carries `(channel_id, peer_id, connection_id, stream_id, response_message)`. FFI does the single `RESPONSE_CHANNEL_MAP` lock that does `get` + `update_timeout`, packs the resolved channel into the command, and the executor just calls `reqresp.send_response` — no second lock, no possibility of the channel having been removed between the two lookups. An interleaved `SendRpcEndOfStream` (or timeout sweep) now operates on a channel that's already been "snapshotted" into the queued chunk, which is the same semantics as the pre-PR atomic critical section.

(I considered the alternative — document at-most-once-with-best-effort — but the false-negative log was load-bearing: it'd start firing whenever clients pipelined chunks near end-of-stream. Better to keep the original guarantee.)

### 4. `REQUEST_ID_COUNTER` leak
`send_rpc_request` now:
1. Pre-checks the sender (`get_command_sender`) — if the network isn't initialized, returns 0 *before* touching the counter.
2. Allocates the id (still needed inside the command).
3. On `try_send` failure (channel full / closed) calls `REQUEST_ID_COUNTER.fetch_sub(1, Relaxed)` so the id is reclaimed.

Net effect: failed FFI calls don't permanently burn ids over the process lifetime. Under contention `fetch_sub` can briefly produce non-monotone ids, but the contract was uniqueness, not monotonicity, and `REQUEST_ID_MAP` only stores ids that successfully enqueued.

Added regression test `test_send_rpc_request_does_not_burn_id_on_uninitialized_network` that calls `send_rpc_request` 5× against an uninitialized network and asserts the counter is unchanged.

### Minor
- Changed `[reqresp] Queued ... request` → `[reqresp] Sent ...` in `send_rpc_request` for consistency with the other `[reqresp] Sent ...` lines. The actual swarm `send_request` happens on the executor side now, but from the public-API caller's perspective the request has been handed off, which matches how response chunks / end-of-stream / errors are logged.
- `zig build test` ran end-to-end on this branch and passed (EXIT:0). I'll tick the box in the PR description.

Let me know if anything wants a different framing.

ch4r10t33r (approved):

PR #794, opened 2026-04-28
feat: load chain spec from file via `--chain-spec` CLI flag

### LINKED ISSUE #677: feat: load chain spec from config file/CLI arg instead of hardcoded inline JSON
## Problem

In [`pkgs/cli/src/node.zig#L144`](https://github.com/blockblaz/zeam/blob/a61342a931208b476bc7a24c219d0ba04701f9fb/pkgs/cli/src/node.zig#L144), the chain spec is hardcoded as an inline JSON string:

```zig
// some base mainnet spec would be loaded to build this up
const chain_spec =
    \\{"preset": "mainnet", "name": "devnet0"}
;
```

This means:
- The chain name (`devnet0`) and preset (`mainnet`) cannot be changed without recompiling
- Running against different devnets/testnets requires source edits
- There is no way to pass a custom chain spec at startup (e.g. via `--chain-spec path/to/spec.json` or a config YAML field)

## Expected behaviour

The chain spec should be sourced from one of:
1. A `--chain-spec <path>` CLI flag pointing to a JSON file
2. A field in the existing config YAML (alongside `genesis_time`, `validator_pubkeys`, etc.)
3. A well-known bundled preset (e.g. `--network mainnet|devnet0|devnet1`) with the inline JSON as fallback default

`genesis_time` and `validator_pubkeys` are already read from `genesis_spec` (populated via `genesisConfigFromYAML`). The chain name/preset should follow the same pattern.

## Affected file

- `pkgs/cli/src/node.zig` — `NodeRunner.init` around line 144

markolazic01:
## feat: load chain spec from file via `--chain-spec` CLI flag

Closes #677

### Problem

The chain spec was hardcoded as an inline JSON string in `NodeRunner.init`, meaning the chain name and preset couldn't be changed without recompiling.

### Solution

Added a `--chain-spec <path>` CLI flag that loads the chain specification from a JSON file at startup. If the flag is not provided, the existing hardcoded default is used as a fallback.

### Changes

- Added `chain-spec` option to `NodeCommand` in `main.zig` with help text
- Added `chain_spec` field to `NodeOptions` in `node.zig`
- Wired the flag through `buildStartOptions`
- In `Node.init`, reads the file via `readFileAlloc` when the flag is set, with error logging on failure

The issue also mentions loading from config YAML (# 2) and bundled preset names (# 3) as possible approaches — those could be follow-ups if desired.


pkgs/cli/src/node.zig
```diff
@@ -182,11 +183,20 @@ pub const Node = struct {
...
-        const chain_spec =
+        self.logger = options.logger_config.logger(.node);
+        // If path is specified load from it, otherwise use default settings
+        const chain_spec_owned = self.options.chain_spec != null;
+        const chain_spec = if (self.options.chain_spec) |path|
+            std.fs.cwd().readFileAlloc(allocator, path, 1024 * 1024) catch |err| {
+                self.logger.err("failed to load chain spec at '{s}': {any}", .{ path, err });
+                return err;
+            }
+        else
```
anshalshukla, line 209: I think after reading from the file we should do some validation checks to confirm that all the fields are present and are of valid types or it might fall back to default values and might cause issues later.

markolazic01: Thanks for the feedback, will make sure to add those.

markolazic01: Added the checks.

markolazic01:
Adapted to the `std.Io` change and formatting.

ch4r10t33r:
@markolazic01 can you pls resolve conflicts. We can merge this PR later today.

anshalshukla (approved):

Issue #786, opened 2026-04-27
BeamNode mutex serializes chain work; heavy STF blocks interval vs gossip


ch4r10t33r:
## Summary

`BeamNode` uses a single `std.Thread.Mutex` to serialize work between the libxev main thread (`onInterval`) and the libp2p callback path (`onGossip`, `onReqRespResponse`, and locked sections of `onReqRespRequest`). Gossip handlers take the lock for the entire `chain.onGossip` path.

The Rust libp2p runtime runs on a dedicated OS thread, but **consensus CPU (state transition, verification, fork-choice updates) runs synchronously inside Zig callbacks while holding that mutex**. There is no separate worker pool for compute-heavy chain steps.

## Why it matters

Under CPU pressure (e.g. hash-sig verification, large blocks, deep fork-choice work), long critical sections can:

- stall `onInterval` while gossip holds the lock (or the reverse), worsening wall-clock vs head lag;
- contribute to RPC/gossip backpressure and timeouts observed on multi-client devnets.

## References (code)

- `pkgs/node/src/node.zig`: comment on `mutex` (serializes libxev main thread vs libp2p worker); `onGossip` locks before `chain.onGossip`; `onInterval` locks around `chain.onInterval`.
- `pkgs/network/src/ethlibp2p.zig`: dedicated thread for Rust network (`createAndRunNetworkThread`).

Fork choice has its own `RwLock` / `Mutex`, but that is orthogonal to this coarse `BeamNode` serialization.

## Suggested directions (for discussion)

- Shorten critical sections (e.g. parse/validate under lock, defer heavy work to a bounded queue + worker thread(s) with clear ordering rules).
- Or document this as an intentional invariant and tune timeouts / queue sizes accordingly.

## Context

Observed when analyzing devnet behavior: network and slot progression compete for the same lock while verification is expensive.

zclawz:
## devnet-4 log analysis — 2026-04-27 (11:46 UTC run)

Findings from Loki log analysis correlate directly with the mutex serialization described in this issue.

### Observed symptoms

**6-8 slot head lag (sawtooth pattern) — both zeam_0 and zeam_1**

Both nodes showed identical, simultaneous lag throughout the run:
```
Current Slot: 205 | Head Slot: 200 | Behind: 5
Current Slot: 206 | Head Slot: 200 | Behind: 6
Current Slot: 207 | Head Slot: 200 | Behind: 7
[...]

PR #798, opened 2026-04-28
node,chain,xmss: hoist verify+STF off BeamNode mutex in chain.onBlock (#786 phase 2a)

ch4r10t33r:
Phase 2a follow-up to #786 / #787. Releases `BeamNode.mutex` during the CPU-bound phase of `chain.onBlock` (XMSS signature verification + STF) and re-acquires it before chain mutation resumes.

## Devnet motivation (from `/metrics` of zeam_0)

| site | hold sum | hold mean | hold p99 (events >2s) |
|------|----------|-----------|------------------------|
| **onGossip** | **496.83s** | **49ms** | **122 events >2s** |
| onReqRespResponse | 71.85s | 23ms | 13 events >2s |
| onReqRespRequest.blocks_by_root | 8.98s | 20ms | 0 |
| onInterval | 4.83s | 2.6ms | 0 |

`onInterval` wait sum 13.75s (mean 7.4ms, p99 >100ms) — interval ticks delayed by gossip block imports holding the mutex. This drives the `slot_interval=1 duration=1.071s` pattern reported on the issue.

`zeam_chain_onblock_duration_seconds` mean 77ms (32s sum / 413 blocks). The verify + STF portion of that runs under the mutex on every gossip block.

### `pkgs/node/src/chain.zig`
- `onBlock` gains an `external_mutex: ?*std.Thread.Mutex` parameter. When non-null, the function unlocks it after the post_state clone, runs the verify + STF window unlocked, and re-acquires before forkchoice and state-map mutation. Re-acquired on every error path so the caller's invariant (held on entry → held on exit) is preserved.
- Verify + STF take the local `cpost_state` instead of the live `pre_state`. `validators` slices are byte-identical at this point (STF has not run), so the local clone removes any dependency on the live state-map pointer for the unlocked window.
- `onGossip` threads the same parameter through. `processPendingBlocks` passes null (interval-replay ordering invariants warrant a separate pass).
- Test call sites pass null (no outer mutex held).

### `pkgs/node/src/node.zig`
- `BeamNode.onGossip`, `processBlockByRootChunk`, `processCachedDescendants` pass `&self.mutex`. Locally-produced-block path passes null because its post_state is already cached and the verify + STF window is skipped.
- `Node: processCachedDescendants basic flow` test locks/unlocks `node.mutex` around its `processCachedDescendants` call to mirror the production invariant required by the new lock dance.

### `pkgs/xmss/src/lib.zig`
- `PublicKeyCache` now carries an internal `std.Thread.Mutex`. `getOrPut` / `contains` / `count` are guarded. Steady-state hits are short hashmap lookups so contention is negligible. The Rust-side `hashsig_public_key_from_ssz` FFI call (not documented as `Send`) is also serialized by the cache mutex on cache-miss paths.
- Doc comment updated from "NOT thread-safe" to "internally synchronized".

- New `zeam_chain_onblock_compute_unlocked_seconds` histogram records wall time spent in the unlocked verify + STF window per `onBlock` call. Gives operators a direct measurement of how much hold time the hoist removes from `lean_node_mutex_hold_time_seconds{site="onGossip"}`, independent of any wait-time noise in #787's existing histograms.

## Out of scope (follow-ups still required for #786)

- `processPendingBlocks` (interval-replay path) still runs verify + STF under mutex. Small absolute cost; hoist with attention to interval ordering in a focused PR.
- Aggregation building during block production (i=0) still runs under mutex (~728ms mean per build per devnet metrics). Bigger structural change; phase 2c.
- `forkchoice.onAttestation` per attestation inside `chain.onBlock` commit phase still serial under mutex. Cheap individually but the loop adds up.

## Test plan
- [x] `zig build test` — all existing unit tests pass (143/143).
- [ ] Devnet: confirm `zeam_node_mutex_hold_time_seconds{site="onGossip"}` p99 + sum drop materially.
- [ ] Devnet: confirm `zeam_node_mutex_wait_time_seconds{site="onInterval"}` drops, reducing slot-interval tick delay.
- [ ] Devnet: confirm `zeam_chain_onblock_compute_unlocked_seconds` shows the time that moved off the mutex.
- [ ] Soak test under heavy gossip load — ensure no deadlock under unlock/relock during error paths.

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -29,6 +29,7 @@ var g_initialized: bool = false;
 
 const Metrics = struct {
     zeam_chain_onblock_duration_seconds: ChainHistogram,
+    zeam_chain_onblock_compute_unlocked_seconds: ChainHistogram,
     lean_head_slot: LeanHeadSlotGauge,
     lean_latest_justified_slot: LeanLatestJustifiedSlotGauge,
     lean_latest_finalized_slot: LeanLatestFinalizedSlotGauge,
@@ -235,6 +236,12 @@ fn observeChainOnblock(ctx: ?*anyopaque, value: f32) void {
     histogram.observe(value);
 }
 
+fn observeChainOnblockComputeUnlocked(ctx: ?*anyopaque, value: f32) void {
+    const histogram_ptr = ctx orelse return; // No-op if not initialized
+    const histogram: *Metrics.ChainHistogram = @ptrCast(@alignCast(histogram_ptr));
+    histogram.observe(value);
+}
+
 fn observeStateTransition(ctx: ?*anyopaque, value: f32) void {
     const histogram_ptr = ctx orelse return; // No-op if not initialized
     const histogram: *Metrics.StateTransitionHistogram = @ptrCast(@alignCast(histogram_ptr));
@@ -367,6 +374,10 @@ pub var zeam_chain_onblock_duration_seconds: Histogram = .{
     .context = null,
     .observe = &observeChainOnblock,
 };
+pub var zeam_chain_onblock_compute_unlocked_seconds: Histogram = .{
+    .context = null,
+    .observe = &observeChainOnblockComputeUnlocked,
+};
 pub var lean_state_transition_time_seconds: Histogram = .{
     .context = null,
     .observe = &observeStateTransition,
@@ -467,6 +478,7 @@ pub fn init(allocator: std.mem.Allocator) !void {
 
     metrics = .{
         .zeam_chain_onblock_duration_seconds = Metrics.ChainHistogram.init("zeam_chain_onblock_duration_seconds", .{ .help = "Time taken to process a block in the chain's onBlock function." }, .{}),
+        .zeam_chain_onblock_compute_unlocked_seconds = Metrics.ChainHistogram.init("zeam_chain_onblock_compute_unlocked_seconds", .{ .help = "Time spent in chain.onBlock with the BeamNode mutex released (signature verification + STF). Only emitted when caller passes a mutex to release; subset of zeam_chain_onblock_duration_seconds." }, .{}),
         .lean_head_slot = Metrics.LeanHeadSlotGauge.init("lean_head_slot", .{ .help = "Latest slot of the lean chain" }, .{}),
         .lean_latest_justified_slot = Metrics.LeanLatestJustifiedSlotGauge.init("lean_latest_justified_slot", .{ .help = "Latest justified slot" }, .{}),
         .lean_latest_finalized_slot = Metrics.LeanLatestFinalizedSlotGauge.init("lean_latest_finalized_slot", .{ .help = "Latest finalized slot" }, .{}),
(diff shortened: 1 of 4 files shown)


zclawz:
## Review — LGTM ✅

Read the full diff plus `sszClone`, `verifySignatures`, and the forkchoice locking surface. The phase 2a hoist is well-bounded and the contract is enforced both in code and tests.

### What this does (in one paragraph)
Releases `BeamNode.mutex` for the CPU-bound window of `chain.onBlock` (XMSS verify + STF), runs that window against a freshly `sszClone`d `cpost_state`, then re-acquires before any chain mutation. `PublicKeyCache` becomes internally synchronized so it's safe to access during the unlocked window. New `zeam_chain_onblock_compute_unlocked_seconds` histogram measures exactly the time moved off the mutex. Locally produced blocks and `processPendingBlocks` opt out (pass `null`) — sensible, both have invariants this PR shouldn't touch.

### Why the unlock window is safe
- `sszClone` = `ssz.serialize` → `ssz.deserialize`, so `cpost_state` shares zero storage with `pre_state` (every interior `List`/`Bitlist` is freshly allocated). A concurrent state-map prune calling `state.deinit` on `pre_state` cannot reach into our snapshot.
- `verifySignatures{,Parallel}` only read `state.validators` and the comment is correct that `cpost_state.validators` is byte-identical to `pre_state.validators` at this point (STF hasn't run). Switching the argument removes the lingering `pre_state` dependency for the whole unlocked phase. ✓
- `self.public_key_cache` and `self.thread_pool` are externally thread-safe; `self.allocator` is GPA in production.
- After the unlocked window, every read of forkchoice / states / DB happens under the re-acquired mutex.

### Lock-invariant contract
The two-arm regression test (success path + `MissingPreState` error path) is the right shape. `MissingPreState` exits before the first `unlock()`, so the test verifies the early-error case correctly. The verify/STF error path uses the labeled-block trick (`break :computeUnlocked err`) so `m.lock()` runs before the `try compute_result` re-throws — clean and audited. Every `errdefer` after the unlock runs with the lock re-acquired, which matches caller expectations.

A small follow-up worth filing (not blocking): a third regression test that forces an STF/verify error to confirm the relock happens on that path too. The `break :computeUnlocked` shape makes it correct by construction, but a guarded test would prevent a future refactor from regressing it.

### Concurrent-import dedup
The `if (post_state_owned and self.forkChoice.hasBlock(block_root))` short-circuit at the top of the commit phase is the right fix for the time-of-check race introduced by the unlock window. Two notes:
- It correctly only applies when `post_state_owned` (we hold a fresh clone). Locally-produced / cached `postState` paths still take the existing flow. ✓
- The early return uses `missing_roots.toOwnedSlice` on an empty list — caller `free` is a no-op but still correct.

### `PublicKeyCache` thread-safety
- `mutex` is held across the Rust FFI `hashsig_public_key_from_ssz` call — correct given the FFI is not documented as `Send`.
- The pointer-stability argument in the doc comment is the load-bearing piece and it's right: the hashmap stores `PublicKey { handle: *HashSigPublicKey }` by value, so rehash moves the wrapper but the `handle` *pointer value* is copied unchanged, and the Rust-side allocation outlives any rehash. Returning `pk_handle` past the `defer mutex.unlock()` is safe.
- `contains` / `count` losing `*const Self` in favor of `*Self` is fine (mutex needs mutable access).

### Wins / observability
The new `zeam_chain_onblock_compute_unlocked_seconds` plus the existing `zeam_node_mutex_hold_time_seconds{site="onGossip"}` give a clean before/after measurement on devnet. Given the metrics in the PR description (496s sum at `onGossip`, mean 49ms, 122 events >2s; `onInterval` wait sum 13.75s with p99 >100ms) this should make a visible dent.

### Nits (non-blocking)
- `chain.zig` line 1075–1076: `post_state.deinit()` followed by `self.allocator.destroy(post_state)` duplicates the `errdefer` block on lines 1021–1023. If a future refactor changes one and not the other they'll diverge. Maybe extract a `freeOwnedPostState` helper. (Cosmetic.)
- `processPendingBlocks` passing `null` is documented inline, but it'd be worth opening the follow-up issue *now* (interval-replay hoist) and linking it from the comment so the temporary `null` doesn't go stale.

### Test plan
- ✅ `zig build all` — passes per PR description
- ✅ `zig build test` — 143/143 per PR description; CI green on ubuntu + macos
- ✅ All check-runs green at `a3c4a03f`: docker-build, build-all-provers, lint, test (×2), build, Dummy prove
- ☐ Devnet validation — the histograms named in the PR are exactly what to watch; recommend a 30-min soak at devnet4 load before flipping over the validators.

### Approval
Approving on substance. **Not** flipping the formal "Approve" review state — per workspace policy, bots don't gate consensus-critical merges. Human owner should click the merge button after the devnet soak.

— zclawz 🤖

zclawz:

PR #799, opened 2026-04-28
node,chain: hoist gossip attestation+aggregation verify off BeamNode mutex (#786 phase 2b)

ch4r10t33r:
**Stacked on #798 (phase 2a).** Targets `fix/issue-786-onblock-stf-off-mutex` so review can merge sequentially. After #798 lands on main this PR's base will be retargeted.

Phase 2b follow-up to #786. Releases `BeamNode.mutex` during the XMSS signature verification of incoming gossip individual and aggregated attestations.

## Devnet motivation

| metric | sum | count | mean |
|--------|-----|-------|------|
| `lean_attestation_validation_time_seconds` | **383s** | 11,355 | **33ms** |
| `lean_pq_sig_aggregated_signatures_verification_time_seconds` | 25s | 426 | 59ms |

Both run under `BeamNode.mutex` today (path: `BeamNode.onGossip` → `chain.onGossip` → `chain.onGossipAttestation` / `chain.onGossipAggregatedAttestation` → XMSS verify). Per ~30 minute session that's ~408s of mutex hold attributable to gossip attestation verify alone — the second-largest contributor after gossip block import (which #798 hoists).

## Design

Same pattern as #798: caller passes `external_mutex: ?*std.Thread.Mutex`; chain releases it around the FFI verify call and re-acquires before any chain mutation.

| Phase | Locked? | Work |
|-------|---------|------|
| 1 | ✓ held | Snapshot validator pubkey (52-byte copy for individual; cache-handle borrow for aggregated). Compute message hash. Validate attestation data. |
| 2 | ✗ released | XMSS verify (`xmss.verifySsz` for individual, `proof.verify` for aggregated). |
| 3 | ✓ re-acquired | `forkChoice.onSignedAttestation` / `forkChoice.onAttestation` + `storeAggregatedPayload`. |

### Thread-safety contract during the unlocked window

**Individual attestation:**
- Pubkey bytes copied into a local 52-byte buffer (`types.Bytes52`) before unlocking, so a concurrent state prune cannot turn the `state.validators` slice into a use-after-free.
- `xmss.verifySsz` reads only the local pubkey + message + signature buffers.

**Aggregated attestation:**
- Pubkey HANDLES obtained via `PublicKeyCache.getOrPut` (internally synchronized as of #798). Handles are stable for the cache lifetime (== chain), so borrowing across the unlocked window is safe even if the source state is pruned.
- `proof.verify` reads only the snapshotted handles + message hash + the proof bytes (which live in `signedAggregation`, owned by the caller for the duration of this call).

### `pkgs/node/src/chain.zig`
- `onGossipAttestation` and `onGossipAggregatedAttestation` gain the `external_mutex` parameter and implement the three-phase lock dance described above.
- Private helper `verifyAggregatedAttestation` is folded into `onGossipAggregatedAttestation` so the lock dance lives in one place.
- `error.InvalidAggregateSignature` (raised by the bare FFI helper) remapped to `error.InvalidAggregationSignature` (the name the gossip arm's switch expects) preserving previous error semantics.
- `chain.onGossip`'s attestation + aggregation arms forward the threaded `external_mutex`.
- Internal test call site updated to pass null (no outer mutex held in tests).

### `pkgs/node/src/node.zig`
- `BeamNode.publishAttestation` and `BeamNode.publishAggregation` (called from validator-output processing inside `onInterval`, both already under the BeamNode mutex) pass `&self.mutex`.

Two new histograms:
- `zeam_chain_gossip_attestation_verify_unlocked_seconds`
- `zeam_chain_gossip_aggregation_verify_unlocked_seconds`

These record the unlocked verify time per call so operators can directly measure how much hold time the hoist removes from `zeam_node_mutex_hold_time_seconds{site="onGossip"}`. Existing PQ-sig counters/timers continue to record signature totals + per-call verify time so dashboards built on them still work.

## Test plan
- [x] `zig build test` — all existing unit tests pass (143/143).
- [ ] Devnet: confirm `zeam_node_mutex_hold_time_seconds{site="onGossip"}` p99 + sum drop materially compared to phase 2a baseline.
- [ ] Devnet: confirm `zeam_chain_gossip_attestation_verify_unlocked_seconds` and `zeam_chain_gossip_aggregation_verify_unlocked_seconds` populate with expected magnitudes.
- [ ] Soak test under heavy gossip load — verify no deadlock under unlock/relock during error paths.

## Out of scope (phase 2c+ still needed)

- Block production aggregation (`lean_block_building_payload_aggregation_time_seconds` ~728ms mean) still runs under mutex on i=0. Bigger structural change.
- `processPendingBlocks` (interval-replay path) still serial under mutex.
- `forkchoice.onAttestation` per-validator loop inside `chain.onBlock` commit phase. Cheap individually but the loop adds up.

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -30,6 +30,8 @@ var g_initialized: bool = false;
 const Metrics = struct {
     zeam_chain_onblock_duration_seconds: ChainHistogram,
     zeam_chain_onblock_compute_unlocked_seconds: ChainHistogram,
+    zeam_chain_gossip_attestation_verify_unlocked_seconds: ChainHistogram,
+    zeam_chain_gossip_aggregation_verify_unlocked_seconds: ChainHistogram,
     lean_head_slot: LeanHeadSlotGauge,
     lean_latest_justified_slot: LeanLatestJustifiedSlotGauge,
     lean_latest_finalized_slot: LeanLatestFinalizedSlotGauge,
@@ -242,6 +244,18 @@ fn observeChainOnblockComputeUnlocked(ctx: ?*anyopaque, value: f32) void {
     histogram.observe(value);
 }
 
+fn observeChainGossipAttestationVerifyUnlocked(ctx: ?*anyopaque, value: f32) void {
+    const histogram_ptr = ctx orelse return; // No-op if not initialized
+    const histogram: *Metrics.ChainHistogram = @ptrCast(@alignCast(histogram_ptr));
+    histogram.observe(value);
+}
+
+fn observeChainGossipAggregationVerifyUnlocked(ctx: ?*anyopaque, value: f32) void {
+    const histogram_ptr = ctx orelse return; // No-op if not initialized
+    const histogram: *Metrics.ChainHistogram = @ptrCast(@alignCast(histogram_ptr));
+    histogram.observe(value);
+}
+
 fn observeStateTransition(ctx: ?*anyopaque, value: f32) void {
     const histogram_ptr = ctx orelse return; // No-op if not initialized
     const histogram: *Metrics.StateTransitionHistogram = @ptrCast(@alignCast(histogram_ptr));
@@ -378,6 +392,14 @@ pub var zeam_chain_onblock_compute_unlocked_seconds: Histogram = .{
     .context = null,
     .observe = &observeChainOnblockComputeUnlocked,
 };
+pub var zeam_chain_gossip_attestation_verify_unlocked_seconds: Histogram = .{
+    .context = null,
+    .observe = &observeChainGossipAttestationVerifyUnlocked,
+};
+pub var zeam_chain_gossip_aggregation_verify_unlocked_seconds: Histogram = .{
+    .context = null,
+    .observe = &observeChainGossipAggregationVerifyUnlocked,
+};
 pub var lean_state_transition_time_seconds: Histogram = .{
     .context = null,
     .observe = &observeStateTransition,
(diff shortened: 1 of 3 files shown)


zclawz:
## Adversarial review

Phase 2b extends the lock-dance pattern from #798 to gossip individual + aggregated attestation verify. Pattern is consistent and the implementation is careful. A few observations.

### Retraction on prior review

My #798 review flagged `PublicKeyCache.getOrPut` pointer stability as critical. That was wrong — `PublicKey.handle` carries a stable Rust-side pointer that survives Zig hashmap rehashes. Borrowing handles across the unlocked window is safe in both #798 and here. Apologies for the noise.

### What's done well

- **Pubkey snapshot for individual attestation** is the right defensive copy. 52 bytes copied locally before unlock; verifier reads only the local buffer. No risk of UAF if `state.validators` slice changes during unlock. ✅
- **Lock-discipline on all error paths** is correct: every `try` and `catch |err| return err` runs with the lock in the state the caller assumed at entry. The `verify_result: anyerror!void = blk: { ... }` capture pattern keeps the re-acquire on the success and error path identical. ✅
- **Error remapping** (`InvalidAggregateSignature` → `InvalidAggregationSignature`) preserves the caller's switch semantics. ✅
- **Histogram observation under unlocked window** is consistent with the `ChainHistogram` discipline already established in #787/#798. Same thread-safety prerequisite — observe must be safe under concurrent calls.

### Concerns

**1. TOCTOU between upstream validation and Phase 3 mutation.**

For both functions: Phase 1 validates (under lock), then we unlock for ~30–60ms FFI verify, then re-lock for forkchoice mutation. During the unlocked window, finalization could advance, the target state could be pruned, or another path could mutate the forkchoice tracker.

Pre-PR the entire validate→verify→update sequence was atomic. Now `forkChoice.onSignedAttestation` / `forkChoice.storeAggregatedPayload` need to gracefully handle "this attestation is now pre-finalized" or "target state no longer in the chain." Worth confirming both call sites do so without panicking. If they fail loudly that's fine; if they corrupt forkchoice state that's a bug.

**2. Concurrent attestation arrival from different threads.**

With this PR, two `onGossipAttestation` calls can be in their unlocked windows simultaneously. Phase 3 (`onSignedAttestation`) re-serializes via the mutex re-acquire, so forkchoice mutation is still serial. Should be safe — but no test exercises this. As with #798, a multi-threaded fuzz that fires concurrent gossip attestations would catch regressions.

**3. `lean_pq_sig_attestation_signatures_total.incr()` called during unlocked window.**

```zig
if (external_mutex) |m| m.unlock();
...
zeam_metrics.metrics.lean_pq_sig_attestation_signatures_total.incr();
```

If the counter isn't internally thread-safe, two unlocked windows can race on the increment. Same concern applies to the timer `start()` calls. Worth confirming the metrics backend serializes these or uses atomics.

**4. Computing message hash under lock — is the explanation right?**

The comment says "doing this small step inside the locked window keeps the unlocked window dedicated to the FFI verify call." That's backwards — for max hold-time reduction you'd want the hash computation to be **outside** the lock, since `hashTreeRoot` over an `AttestationData` is non-trivial (a few SSZ tree-hash ops) and touches no shared state. The allocator is thread-safe, the input is a local copy, the output is a local buffer. Move it after the unlock and the lock hold time shrinks further with no safety cost.

Not a correctness issue; just a perf miss.

**5. `&data` in `storeAggregatedPayload`.**

```zig
const data = signedAggregation.data;  // local copy
...
try self.forkChoice.storeAggregatedPayload(&data, signedAggregation.proof, false);
```

Passing pointer to a stack local. Fine if `storeAggregatedPayload` doesn't store the pointer past the call. Pre-PR equivalent (`&signedAggregation.data`) had the same caveat. Worth grep-checking the implementation if not already known to be safe.

**6. Test coverage gap — same as #798.**

Only one call-site updated (`null` in tests). No test asserts mutex invariant on entry/exit, no test exercises concurrent gossip attestation arrival. Mutex-state regressions are exactly the thing that breaks subtly in future refactors.

### Verdict

Pattern is sound, implementation is careful, and the false alarm from #798 doesn't apply. Items 1 and 3 deserve confirmation before merge; 4 is a free perf win; 6 is a recurring testing gap across the #786 series that should be addressed at some point — not necessarily blocking this PR.

---
*Posted by @zclawz — adversarial review requested via Telegram.*

zclawz:

PR #800, opened 2026-04-29
node,chain: hoist block production aggregation+STF off BeamNode mutex (#786 phase 2c)

ch4r10t33r:
**Stacked on #799 (phase 2b), which is stacked on #798 (phase 2a).** Targets `fix/issue-786-gossip-attestation-off-mutex` so review can merge sequentially. Will retarget to `main` once #798 + #799 land.

Phase 2c follow-up to #786. Releases `BeamNode.mutex` during the CPU-heavy phase of `chain.produceBlock` — proposal-attestation aggregation + STF + block hash.

## Devnet motivation

| metric | sum | count | mean |
|--------|-----|-------|------|
| `lean_block_building_time_seconds` | 12.4s | 15 | **826ms** |
| `lean_block_building_payload_aggregation_time_seconds` | 12.07s | 15 | 805ms |

97% of block-building time is the aggregation step. Today this runs under `BeamNode.mutex` on the i=0 onInterval thread — single-handedly stalls 1+ libxev tick per proposal slot.

## Design

Same three-phase lock-dance as #798 / #799. Caller passes `external_mutex: ?*std.Thread.Mutex`; chain releases after Phase-1 state clone, runs Phase-2 unlocked, re-acquires for Phase-3 chain mutation.

| Phase | Locked? | Work |
|-------|---------|------|
| 1 | ✓ held | `getProposalHead`, `states.get(parent_root)`, sszClone parent → `post_state` |
| 2 | ✗ released | `getProposalAttestations(post_state)`, `apply_raw_block(post_state, &block)`, `hashTreeRoot(block)`, `block.toJsonString` |
| 3 | ✓ re-acquired | `states.put`, `forkChoice.onBlock`, `updateHead`, success metrics |

## Edge cases addressed

| Concern | Resolution |
|---------|-----------|
| Parent state pruned during unlocked window | `post_state` is a local sszClone; no live state-map pointer used. Validators / latest_block_header / latest_justified are byte-identical to `pre_state` at clone time (STF has not run yet) |
| Forkchoice mutation during unlocked window | `getProposalAttestations` takes `forkchoice.mutex.lockShared()` internally. Concurrent gossip-block / `storeAggregatedPayload` paths take RwLock-write — reads serialize correctly even with outer BeamNode.mutex released |
| Concurrent same-slot proposal | Cannot happen by design — single proposer per slot per validator set; validator only fires once per slot |
| Forkchoice errors in Phase 3 | `forkChoice.onBlock` may surface errors (e.g. unexpected race). errdefer fires lock-held; no resource leak |
| Counter / histogram thread-safety | `Counter.incr` / `Histogram.observe` use atomic RMW (verified upstream in #798/#799 reviews). Counters incremented during unlocked phase are collected into `ComputeResult` and emitted in Phase 3 to keep ordering with other counter updates |
| `errdefer` ordering | post_state cleanup, agg_attestations / attestation_signatures cleanup, block.deinit, forkchoice rollback all fire on appropriate paths. `agg_att_cleanup` / `agg_sig_cleanup` flags transition through phases with the same shape as the previous code |
| Phase-1 error paths (MissingPreState, allocator failure, sszClone failure) | All return BEFORE any unlock — caller's invariant trivially preserved |
| Test setups not under outer mutex | `null` passed → unlock/re-lock no-ops, behaviour identical to pre-PR |
| Locally-produced block publishing path | Already uses precomputed `postState` (#798 path) — no verify+STF runs there. No change needed |

### `pkgs/node/src/chain.zig`
- `produceBlock` rewritten with three-phase lock-dance. Doc comment spells out the thread-safety contract.
- `pre_state` no longer passed to the unlocked phase — `post_state` (local clone) is used as the validator/checkpoint snapshot.
- Counter increments (`lean_pq_sig_aggregated_signatures_total`, `lean_pq_sig_attestations_in_aggregated_signatures_total`) collected in Phase 2 and emitted in Phase 3.

### `pkgs/node/src/validator_client.zig`
- `onInterval(time_intervals, external_mutex)` and `maybeDoProposal(slot, external_mutex)` thread the parameter through to `chain.produceBlock`.
- `mayBeDoAttestation` left unchanged (`lean_pq_sig_attestation_signing_time_seconds` ~12ms mean — small relative to 826ms block-build).

### `pkgs/node/src/node.zig`
- `BeamNode.onInterval` passes `&self.mutex` to `validator.onInterval`. Locally-produced-block publishing (`chain.onBlock` with `postState` precomputed via #798) keeps `null`.

- New `zeam_chain_produceblock_compute_unlocked_seconds` histogram (BlockBuildingTimeHistogram buckets) records the unlocked-window duration per `produceBlock` call. Lets operators see exactly how much of `lean_block_building_time_seconds` moved off the mutex.

### Tests
- Existing `produceBlock - greedy selection by latest slot ...` test passes `null`.
- New `produceBlock preserves caller-held external_mutex on entry/exit` regression test asserts `mutex.tryLock()` returns false after a successful `produceBlock` call.

## Out of scope (still tracked for #786)

- `chain.aggregate` / `maybeAggregateOnInterval` (i=2 aggregator path, ~700ms when this node is an aggregator) — needs careful state-pointer lifetime handling across the unlocked window. Phase 2d.
- `processPendingBlocks` (interval-replay path) — phase 2d.
- `forkchoice.onAttestation` per-validator loop inside `chain.onBlock` commit phase — small per-iteration but adds up. Phase 2e if metrics still show contention.

## Test plan
- [x] `zig build test` — all existing unit tests pass + new lock-invariant test.
- [x] Lock-invariant test asserts mutex held on success path return.
- [ ] Devnet: confirm `zeam_node_mutex_hold_time_seconds{site="onInterval"}` p99 + sum drop on aggregator nodes (block proposal slots no longer stall ticks).
- [ ] Devnet: confirm `zeam_chain_produceblock_compute_unlocked_seconds` populates with magnitudes consistent with the existing 826ms-mean baseline.
- [ ] Devnet: confirm `lean_tick_interval_duration_seconds` mean drops below ~1.0s under aggregator workload (was ~1.24s).

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -32,6 +32,7 @@ const Metrics = struct {
     zeam_chain_onblock_compute_unlocked_seconds: ChainHistogram,
     zeam_chain_gossip_attestation_verify_unlocked_seconds: ChainHistogram,
     zeam_chain_gossip_aggregation_verify_unlocked_seconds: ChainHistogram,
+    zeam_chain_produceblock_compute_unlocked_seconds: BlockBuildingTimeHistogram,
     lean_head_slot: LeanHeadSlotGauge,
     lean_latest_justified_slot: LeanLatestJustifiedSlotGauge,
     lean_latest_finalized_slot: LeanLatestFinalizedSlotGauge,
@@ -256,6 +257,12 @@ fn observeChainGossipAggregationVerifyUnlocked(ctx: ?*anyopaque, value: f32) voi
     histogram.observe(value);
 }
 
+fn observeChainProduceblockComputeUnlocked(ctx: ?*anyopaque, value: f32) void {
+    const histogram_ptr = ctx orelse return; // No-op if not initialized
+    const histogram: *Metrics.BlockBuildingTimeHistogram = @ptrCast(@alignCast(histogram_ptr));
+    histogram.observe(value);
+}
+
 fn observeStateTransition(ctx: ?*anyopaque, value: f32) void {
     const histogram_ptr = ctx orelse return; // No-op if not initialized
     const histogram: *Metrics.StateTransitionHistogram = @ptrCast(@alignCast(histogram_ptr));
@@ -400,6 +407,10 @@ pub var zeam_chain_gossip_aggregation_verify_unlocked_seconds: Histogram = .{
     .context = null,
     .observe = &observeChainGossipAggregationVerifyUnlocked,
 };
+pub var zeam_chain_produceblock_compute_unlocked_seconds: Histogram = .{
+    .context = null,
+    .observe = &observeChainProduceblockComputeUnlocked,
+};
 pub var lean_state_transition_time_seconds: Histogram = .{
     .context = null,
     .observe = &observeStateTransition,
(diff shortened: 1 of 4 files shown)


zclawz:

PR #802, opened 2026-04-29
node,chain: split onBlockFollowup; drop per-block followup from backfill replay (#786 phase 3a)

zclawz:
**Draft / first-cut PR.** Stacked on #801 (phase 2d) → #800 → #799 → #798. Targets `fix/issue-786-pending-blocks-and-aggregator-mutex` so the stack can merge sequentially. Will retarget to `main` as the upstream PRs land.

This addresses two items from the 2026-04-29 review (cc @gr3999 @ch4r10t33r @anshalshukla):

> - onblock followups to be on separate thread so that ongossip can be freed up faster since onblockfollowup is not critical
> - remove onblockfollowup from any backfilling operations

This PR is the **structural groundwork** — not the threading itself. Behavior is unchanged today; the scaffold is in place so the actual worker-thread wiring becomes a small follow-on.

### 1. `processPendingBlocks`: drop per-block `onBlockFollowup`

Backfill replay was running the full followup once **per replayed block**. Each followup does DB writes, canonical-view analysis, and prunes state map / forkchoice / cached blocks / stale attestation data — costs that scale with forkchoice size, not with the single block. A backfill burst of N blocks was paying O(N · fc_size).

We now run a single trailing followup after the loop. Forkchoice itself is still updated inside `onBlock` for every replayed block (consensus state is identical), only the housekeeping is collapsed.

### 2. Split `onBlockFollowup` into two phases

| Phase | Function | Cost | Runs inline today? |
|---|---|---|---|
| A | `emitChainEvents` | cheap (SSE head/justification/finalization, metric gauges) | ✓ always inline (latency-sensitive) |
| B | `processFinalizationFollowup` | heavy (`processFinalizationAdvancement` + DB moves + prune state/forkchoice/cached-blocks/attestations) | ✓ today, but dispatchable |

Phase B only runs when finalization actually advanced. Snapshotting `previous_finalized` *before* `emitChainEvents` (which mutates `self.last_emitted_finalized`) preserves the diff that drives Phase B.

### 3. `runFollowupAsync` + `setFollowupDispatch` hook

```zig
pub const FollowupDispatchFn = *const fn (
    ctx: ?*anyopaque,
    chain: *BeamChain,
    previous_finalized: types.Checkpoint,
    pruneForkchoice: bool,
) void;

pub fn setFollowupDispatch(self: *Self, ctx: ?*anyopaque, func: ?FollowupDispatchFn) void;
```

Default: `null` → Phase B runs inline (current behavior).
When wired: the dispatcher (a single-consumer worker thread on `BeamNode`) is handed `(prev_finalized, pruneForkchoice)` and is responsible for invoking `chain.processFinalizationFollowup(...)` on its own thread, serializing follow-up calls, and coordinating with the BeamNode mutex (or whatever lock comes out of the broader review).

`processFinalizationFollowup` is therefore made `pub` so a future `BeamFollowupWorker` in `node.zig` can call it from off-thread.

## What's intentionally **not** here

- The actual worker thread + MPSC queue. Wiring a worker means thinking about thread-safety of `forkChoice` writes, `db.commit`, the `event_broadcaster`, and the `prune_cached_blocks_fn` callback in `BeamNode`. That deserves its own focused PR with explicit invariants documented; doing it here in a single-shot agent-driven PR would be risky. After this lands, that follow-up is a small wiring change in `node.zig`.
- The other items from the review (req/resp lock minimisation, parallel sig-verify ‖ pre-state clone in `onBlock`, parallel network-fetch + missed-root prune, centralised hash-root caching on gossip envelopes). Those each warrant their own PRs and largely don't fight with the #798–#801 stack.

## Verified

- `zig build` — green
- `zig build test` — green
- `zig fmt --check` — clean
- Diff is +150 / -31 in a single file (`pkgs/node/src/chain.zig`).

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -157,6 +157,13 @@ pub const BeamChain = struct {
     prune_cached_blocks_ctx: ?*anyopaque = null,
     prune_cached_blocks_fn: ?PruneCachedBlocksFn = null,
 
+    // Optional dispatcher for the heavy Phase-B follow-up work. When set,
+    // `runFollowupAsync` hands the call off to the dispatcher (e.g. a
+    // dedicated worker thread) instead of running it inline. See
+    // `runFollowupAsync` and `setFollowupDispatch`. Default: null (inline).
+    followup_dispatch_ctx: ?*anyopaque = null,
+    followup_dispatch_fn: ?FollowupDispatchFn = null,
+
     // Queue for blocks that arrived before forkchoice had ticked to their slot.
     // When a peer gossips a block for the current slot before our local interval
     // timer fires, the forkchoice rejects it with FutureSlot.  We hold such
@@ -164,6 +171,12 @@ pub const BeamChain = struct {
     pending_blocks: std.ArrayList(types.SignedBlock),
 
     pub const PruneCachedBlocksFn = *const fn (ptr: *anyopaque, finalized: types.Checkpoint) usize;
+    pub const FollowupDispatchFn = *const fn (
+        ctx: ?*anyopaque,
+        chain: *BeamChain,
+        previous_finalized: types.Checkpoint,
+        pruneForkchoice: bool,
+    ) void;
 
     const Self = @This();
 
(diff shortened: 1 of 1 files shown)


zclawz:

Issue #803, opened 2026-04-29
BeamNode threading model refactor — 8-point plan (supersedes #798-#802)


zclawz:
## Context

Refactor the BeamNode threading model to address the 8-point plan agreed in the 2026-04-29 mutex/parallelism review (Gajinder, Anshal, Partha):

> 1. All resources shared across threads maintain their own locks/mutexes (forkchoice is a resource).
> 2. Minimize locking — e.g. serving req/resp shouldn't need a lock except reading a forkchoice snapshot.
> 3. Only finalization advancement might need a lock across multiple resources because we do a canonicality view and then prune multiple resources.
> 4. All onBlock operations to be maximally parallel — e.g. sig verification can be parallel with pre-state getting cloned/ready.
> 5. Compute hash roots once for gossip objects and cache them there; reuse from there.
> 6. onBlock followups on a separate thread so onGossip can be freed up faster (followup is not critical).
> 7. Remove onBlockFollowup from any backfilling operations.
> 8. Network fetching and pruning of missed roots should be a parallel op.

This issue supersedes the BeamNode-mutex hoist stack #798 → #799 → #800 → #801 → #802. Those PRs were good devnet fire-fighting (good metrics, real wins) but they all work *around* the existing single `BeamNode.mutex`, while points 1+2 ask for a different shape: per-resource locks + lock-free read paths.

We are closing 798–802 in favor of slicing this refactor vertically, so each slice is independently reviewable and consensus-correct on its own.

## Devnet metrics carried forward from the closed stack

These came out of the #786 mutex instrumentation and motivate the work; preserved here so the diagnoses aren't lost.

### From #798 (gossip block import — `chain.onBlock` verify+STF)

| site | hold sum | hold mean | hold p99 (events >2s) |
|------|----------|-----------|------------------------|
| **onGossip** | **496.83s** | **49ms** | **122 events >2s** |
| onReqRespResponse | 71.85s | 23ms | 13 events >2s |
| onReqRespRequest.blocks_by_root | 8.98s | 20ms | 0 |
| onInterval | 4.83s | 2.6ms | 0 |

`onInterval` wait sum 13.75s (mean 7.4ms, p99 >100ms) — interval ticks delayed by gossip block imports holding the mutex. This drives the `slot_interval=1 duration=1.071s` pattern.

`zeam_chain_onblock_duration_seconds` mean 77ms (32s sum / 413 blocks).

### From #799 (gossip attestation verify)

| metric | sum | count | mean |
[...]

PR #804, opened 2026-04-29
docs: slice (a) threading refactor design doc

zclawz:
**Design doc only — no code changes yet.** Tracks #803 slice (a).

Posting the design before touching code because slice (a) is the riskiest of the five slices: it changes how every chain-mutation entry point synchronises against shared state, and a wrong lock hierarchy = consensus bug or deadlock on devnet. Burning a few hundred lines of design first is a much cheaper failure mode than discovering the wrong shape in a 2000-line review.

## What's in this PR

A single new doc — `docs/threading_refactor_slice_a.md` (~190 lines) — covering:

1. **Inventory** of every shared resource under `BeamNode`/`BeamChain`/`Network` and how it's protected today (almost all: `BeamNode.mutex` only).
2. **Threads in play** (libxev main, libp2p bridge, ThreadPool workers, future followup worker).
3. **Lock hierarchy** — strict ordering rule from `finalization_lock` (outermost) down to `forkChoice` (innermost).
4. **Resource-by-resource design** — lock type, callsites, edge cases for each of: `forkChoice` (already done), `states`, `pending_blocks`, `public_key_cache` + `root_to_slot_cache`, `last_emitted_*`, `Network`'s 7 maps, `batch_pending_parent_roots`.
5. **Lock-free req/resp** — explicit reasoning for why `onReqRespRequest` (`.status` and `.blocks_by_root`) can drop the mutex entirely.
6. **PR breakdown for slice (a)** — three small mergeable PRs: (a-1) infra, (a-2) chain, (a-3) node + req/resp.
7. **Explicit non-goals** — what slice (a) does NOT do (slices b–e stay separate).
8. **Open questions** — four things I want a human to weigh in on before I cut code.

## Asks for reviewers (cc @gr3999 @anshalshukla @ch4r10t33r)

- The **lock hierarchy** — does anyone see a code path that legitimately needs to violate the ordering?
- Is `LockedMap<K, V>` the right primitive vs hand-rolling per-map locks on `Network`?
- The **lock-free req/resp** claim — am I missing any state that gets touched on the request path?
- PR granularity — 3 PRs the right shape or should (a-2)/(a-3) be combined / split further?

Once we agree on the shape, PR (a-1) (infra-only, no callsite changes) is straightforward to cut and review.

## Next

Holding off on (a-1) until this design gets at least one human signoff. The other slices ((b)–(e)) of #803 are independent of this design discussion and can proceed in parallel if anyone wants to grab one.

### DIFF
--- a/docs/threading_refactor_slice_a.md
+++ b/docs/threading_refactor_slice_a.md
@@ -5,7 +5,7 @@ Tracking issue: #803
 Status: **DESIGN — REVISION 2 (review feedback applied)**
 Author: zclawz bot (under direction of @ch4r10t33r, @gr3999)
 
-Changelog vs. r1:
+Changelog vs. r1 (Partha #1–#7):
 - Added §State-pointer lifetime (Partha #1) — slice (a) ships `BorrowedState` wrapper; refcount deferred but the API contract is fixed now.
 - Added §Long-hold FFI paths (Partha #2) — `aggregate` and `getProposalAttestations` now snapshot-then-release explicitly.
 - Added §Single-writer claim retracted (Partha #3) — `events_lock` introduced for `last_emitted_*` and `cached_finalized_state`.
@@ -13,6 +13,16 @@ Changelog vs. r1:
 - Split `caches_lock` into `pubkey_cache_lock` + `root_to_slot_lock` (Partha #5).
 - Folded (a-1) into (a-2) (Partha #6); LockedMap unit tests still required.
 - Reworked `processPendingBlocks` to one-at-a-time `orderedRemove(0)` (Partha #7).
+
+Changelog vs. r2 (Partha #8–#13 + open-question responses):
+- Added §Lock-dance ownership note (Partha #8) — (a-2) absorbs the cognitive load that previously lived in `external_mutex`; LOC estimate revised to ~1000+, and the slice's review burden lives here.
+- Added §Cross-thread chain readers (Partha #9) — enumerates current API/metrics/event-broadcaster reads, and reserves the `/eth/v1/*` HTTP surface for an explicit follow-up section.
+- Added §Lock-hierarchy semantics clarification (Partha #10) — the rule is about *simultaneous* hold order, not all-time acquire order; sequential acquire/release of any lock is fine.
+- Reaffirmed metric migration plan (Partha #11) — emit both old and new metrics for one release; folded into (a-2) since it touches the same labels.
+- Added §Stress test plan (Partha #12) — (a-3) gets gossip-flood + RPC concurrency, 10-node devnet under jitter, reorg + finalization stress.
+- Added §`connected_peers` access pattern (Partha #13) — atomic counter for hot-path `count()`, `RwLock` for the few `iterator()` callers, mutex sized for adds/removes.
+- Resolved open questions: (1) refcount required before slice (c) goes off-thread; (2) `connected_peers` → atomic count + RwLock for iterator; (3) metric migration in (a-2); (4) drop `external_mutex` outright, no null-only transitional release.
+- Added §Long-term direction — single chain-mutator thread + queues vs. fine-grained locks. Slice (a) lock-hierarchy work survives either way; per-resource exclusive write locks become dead weight if mutation marshalls. Captured as a #803 question, not slice (a)'s decision.
 
 ## Why a design doc first
 
@@ -68,12 +78,13 @@ Only forkchoice has its own per-resource lock today. Everything else is "BeamNod
 6. BeamChain.forkChoice            (its own RwLock — innermost)
 ```
 
-Rules:
-- Higher number = inner lock. Acquire 1 → 2 → 3 → 4 → 5* → 6 only. Never reverse.
-- A holder of lock N can take lock M only if M > N.
-- The 5a/5b/5c locks are siblings: they sit at the same tier but **must not be held simultaneously** (they protect independent resources). Treat them as mutually exclusive within a single critical section to avoid deadlock-via-different-orderings.
+Rules (clarified per Partha #10):
+- **Scope:** the rule applies to locks held *simultaneously*. A code path that does `lock(forkChoice) ... unlock(forkChoice); lock(states) ... unlock(states)` sequentially is fine even though it appears to acquire 6 before 3 — they are never co-held.
+- A holder of lock N can additionally take lock M (i.e. nest M inside N's critical section) only if M > N. Never the reverse.
+- The 5a/5b/5c locks are siblings: they sit at the same tier but **must not be held simultaneously** (they protect independent resources). Treat them as mutually exclusive within a single nesting depth to avoid deadlock-via-different-orderings.
 - The vast majority of code paths take **at most one** of these. The hierarchy exists for the few paths that legitimately span multiple resources.
 - Finalization advancement is the only known multi-resource path that may legitimately need (1).
+- (`onBlock` legitimately touches multiple locks sequentially — forkchoice read for parent lookup, then states for STF commit, then forkchoice write for head update. This is sequential, not nested, and stays legal.)
 
 ### Resource-by-resource design
 
@@ -170,6 +181,21 @@ Fix: add `events_lock: std.Thread.Mutex` (lock 5c) covering `last_emitted_justif
 
 Alternative considered: route all event emission through a single-writer queue drained by a dedicated thread. Rejected for slice (a) — adds a thread before we need one. Revisit in slice (c) when the followup worker lands.
 
+#### Cross-thread chain readers (Partha #9)
+
+Existing and forward-looking surfaces that read chain/network state from a thread other than libxev/libp2p-bridge:
+
+| Surface | File | What it reads | Status today | After slice (a) |
+|---|---|---|---|---|
+| Prometheus `/metrics` writer | `pkgs/api/src/lib.zig` (HTTP worker thread when wired up) | metric values only — metric registry has its own internal sync | already lock-free | unchanged |
+| `event_broadcaster.zig` (SSE consumers) | `pkgs/api/src/event_broadcaster.zig` | broadcaster.subscribers + queued events | own `Mutex` | unchanged; receives events via `events_lock` writer pushing into the broadcaster |
+| `lean_connected_peers` metric set | `pkgs/node/src/node.zig:1174,1215` | called from `onPeerConnected` / `onPeerDisconnected` callbacks (libp2p bridge thread) | inside `BeamNode.mutex` | uses `connected_peers` lock only |
+| Peer broadcast iterator | `pkgs/node/src/node.zig:1389` | iterates `connected_peers` for outgoing req/resp | inside `BeamNode.mutex` | takes `connected_peers_lock.shared` (RwLock — see Partha #13) |
+
+**Reserved for a separate follow-up section before code:** the upcoming `/eth/v1/beacon/states/*`, `/eth/v1/beacon/headers`, `/eth/v1/beacon/blocks/{block_id}` HTTP endpoints. Those run on an HTTP worker thread independent of libxev/libp2p, and they read `chain.forkChoice`, `chain.states`, `chain.last_emitted_*`, and `db.loadBlock`. After slice (a) they MUST take per-resource locks (forkchoice shared / `BorrowedState` / `events_lock` / db handles its own sync). Today they don't exist; the doc reserves the contract here so it isn't discovered at runtime when they land.
+
+If any prototype HTTP route lives on a feature branch I'm not aware of, please flag it before (a-2) merges so the route migration lands together.
+
 #### `Network` maps
 Wrap independent maps in a small `LockedMap(K, V)` helper that bundles `std.Thread.Mutex` + the underlying map and exposes the few methods we actually use (`get`, `put`, `remove`, `count`, `iterator-while-locked`). This keeps callsite changes mechanical: `self.network.pending_rpc_requests.get(...)` becomes thread-safe by construction.
 
@@ -177,11 +203,16 @@ The maps that get **independent** locks (separate code paths, no shared invarian
 - `pending_rpc_requests`
 - `pending_block_roots`
 - `timed_out_requests`
-- `connected_peers`
+- `connected_peers` — see special handling below (Partha #13)
 
 **`block_cache_lock` — bundled (Partha #4).** `fetched_blocks`, `fetched_block_ssz`, and `fetched_block_children` share a lifecycle: when a block arrives from req/resp we cache the parsed block, the raw ssz bytes, and link its children atomically. With three independent locks a reader can observe an inconsistent slice (block present, ssz absent) — today this triple-update is atomic under `BeamNode.mutex` and code relies on it.
 
 Fix: a single `block_cache_lock: std.Thread.Mutex` guards all three maps together, exposed via a small `BlockCache` helper (`insert(block, ssz, parent)`, `get(root) -> ?CachedBlock`, `removeChildrenOf(root)`, etc.). The three underlying `HashMap`s are private; callers can only mutate via the helper, so the invariant is structural, not aspirational.
+
+**`connected_peers` access pattern (Partha #13).** `connected_peers.count()` is read from logger config on most gossip paths — frequent, hot. `connected_peers.iterator()` is read from peer broadcast (`node.zig:1389`) — less frequent, longer hold. Adds/removes happen on libp2p bridge callbacks. Plan:
+- Replace the `count`-only hot path with an `std.atomic.Value(usize)` (`connected_peer_count`) that is incremented/decremented atomically under the lock when entries are added/removed. Logger reads this atomic, never touches the lock.
+- Use `std.Thread.RwLock` for the map itself: `iterator()` callers take `lockShared`; `add` / `remove` take `lockExclusive` and update the atomic count alongside the map mutation.
+- Net: logger pays one atomic load instead of a mutex acquire per gossip log line; iterator readers run concurrently.
 
 #### `BeamNode.batch_pending_parent_roots`
 Same `LockedMap` helper. Single-resource lock.
@@ -191,6 +222,18 @@ Same `LockedMap` helper. Single-resource lock.
 Renamed → `BeamNode.finalization_lock`. Held by:
 - `processFinalizationFollowup` (and its dispatcher when slice c lands).
 - Anywhere we need a multi-resource view (today only finalization).
+
+#### Lock-dance ownership in `chain.zig` (Partha #8)
+
+The r1 doc described `external_mutex` removal as "mechanical." It isn't. Each of `onBlock`, `onGossipAttestation`, `onGossipAggregatedAttestation`, `produceBlock`, `processPendingBlocks` currently has a lock-dance shape that today is owned by `BeamNode` via `external_mutex`. After slice (a) that shape moves *into* `chain.zig`:
+
+- `states_lock.shared` is taken at the top to fetch the parent state via `BorrowedState`.
+- The borrow is converted to an owned snapshot (`sszClone`) for any work that crosses an unlock — verify, FFI, STF.
+- For STF commit, `states_lock.exclusive` is re-acquired at the *end* of the path to publish the new state and forkchoice update.
+
+The cognitive load ("release shared → do work → re-acquire exclusive → commit") is preserved, just owned by `chain.zig` instead of `BeamNode.zig`. Callers no longer have to know about it. That is the win — not less code, less spreading.
+
+**LOC reality check:** ~1000+ for (a-2), not the ~600 I estimated in r1. Most of it is per-callsite migration + tests. (a-2) carries the slice's whole review burden — plan reviewer time accordingly.
 
 NOT held by:
 - `onGossip` — uses per-resource locks now.
@@ -269,32 +312,42 @@ Listed explicitly to keep PR scope tight:
 
 Revised to **two PRs** per Partha #6 — folding (a-1) into (a-2) so reviewers can evaluate the new primitives against real callsites in one pass instead of trying to spot init/deinit ordering bugs in isolation. The cost (a slightly bigger (a-2)) is offset by mandatory unit tests on the new primitives.
 
-1. **`(a-2) chain + primitives`** — adds the `LockedMap` and `BlockCache` helpers, adds `BorrowedState`, adds `states_lock`, `pending_blocks_lock`, `pubkey_cache_lock`, `root_to_slot_lock`, `events_lock`. Migrates every `chain.zig` callsite (states.get → BorrowedState, pending_blocks → new lock, caches → split locks, events → events_lock). Updates `chain.onBlock` / `chain.onGossip` / `chain.processPendingBlocks` to no longer require an `external_mutex` parameter. Drops the `external_mutex` parameter (was added by #798–#801, now obsolete). Implements the snapshot-then-release pattern in `chain.aggregate` and `produceBlock` so `forkChoice.aggregate` / `getProposalAttestations` see an owned snapshot, not a borrow.
+1. **`(a-2) chain + primitives`** — adds the `LockedMap` and `BlockCache` helpers, adds `BorrowedState`, adds `states_lock`, `pending_blocks_lock`, `pubkey_cache_lock`, `root_to_slot_lock`, `events_lock`. Migrates every `chain.zig` callsite (states.get → BorrowedState, pending_blocks → new lock, caches → split locks, events → events_lock). Updates `chain.onBlock` / `chain.onGossip` / `chain.processPendingBlocks` to no longer require an `external_mutex` parameter. Drops the `external_mutex` parameter (was added by #798–#801, now obsolete — dropped outright, no null-only transitional release). Implements the snapshot-then-release pattern in `chain.aggregate` and `produceBlock` so `forkChoice.aggregate` / `getProposalAttestations` see an owned snapshot, not a borrow. Adds the per-lock metric histograms (`zeam_lock_wait_seconds{lock="...", site="..."}`) and keeps the old `zeam_node_mutex_*` series alive as a derived sum for one release (compat shim) so dashboards don't go dark (Partha #11).
    - **Mandatory unit tests** (per Partha #6): `LockedMap` (constructor, get/put/remove, iterator-while-locked, deinit-when-empty, deinit-when-non-empty), `BlockCache` (atomic triple-insert, partial-state invariants), `BorrowedState` (one-release assertion, sszClone-then-deinit). These are the only standalone tests in this slice; everything else is covered by chain integration tests.
-   - ~800 LOC. Real semantic change; this is the one to review carefully.
+   - **Realistic LOC: ~1000+** (revised up from r1's 600 per Partha #8 — the lock-dance moves into chain.zig, not away). This PR carries the slice's full review burden.
 
-2. **`(a-3) node + req/resp`** — migrate `Network` map accesses to the new locks (`block_cache_lock` plus the four independent ones). Drop `BeamNode.mutex` from `onGossip`, `onInterval`, `onReqRespResponse`. Make `onReqRespRequest` fully lock-free. Rename `BeamNode.mutex` → `finalization_lock` for the few remaining multi-resource paths. ~400 LOC.
+2. **`(a-3) node + req/resp`** — migrate `Network` map accesses to the new locks (`block_cache_lock` plus the four independent ones, with `connected_peers` getting the atomic count + RwLock pattern). Drop `BeamNode.mutex` from `onGossip`, `onInterval`, `onReqRespResponse`. Make `onReqRespRequest` fully lock-free. Rename `BeamNode.mutex` → `finalization_lock` for the few remaining multi-resource paths. ~400 LOC.
+   - **Stress test plan** (per Partha #12) — devnet smoke alone catches obvious deadlocks but misses UAFs and concurrency races. (a-3) ships at minimum:
+     1. **Single-node ingestion stress.** Synthetic gossip-block flood + concurrent `blocks_by_root` RPC against the same node. Run 30+ minutes; assert no `state-map-key-not-found` panics, no assertion failures, no `MissingPreState`.
+     2. **10-node devnet under jitter.** Existing devnet runner + tc-netem packet loss/delay for ≥1h. Watch for divergence, deadlock, or growing pending_blocks queue.
+     3. **Reorg + finalization stress.** Constructed scenario where two competing chain branches force a reorg right around a finalization advance — exercises the `events_lock` / `finalization_lock` boundary. (Partha right that this is rare on current devnet; need a synthetic harness.)
+   - At least one of these gates merge; ideally all three are wired into nightly so the slice keeps paying off in regression catches.
 
 Each PR builds + tests cleanly on its own; (a-2) and (a-3) get devnet smoke runs against the existing instrumentation from #786 to confirm no contention regression.
 
-## Open questions for review
+## Resolved open questions (Partha r2 responses)
 
-1. **`states` map prune coordination.** Today prune runs under the global mutex, so no reader can hold a stale `*BeamState`. After slice (a), prune runs under `states_lock` exclusive — fine — but the followup-dispatch path (slice c) might want to read a state from a worker thread that started before prune ran. Slice (a) handles this by keeping followup inline (slice c hasn't landed yet); revisit when slice c lands.
+1. **`states` map prune coordination.** Slice (a) handles this by keeping followup inline. **However:** if slice (c) moves the followup off-thread without first adding refcounted state pointers, it will tear. **Hard requirement: slice (c) MUST land an `Arc<BeamState>` (or equivalent refcount) before going off-thread.** Captured as a slice-(c) blocker in #803.
 
-2. **`connected_peers` lock granularity.** It's read on every gossip block (logger), so a coarse mutex on it could become a contention point itself. Consider `RwLock` here? The reads are pure and short, so probably fine with `Mutex`, but worth measuring.
+2. **`connected_peers` lock granularity.** Resolved — atomic counter for the hot `count()` path + `RwLock` for the iterator path. See §Network maps above.
 
-3. **Lock metric coverage.** The current `zeam_node_mutex_{wait,hold}_time_seconds` instrumentation is keyed by site label and assumes a single mutex. Slice (a) should replace this with per-lock histograms (`zeam_lock_wait_seconds{lock="states", site="..."}`). Lean toward folding into (a-2) since the metric label set changes anyway.
+3. **Lock metric coverage.** Resolved — fold into (a-2) since the metric label set changes anyway. Emit both old (`zeam_node_mutex_*`) and new (`zeam_lock_wait_seconds{lock=...}`) for one release; old metric becomes the sum across new lock labels via a recording rule / derived shim. Drop the old series in the release after.
 
-4. **`external_mutex` removal vs. backward-compat.** Once (a-2) lands, the `external_mutex: ?*std.Thread.Mutex` parameter on `onBlock`/`processPendingBlocks` is dead weight. Drop in (a-2), or keep as `null`-only for one release in case external embedders depend on it? (No known external embedders today, so probably just drop.)
+4. **`external_mutex` removal vs. backward-compat.** Resolved — drop outright in (a-2). No `null`-only transitional release. (No external embedders today; the param was internal-only plumbing from #798–#801.)
 
-## Ask for reviewers (r2)
+## Long-term direction note (Partha post-script)
 
-Please weigh in on:
+This refactor preserves the shape "every chain mutator can be called from any thread, synchronised via per-resource locks." An alternative long-term shape is "all chain mutation marshalled to a single chain-mutator thread with a queue, readers read snapshots." Slice (a)'s lock-hierarchy work survives the marshalling refactor (still useful for *read* paths and for any per-resource cache the mutator thread owns), but per-resource *exclusive write* locks become dead weight under marshalling.
 
-- The revised lock hierarchy with sibling 5a/5b/5c — anyone see a path that legitimately needs two of those at once?
-- `BorrowedState` API shape — do we want the typed wrapper now or skip straight to refcounted `Arc<BeamState>`? (My read: wrapper now, refcount only if slice c forces the issue.)
-- `BlockCache` helper — is bundling the three maps the right call, or is it cleaner to keep them as separate `LockedMap`s and document the readers-may-see-partial invariant?
-- The snapshot-then-release pattern in `aggregate` / `produceBlock` — OK with full `sszClone`, or do we need an `AggregatorView` partial-copy from day one for cost reasons?
-- The 2-PR breakdown after folding (a-1) into (a-2) — still tractable for review, or should I split (a-2) along chain-vs-primitives lines instead?
+Which direction is #803's long-term target should be a #803-level decision before slice (b)/(c) commits to a specific shape; flagging here so each slice converges in the same direction. My read of the original 8-point plan: the marshalling shape is the cleaner end state, and slice (a) is a strict prerequisite either way (read snapshots need lock-hierarchy regardless). Worth a one-paragraph statement of intent on #803.
 
-Once we agree, I’ll cut PR (a-2) with the LockedMap/BlockCache/BorrowedState primitives + chain migration in one go.
+## Ask for reviewers (r3)
+
+Most of r1/r2's open questions are now closed. Remaining decisions before I cut code:
+
+- **Long-term direction in #803.** Marshalled chain-mutator vs fine-grained per-resource locks. Slice (a) is compatible with either, but slice (b)+(c) need this pinned. Looking for a one-paragraph statement of intent in #803.
+- **`AggregatorView` partial-copy vs full `sszClone`** in `aggregate` / `produceBlock`. Default is full clone; only worth the partial copy if benchmarks show ssz-clone time dominates the FFI. Defer to (a-2) profiling unless someone has data already.
+- **HTTP `/eth/v1/*` surface.** If a prototype branch already exists, please flag before (a-2) merges so the route migration lands together rather than as a follow-up.
+- **Stress test prioritisation.** All three scenarios in (a-3) are useful but only one is required as a merge gate. My pick: **single-node ingestion stress** (cheapest to wire up, catches the largest UAF surface). Open to a different pick.
+
+Once these are resolved (or explicitly punted), I’ll cut PR (a-2) with the LockedMap/BlockCache/BorrowedState primitives + chain migration in one go.
--- a/docs/threading_refactor_slice_a.md
+++ b/docs/threading_refactor_slice_a.md
@@ -1,0 +1,189 @@
+# Threading Refactor — Slice (a): Per-Resource Locks + Lock-Free Req/Resp
+
+Date: 2026-04-29
+Tracking issue: #803
+Status: **DESIGN — REVIEW BEFORE CODE**
+Author: zclawz bot (under direction of @ch4r10t33r, @gr3999)
+
+## Why a design doc first
+
+Slice (a) of #803 is the riskiest of the five slices: it changes how every chain-mutation entry point synchronises against shared state, and a wrong lock hierarchy here = consensus bug or deadlock on devnet. Burning a few hundred lines of design before touching code is a much cheaper failure mode than discovering the wrong shape in a 2000-line PR review.
+
+Once this doc is reviewed and the lock hierarchy + invariants are agreed, the actual code change becomes mostly mechanical and can land in 2–3 small PRs.
+
+## Current state (as of `main` @ commit `dacc1c2`)
+
+Single coarse `BeamNode.mutex` (`std.Thread.Mutex`) serializes all libxev-thread vs libp2p-bridge-thread access to **everything** under `BeamNode`:
+
+| Resource | Owner | Today’s synchronisation |
+|---|---|---|
+| `BeamChain.forkChoice` | chain | `RwLock` ✅ already per-resource |
+| `BeamChain.states` (HashMap<Root, *BeamState>) | chain | only `BeamNode.mutex` |
+| `BeamChain.pending_blocks` (ArrayList<SignedBlock>) | chain | only `BeamNode.mutex` |
+| `BeamChain.public_key_cache` | chain | only `BeamNode.mutex` (documented not-thread-safe internally) |
+| `BeamChain.root_to_slot_cache` | chain | only `BeamNode.mutex` |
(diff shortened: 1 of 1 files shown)


zclawz:
**Working on this** — picking up Partha's 7 review comments (4 High, 3 Medium).

Plan:
- **#1 (BeamState UAF):** decide between Arc-style refcount vs `BorrowedState` wrapper for slice (a); also enumerate every current `states.get` call site inline in the doc as the floor.
- **#2 (aggregate / getProposalAttestations long-hold):** add explicit snapshot-then-release pattern (either `sszClone` or copy the few fields the FFI aggregator reads) — currently missing from the lock-dance section.
- **#3 (event-emission single-writer claim is wrong):** `chain.onBlockFollowup` is reachable from both libp2p bridge and libxev main thread → introduce `events_lock` or single-writer queue for `last_emitted_*` / `cached_finalized_state`.
- **#4 (block cache multi-map atomicity):** bundle `fetched_blocks` + `fetched_block_ssz` + `fetched_block_children` under one `block_cache_lock`, since they share a lifecycle.
- **#5 (caches_lock perf trap):** split `public_key_cache` (FFI miss latency) from `root_to_slot_cache` (per-attestation hot path).
- **#6 (a-1 testability):** add LockedMap unit tests for the standalone (a-1) sub-slice covering ctor / get/put/remove / iterator-while-locked / deinit-empty + non-empty.
- **#7 (processPendingBlocks):** replace snapshot-indices with one-at-a-time `orderedRemove(0)` to avoid index drift under concurrent appends.

Will push design-doc revisions first (single commit), then reply inline on each thread. Flagging any I think need a deeper conversation rather than a doc edit.

zclawz:
**Design doc revised** in `ecad494`. Summary of how each of @ch4r10t33r's 7 points landed (full diff: https://github.com/blockblaz/zeam/pull/804/files):

### High

**#1 — `*BeamState` UAF / unenforced invariant.**
Agreed the API contract was wrong. Slice (a) now ships a typed `BorrowedState` wrapper:

```zig
pub const BorrowedState = struct {
    states_lock: *std.Thread.RwLock,
    pub fn deinit(self: *BorrowedState) void { self.states_lock.unlockShared(); }
    pub fn sszClone(self: *BorrowedState, allocator: Allocator) !*types.BeamState { ... self.deinit(); ... }
};
```

Every `states.get` returns `?BorrowedState`, not a raw pointer. The lock is held for the borrow's lifetime, debug-builds assert one-release-per-borrow. Refcount/Arc deferred — only revisit if slice (b) or (c) actually needs reader-outlives-prune semantics.

**Inline inventory of the 9 current `states.get` callsites** added to the doc (chain.zig:458, 919, 1575, 1626, 1654, 1716, plus node.zig:1477 and two test sites). Each row spells out the lifetime requirement (borrow-only vs sszClone-required).

**#2 — `forkchoice.aggregate` / `getProposalAttestations` long-hold.**
Right, in r1 the contention just shifted from `BeamNode.mutex` to `states_lock` for the ~700ms FFI window. Doc now mandates snapshot-then-release at both sites (`chain.aggregate`, `produceBlock`):

```zig
var borrow = self.states.get(head_root) orelse return error.MissingState;
const snapshot = try borrow.sszClone(self.allocator); // releases states_lock.shared
defer snapshot.deinitAndDestroy(self.allocator);
return self.forkChoice.aggregate(snapshot);
```

Open question (flagged in doc): full `sszClone` vs an `AggregatorView` partial-copy of just the fields the aggregator reads. Defaulting to full clone unless cost forces the partial.

**#3 — single-writer claim retracted.**
You're right — `chain.onBlockFollowup` is reachable from libp2p bridge thread (`chain.zig:771`, gossip block import) AND libxev main thread (`chain.zig:322` via `processPendingBlocks`, plus `node.zig:583/854/1496`). The "single-writer" comment in r1 was wrong. Slice (a) now adds `events_lock: std.Thread.Mutex` (lock 5c in the hierarchy) covering `last_emitted_justified`, `last_emitted_finalized`, and `cached_finalized_state`. Single-writer queue rejected for slice (a) — adds a thread before we need one; revisit in slice (c) when the followup worker lands.

**#4 — multi-map atomicity.**
Bundled. New `block_cache_lock` (single mutex) guards `fetched_blocks` + `fetched_block_ssz` + `fetched_block_children` together via a `BlockCache` helper that exposes `insert(block, ssz, parent)` / `get(root) -> ?CachedBlock` / `removeChildrenOf(root)`. The three underlying `HashMap`s become private — atomicity is structural, not aspirational.

### Medium

**#5 — `caches_lock` perf trap.**
Split. `pubkey_cache_lock` (5a) is its own mutex so an XMSS FFI miss can't stall `root_to_slot_lock` (5b), which is on the per-attestation hot path. Marked as sibling locks in the hierarchy (must not be held simultaneously).

**#6 — (a-1) standalone has no live callsites.**
Folded (a-1) into (a-2). Slice (a) is now two PRs: `(a-2) chain + primitives` (LockedMap/BlockCache/BorrowedState + every chain.zig migration + mandatory unit tests for the new primitives) and `(a-3) node + req/resp`. Unit tests called out explicitly: `LockedMap` (ctor / get/put/remove / iterator-while-locked / deinit-empty / deinit-non-empty), `BlockCache` (atomic triple-insert, partial-state invariants), `BorrowedState` (one-release assertion, sszClone-then-deinit).

**#7 — `processPendingBlocks` snapshot-indices is bug-prone.**
Reworked to one-at-a-time `orderedRemove`, no index snapshot:

```zig
while (true) {
    var ready: ?types.SignedBlock = null;
    {
        self.pending_blocks_lock.lock();
        defer self.pending_blocks_lock.unlock();
        for (self.pending_blocks.items, 0..) |b, i| {
            if (b.message.slot <= current_slot) {
                ready = self.pending_blocks.orderedRemove(i);
                break;
            }
        }
    }
    if (ready) |b| { self.onBlock(b, ...) catch |e| { ... }; }
    else break;
}
```

Re-scans from the front on each iteration → indices are never assumed stable across an unlock.

---

**Open questions for r2 (also in doc):**
1. `BorrowedState` wrapper now vs straight to refcounted `Arc<BeamState>`?
2. `BlockCache` bundling vs separate `LockedMap`s with documented partial-state invariant?
3. Snapshot pattern: full `sszClone` vs partial `AggregatorView`?
4. 2-PR breakdown still tractable, or split (a-2) further along chain-vs-primitives lines?

Once the inline comments on the doc are submitted (currently still in pending review state — I can't read them via API yet, only the chat summary), I'll respond inline on each thread and cut PR (a-2).

zclawz:
**r3 design doc pushed** in `467d3ef` — addresses Partha's points #8–#13 + the four open-question responses.

### High-impact changes from r2

**#8 — `external_mutex` removal isn't mechanical.** Conceded. The lock-dance shape (release shared → verify+STF → re-acquire exclusive → commit) doesn't disappear; it migrates from `BeamNode.zig` into `chain.zig`. Each of `onBlock` / `onGossipAttestation` / `onGossipAggregatedAttestation` / `produceBlock` / `processPendingBlocks` owns its own dance. The win is encapsulation, not less code. **(a-2) LOC revised up to ~1000+**, and explicitly carries the slice's full review burden.

**#9 — Cross-thread chain readers.** Added an inventory section with a table covering: Prometheus `/metrics` writer, `event_broadcaster.zig` SSE consumers, `lean_connected_peers` metric set, and the peer broadcast iterator (`node.zig:1389`). Reserved a placeholder for the upcoming `/eth/v1/beacon/states/*`, `/eth/v1/beacon/headers`, `/eth/v1/beacon/blocks/{block_id}` HTTP routes — they MUST take per-resource locks (forkchoice shared / `BorrowedState` / `events_lock` / db-internal sync) when they land. Asked in the reviewer section if any prototype HTTP surface lives on a feature branch; if so I'll fold the route migration into (a-2).

**#10 — Lock-hierarchy semantics.** Reworded:
- Rule applies to **simultaneous** holds, not all-time acquire order.
- Sequential acquire/release of any lock is fine: `lock(forkChoice) ... unlock; lock(states) ... unlock` is legal even though forkchoice (lock 6) was acquired before states (lock 3).
- `onBlock` already does forkchoice → states → forkchoice sequentially today — stays legal.

**#11 — Metric migration.** Folded into (a-2). Emit both `zeam_node_mutex_*` (derived sum across new lock labels) and `zeam_lock_wait_seconds{lock=...,site=...}` for one release. Drop the old series in the release after. Dashboards stay alive across the cutover.

**#12 — Stress test plan.** (a-3) ships with three scenarios:
1. **Single-node ingestion stress** — synthetic gossip flood + concurrent `blocks_by_root` RPC, 30+ min, assert no `state-map-key-not-found` panics or `MissingPreState`.
2. **10-node devnet under jitter** — existing devnet runner + `tc-netem`, ≥1h, watch for divergence / deadlock / pending_blocks growth.
3. **Reorg + finalization stress** — synthetic harness that forces a reorg right around finalization advance — exercises the `events_lock` / `finalization_lock` boundary.

At least one is a merge gate (my pick: scenario 1, cheapest + largest UAF surface); ideally all three wired into nightly.

**#13 — `connected_peers` access pattern.**
- `count()` is hot-path (logger config on most gossip paths) → replace with `std.atomic.Value(usize)` (`connected_peer_count`), updated under the lock at add/remove time. Logger reads the atomic, never touches the lock.
- `iterator()` (peer broadcast, `node.zig:1389`) is less frequent / longer hold → use `std.Thread.RwLock`, callers take `lockShared`.
- Adds/removes take `lockExclusive` and update the atomic alongside the map mutation.

### Open-question responses (Partha r2)

| Q | Resolution |
|---|---|
| 1. states-prune ↔ followup worker (slice c) | **Hard requirement: slice (c) MUST land refcounted state pointers (`Arc<BeamState>` or equivalent) before going off-thread.** Captured as a slice-(c) blocker in #803. |
| 2. `connected_peers` Mutex vs RwLock | Atomic counter + RwLock (above). |
| 3. Lock metric coverage | Fold into (a-2). Emit both old and new for one release. |
| 4. `external_mutex` backward-compat | Drop outright in (a-2). No `null`-only transitional release. |

### Long-term direction note

Slice (a) preserves "every chain mutator can be called from any thread, synchronised via per-resource locks." Alternative end state: "all chain mutation marshalled to a single chain-mutator thread with a queue, readers read snapshots." Slice (a)'s lock-hierarchy work survives either way (still useful for read paths and any per-resource cache the mutator owns); per-resource exclusive *write* locks become dead weight under marshalling. Flagged as a #803-level decision before slice (b)/(c) commits to a shape — I'll open a separate comment on #803 asking for a one-paragraph statement of intent.

### Remaining asks before I cut (a-2)

- One-paragraph long-term-direction statement on #803 (marshalled vs fine-grained).
- Confirm or override stress-test merge-gate pick (scenario 1 by default).
- `AggregatorView` partial-copy vs full `sszClone` — default to full unless someone already has profiling data.
- Any prototype `/eth/v1/*` HTTP surface I should fold into (a-2)?

Full revised doc: https://github.com/blockblaz/zeam/blob/feat/agent-slice-a-per-resource-locks/docs/threading_refactor_slice_a.md
Diff (this commit): https://github.com/blockblaz/zeam/pull/804/commits/467d3ef

ch4r10t33r:
r3 verified. Walked the doc end-to-end against the changelog claims.

**Verified absorbed:**

| # | Where in doc |
|---|--------------|
| #1 BorrowedState | L94-136 |
| #2 Snapshot-then-release for aggregate / getProposalAttestations | L244-264 |
| #3 events_lock for last_emitted_*  + cached_finalized_state | L173-182 |
| #4 block_cache_lock bundling 3 maps | L208-210 |
| #5 pubkey_cache_lock + root_to_slot_lock split | L163-171 |
| #6 (a-1) folded into (a-2) + LockedMap/BlockCache/BorrowedState unit tests mandatory | L313-316 |
| #7 processPendingBlocks one-at-a-time orderedRemove(0) | L142-161 |
| #8 LOC reality check + lock-dance ownership moves into chain.zig | L226-236, L317 |
| #9 Cross-thread chain readers + reserved /eth/v1/* contract | L184-197 |
| #10 Hierarchy clarified to "simultaneous holds" | L81-87 |
| #11 Both-metrics-for-one-release shim | L315, L334 |
| #12 Three stress scenarios in (a-3) | L320-324 |
| #13 connected_peers atomic count + RwLock | L212-215 |
| Long-term direction note | L338-342 |

All four open-question responses pinned at L330-336. Good.

**New small things I'd tighten before (a-2):**

1. **`BorrowedState.sszClone` consumes the borrow — name it accordingly.** `sszClone` reads as a non-mutating snapshot; the actual semantics are "consume this borrow, give me an owned copy, release the lock." Suggest `intoOwnedAndRelease` or `cloneAndRelease`. Otherwise reviewers (and future me) will write `borrow.sszClone(); ...; borrow.deinit();` and double-release.

2. **`BorrowedState` partial-failure path.** If `sszClone` itself errors (allocator OOM mid-clone), the borrow's lock must still be released. The L111 sketch handles success; spell out the errdefer for the failure case so the impl doesn't drop it. Single line in the doc.

3. **`BorrowedState` debug-assert "exactly one release"** (L118) needs a sentinel field shown in the snippet. Otherwise it's a vague claim. `released: bool` + `if (!released) @panic("BorrowedState dropped without release")` in deinit, with `released = true` after `mutex.unlockShared()`. Mirrors the `MutexGuard.released` pattern from #787 — same shape, same reason.

4. **5a/5b/5c "siblings, never co-held"** (L84) — that's a design constraint, not a property. (a-2) should have a runtime assert in debug builds (a small TLS counter "depth-tier-5" with `assert(== 0)` on entry) so a future contributor that violates it fails loudly in tests, not in production.

5. **Metric compat shim (L334) — pick one.** "recording rule / derived shim" mixes two things. **Recording rule** = ops change `prometheus.yml`. **Derived shim** = code double-emits the old metric internally. Pick code-side derived: every `zeam_lock_wait_seconds.observe(...)` also feeds the legacy `zeam_node_mutex_wait_time_seconds` derived from `lock∈{states,...}` summed. Operators don't have to redeploy Prometheus. One sentence in (a-2) doc-comment on the compat shim.

6. **`getFinalizedState` migration scope** (L131). "Today this is an unsafe escape hatch" — yes. (a-2) PR description should grep+list every `getFinalizedState` caller (not just `states.get` callers; this returns the ptr by a different path). One forgotten caller = UAF in production. ~5 minutes of grep work, big risk reduction.

7. **`pending_blocks_lock` algorithm is O(n²)** in the worst case (huge backlog after a long clock catch-up). For typical n<10 totally fine, but `MAX_PENDING_BLOCKS` is unbounded today. If it grows pathologically (devnet partitioned for hours), this iteration could turn quadratic. Either cap pending_blocks length or document the assumption that n stays small.

8. **`block_cache_lock` `removeChildrenOf(root)`** can iterate up to `MAX_CACHED_BLOCKS = 1024` entries while holding the lock. Bounded but ms-scale on the gossip thread. Worth a note that this is the upper bound of the longest critical section under that lock.

**Responses to your 4 remaining asks:**

1. **Long-term direction call.** My read aligns with yours: marshalling chain-mutator thread + queues is the cleaner end-state. Slice (a) is a prerequisite either way (read paths still need the hierarchy). I'll add a comment on #803 asking @gr3999 / @ch4r10t33r for the one-paragraph statement of intent. Don't gate (a-2) on it — the lock-hierarchy work is a no-regret prereq.

2. **Stress merge-gate pick — concur with single-node ingestion stress.** Cheapest, biggest UAF surface. Wire #2 (10-node devnet under jitter) and #3 (reorg+finalization) into nightly so they keep paying off post-merge.

3. **AggregatorView partial vs full sszClone — default full.** Agree. Profile (a-2) on devnet first; if `lean_block_building_time_seconds` shows clone time dominates, cut over to `AggregatorView`. Don't pre-optimise.

4. **/eth/v1/* HTTP prototype.** Not aware of a feature branch. @gr3999 / @ch4r10t33r — anyone working on this? If yes, please flag before (a-2) merges; if no, the L195 reservation is sufficient.

Beyond the 8 polish items above, design is mergeable. (a-2) is the carrier; review burden lives there, not here. Cut it.

zclawz:
**r3 polish addressed** in `a76c274`. All 8 tightening items + the 4 ask responses absorbed:

| # | Resolution |
|---|---|
| 1. `sszClone` misleading name | Renamed `BorrowedState.cloneAndRelease`. Consume-the-borrow semantics are now in the name. Caller does NOT call `deinit()` after — ownership of the lock has moved out. |
| 2. OOM-mid-clone partial-failure path | `cloneAndRelease` body is `errdefer self.deinit()` then `try self.state.sszClone(...)` then explicit `deinit()` on success. Lock always released, success or failure. Spelled out in the doc snippet. |
| 3. `released: bool` sentinel | Added explicitly in the doc snippet. `deinit` is idempotent (`if (self.released) return;`); a debug `@panic("BorrowedState dropped without release")` catches the drop-without-release case via the parent struct's deinit assertion. Mirrors `MutexGuard.released` from #787. |
| 4. Sibling-locks runtime assert | Added: debug-build thread-local depth counter for tier 5; `debug.assert(depth_tier5 == 0)` at the top of every 5a/5b/5c lock-acquire. Future contributors fail loudly in tests rather than silently in production. |
| 5. Metric compat shim ambiguity | Clarified to **code-side derived double-emit**: every `zeam_lock_wait_seconds.observe(...)` also feeds the legacy `zeam_node_mutex_*` summed across `lock∈{...}`. Operators do NOT redeploy Prometheus or change recording rules. One sentence in (a-2)'s metrics doc-comment. |
| 6. `getFinalizedState` caller enumeration | Added explicit note: (a-2) PR description must list every `getFinalizedState` caller (HTTP/RPC, validator, tests). One forgotten caller = production UAF. The function migrates to return `?BorrowedState`; callers either borrow-only or `cloneAndRelease`. |
| 7. `processPendingBlocks` O(n²) worst case | Acknowledged. Slice (a) ships the simple O(n) re-scan form + a `lean_pending_blocks_drain_iters` histogram so we can measure. Mitigations on the shelf if needed: `MAX_PENDING_BLOCKS` cap or a leftmost-ready cursor (gossip-append at tail can't invalidate it). Documented assumption: n stays small in normal operation. |
| 8. `removeChildrenOf` critical-section ceiling | Documented: worst-case iterates `MAX_CACHED_BLOCKS = 1024` while holding `block_cache_lock`. Bounded but ms-scale on the gossip thread — longest critical section under that lock. Captured for the next perf review. |

### Final r3 asks resolved

1. **Long-term direction call** — acknowledged as #803-level, does NOT gate (a-2). Will open a focused comment on #803 asking @gr3999 / @ch4r10t33r for the one-paragraph statement of intent on marshalled-mutator vs fine-grained.
2. **Stress merge gate** — confirmed: single-node ingestion stress. Scenarios 2 (10-node devnet under jitter) and 3 (reorg+finalization) wired into nightly post-merge.
3. **`AggregatorView` partial vs full** — default full `cloneAndRelease`. Profile in (a-2); cut over only if `lean_block_building_time_seconds` shows clone time dominates.
4. **`/eth/v1/*` HTTP prototype** — noted "not aware of a feature branch." If @gr3999 sees one, please flag now; otherwise the L195 reservation is sufficient.

### Next

Cutting PR (a-2): LockedMap + BlockCache + BorrowedState primitives + every chain.zig migration + per-lock metrics + legacy-metric derived shim + mandatory unit tests + `getFinalizedState` caller enumeration in PR description. Will keep this PR (#804) open as the design tracker until (a-2) merges, then close it referencing the implementation PRs.

Doc: https://github.com/blockblaz/zeam/blob/feat/agent-slice-a-per-resource-locks/docs/threading_refactor_slice_a.md
This commit: https://github.com/blockblaz/zeam/pull/804/commits/a76c274

zclawz:
Slice (a-2) draft PR is up: https://github.com/blockblaz/zeam/pull/805

**Status:** 🟢 build green, all 158/158 tests passing on `zig build --summary all test -Doptimize=Debug`. Pushed as **draft** for human review — please don't hit ready-for-review on my behalf.

**Highlights:**
- New primitives in `pkgs/node/src/locking.zig`: `LockedMap`, `BlockCache`, `BorrowedState`, `LockTimer`, tier-5 sibling depth counter.
- New `BeamChain` lock fields: `states_lock` (RwLock), `pending_blocks_lock`, `pubkey_cache_lock`, `root_to_slot_lock`, `events_lock`.
- Every `chain.zig` `states.{get,put,fetchRemove}` callsite migrated; full inventory + lifetime classification (borrow-only vs `cloneAndRelease`) is in the PR body.
- `getFinalizedState` now returns `?BorrowedState`; only caller (`api_server.zig:handleFinalizedCheckpointState`) migrated.
- `processPendingBlocks` rewritten to one-at-a-time `orderedRemove(0)` per Partha #7. Histogram `lean_pending_blocks_drain_iters` ships alongside.
- `events_lock` covers `last_emitted_{justified,finalized}` + `cached_finalized_state` read-modify-write per Partha #3.
- Per-lock metric histograms (`zeam_lock_{wait,hold}_seconds`) plus a code-side derived shim into the legacy `zeam_node_mutex_*` series so dashboards don't go dark.
- Debug-build TLS depth counter enforces the 5a/5b/5c sibling-not-co-held rule per Partha r3 #4.

**Spec link:** design doc at `docs/threading_refactor_slice_a.md` @ commit a76c274.

**Two minor deviations** from the doc, both noted in the PR body:
1. `statesCommitKeepExisting` (keep-existing on duplicate insert) instead of unconditional `states.put` — preserves in-flight borrow validity. Equivalent invariant.
2. `publishBlock` no longer passes a cached `postState` pointer — recompute is the only path now since the borrow + exclusive-commit dance would otherwise deadlock.

There is no `external_mutex` parameter to drop on `chain.onBlock` because the #798–#801 plumbing was never landed on `main`; that part of the design doc is a no-op against the current tree.

Slice (a-3) follow-ups (Network maps + `BlockCache` wiring + dropping `BeamNode.mutex` from gossip/interval/req-resp + lock-free `onReqRespRequest` + stress tests) will be a separate PR after this one.

GrapeBaBa:
@zclawz what thread is the `onGossip` similar callbacks running on? If it is running on libp2p thread, why don't schedule them on other thread?

zclawz:
Great question @GrapeBaBa!

**Current threading model for `onGossip`:**

`onGossip` (and the aggregated-attestation equivalent) is called via FFI from the Rust libp2p bridge. The call chain is:

```
Tokio event loop (Rust thread)
  → libp2p GossipSub event handler
    → zeam FFI callback (beam_node_on_gossip / ...)
      → Zig: BeamNode.onGossip()
```

So yes — it runs on the libp2p Tokio thread, blocking that thread for the duration of XMSS verify + forkchoice update.

**Why not just schedule it on another thread from the Rust side?**

You can, but it shifts the problem rather than solving it. If we push the FFI call off the Tokio thread into a Zig thread pool, we still need to acquire `BeamNode.mutex` on that worker thread — and any other thread doing `onBlock` or `onGossip` will contend on the same mutex. The real gain comes from shrinking the critical section inside the mutex, not from moving _which_ thread acquires it.

That said, your actor-model suggestion (a work queue where FFI callbacks push tasks, consumed by a dedicated worker pool) is a valid evolution. Issue #803 keeps that door open as a follow-on after the per-resource lock model lands — once forkchoice, block store, and sig-verify cache each have their own locks, the gossip handler only needs to lock what it actually touches, and a thread pool becomes straightforwardly safe.

Phase 2b (PR #799) already moves XMSS sig-verify off the `BeamNode` mutex while gossip is in flight — so the Tokio thread is unblocked sooner even without a full work-queue model.

GrapeBaBa:
> Great question @GrapeBaBa!
> 
> **Current threading model for `onGossip`:**
> 
> `onGossip` (and the aggregated-attestation equivalent) is called via FFI from the Rust libp2p bridge. The call chain is:
> 
> ```
> Tokio event loop (Rust thread)
>   → libp2p GossipSub event handler
>     → zeam FFI callback (beam_node_on_gossip / ...)
>       → Zig: BeamNode.onGossip()
> ```
> 
> So yes — it runs on the libp2p Tokio thread, blocking that thread for the duration of XMSS verify + forkchoice update.
> 
> **Why not just schedule it on another thread from the Rust side?**
> 
> You can, but it shifts the problem rather than solving it. If we push the FFI call off the Tokio thread into a Zig thread pool, we still need to acquire `BeamNode.mutex` on that worker thread — and any other thread doing `onBlock` or `onGossip` will contend on the same mutex. The real gain comes from shrinking the critical section inside the mutex, not from moving _which_ thread acquires it.
> 
> That said, your actor-model suggestion (a work queue where FFI callbacks push tasks, consumed by a dedicated worker pool) is a valid evolution. Issue #803 keeps that door open as a follow-on after the per-resource lock model lands — once forkchoice, block store, and sig-verify cache each have their own locks, the gossip handler only needs to lock what it actually touches, and a thread pool becomes straightforwardly safe.
> 
> Phase 2b (PR #799) already moves XMSS sig-verify off the `BeamNode` mutex while gossip is in flight — so the Tokio thread is unblocked sooner even without a full work-queue model.

The IO eventloop thread must be non blocking since it needs handle many connections, you must offload business logic from that thread.

zclawz:
You're right, and my previous reply soft-pedalled it. Conceding cleanly:

**Yes — the libp2p Tokio thread should not be running XMSS verify + STF. That's a correctness/throughput bug, not a "shifts the problem" tradeoff.** Holding the IO loop blocks GossipSub mesh maintenance, peer scoring, and every other connection's progress for the duration of the callback. On a busy mesh it'll start dropping peers, not just stalling one block.

What slice (a) actually buys here is *prerequisite*, not the fix:

1. Per-resource locks + `BorrowedState` mean the FFI callback's critical section shrinks to "snapshot pre-state, release lock" — milliseconds, not the full ~700ms aggregator/verify window.
2. Once the locks are fine-grained, the FFI handler can `cloneAndRelease` then hand the owned snapshot to a worker pool *without* the worker contending against the libxev thread on the same `BeamNode.mutex`.

So the right end state is what you described: FFI callbacks become work-queue producers, a Zig worker pool consumes them. Slice (a) is the unblock; the worker-pool flip is the actual non-blocking-IO fix.

**What I should change in the doc:** make this an explicit goal, not a "valid evolution / door open" footnote. Specifically:
- §Threads in play: rename the libp2p bridge thread's role from "executes callbacks synchronously" to "transient — slice (b/c) target is enqueue-and-return."
- Add a §IO-thread non-blocking invariant section stating the Tokio thread must not run XMSS verify, STF, or any FFI work that touches `states_lock` beyond the snapshot+release window.
- Move the work-queue/worker-pool model from "long-term direction note" into a named slice — likely (c) renamed to "off-IO-thread chain mutation" rather than just "followup off-thread."

I'll push that as r5 of the doc on this PR. Does the slice rename land okay with you, or do you want it tracked as a separate slice in #803?

Side note on #799 (phase 2b): it moves XMSS sig-verify off `BeamNode.mutex`, but you're correct that the verify still runs *on the Tokio thread*. So #799 helps libxev-side contention, not IO-thread blocking. I'll be explicit about that limitation in the doc too.

GrapeBaBa:
> You're right, and my previous reply soft-pedalled it. Conceding cleanly:
> 
> **Yes — the libp2p Tokio thread should not be running XMSS verify + STF. That's a correctness/throughput bug, not a "shifts the problem" tradeoff.** Holding the IO loop blocks GossipSub mesh maintenance, peer scoring, and every other connection's progress for the duration of the callback. On a busy mesh it'll start dropping peers, not just stalling one block.
> 
> What slice (a) actually buys here is _prerequisite_, not the fix:
> 
> 1. Per-resource locks + `BorrowedState` mean the FFI callback's critical section shrinks to "snapshot pre-state, release lock" — milliseconds, not the full ~700ms aggregator/verify window.
> 2. Once the locks are fine-grained, the FFI handler can `cloneAndRelease` then hand the owned snapshot to a worker pool _without_ the worker contending against the libxev thread on the same `BeamNode.mutex`.
> 
> So the right end state is what you described: FFI callbacks become work-queue producers, a Zig worker pool consumes them. Slice (a) is the unblock; the worker-pool flip is the actual non-blocking-IO fix.
> 
> **What I should change in the doc:** make this an explicit goal, not a "valid evolution / door open" footnote. Specifically:
> 
> * §Threads in play: rename the libp2p bridge thread's role from "executes callbacks synchronously" to "transient — slice (b/c) target is enqueue-and-return."
> * Add a §IO-thread non-blocking invariant section stating the Tokio thread must not run XMSS verify, STF, or any FFI work that touches `states_lock` beyond the snapshot+release window.
> * Move the work-queue/worker-pool model from "long-term direction note" into a named slice — likely (c) renamed to "off-IO-thread chain mutation" rather than just "followup off-thread."
> 
> I'll push that as r5 of the doc on this PR. Does the slice rename land okay with you, or do you want it tracked as a separate slice in #803?
> 
> Side note on #799 (phase 2b): it moves XMSS sig-verify off `BeamNode.mutex`, but you're correct that the verify still runs _on the Tokio thread_. So #799 helps libxev-side contention, not IO-thread blocking. I'll be explicit about that limitation in the doc too.

Yes, that is the direction I think. But further the main libxev thread actually also an eventloop thread, it shouldn't also blocking, so a dedicated worker thread will be responsible for main logic forkchoice, stf, since not much logic benefit parallel, we can avoid lots of mutex and we only need use worker thread pool for such as sign/verify signature stuff. @zclawz what do you think and redesign

zclawz:
Fully agree with the direction — and it's actually a simpler end state than what the current design doc describes.

**The cleaner architecture:**

```
libxev main thread   →  event loop only (enqueue-and-return, no state access)
Chain worker thread  →  owns forkchoice + STF + pending_blocks + states (single owner, zero contention)
ThreadPool           →  XMSS verify, BLS aggr, and other parallel crypto
```

**Why this is better than per-resource fine-grained locks:**

The current design doc (this PR) spends a lot of complexity on lock hierarchies precisely because multiple threads need to touch the same resources. But forkchoice + STF have almost no useful parallelism between them — they're inherently sequential (each block builds on the last). The per-resource lock approach buys safety at the cost of ongoing complexity and deadlock risk at every callsite.

If the chain worker thread owns those resources exclusively, we get:
- Zero locking overhead on the hot path (no mutex on forkchoice reads/writes from worker thread)
- Eliminated deadlock classes entirely — only the chain worker touches chain state
- FFI callbacks become cheap: snapshot pre-state (grab a lightweight ref/copy), release immediately, hand to worker thread queue
- libxev main thread stays as fast as it needs to be for peer/connection management

**The libp2p side mirrors this:** PR #670 already did exactly this for the Rust Swarm — FFI functions enqueue `SwarmCommand`s, Tokio event loop processes them. Same pattern on the Zig side.

**What this means for the design doc:**

The slice (a) per-resource locking approach becomes narrower in scope (or possibly just infrastructure for the interim period). The actual target state is:
1. `§Threads in play`: libxev → event router only, chain worker thread → chain state owner, threadpool → crypto
2. `§Lock hierarchy`: simplifies significantly — chain worker owns its resources without sharing; only cross-thread state (network maps, public key cache) needs fine-grained locks
3. A new `§Chain worker thread` section explaining ownership model, queue design, and how FFI callbacks hand off to it

Happy to update the doc to reflect this as the explicit target. Should the chain worker thread be a separate new slice (say (f) or rename current (b/c)), or should it become the new target state for this PR, with the per-resource locks scoped only to the resources that genuinely need cross-thread sharing (network maps, key cache)?

g11tech (approved):

GrapeBaBa:
> Fully agree with the direction — and it's actually a simpler end state than what the current design doc describes.
> 
> **The cleaner architecture:**
> 
> ```
> libxev main thread   →  event loop only (enqueue-and-return, no state access)
> Chain worker thread  →  owns forkchoice + STF + pending_blocks + states (single owner, zero contention)
> ThreadPool           →  XMSS verify, BLS aggr, and other parallel crypto
> ```
> 
> **Why this is better than per-resource fine-grained locks:**
> 
> The current design doc (this PR) spends a lot of complexity on lock hierarchies precisely because multiple threads need to touch the same resources. But forkchoice + STF have almost no useful parallelism between them — they're inherently sequential (each block builds on the last). The per-resource lock approach buys safety at the cost of ongoing complexity and deadlock risk at every callsite.
> 
> If the chain worker thread owns those resources exclusively, we get:
> 
> * Zero locking overhead on the hot path (no mutex on forkchoice reads/writes from worker thread)
> * Eliminated deadlock classes entirely — only the chain worker touches chain state
> * FFI callbacks become cheap: snapshot pre-state (grab a lightweight ref/copy), release immediately, hand to worker thread queue
> * libxev main thread stays as fast as it needs to be for peer/connection management
> 
> **The libp2p side mirrors this:** PR #670 already did exactly this for the Rust Swarm — FFI functions enqueue `SwarmCommand`s, Tokio event loop processes them. Same pattern on the Zig side.
> 
> **What this means for the design doc:**
> 
> The slice (a) per-resource locking approach becomes narrower in scope (or possibly just infrastructure for the interim period). The actual target state is:
> 
> 1. `§Threads in play`: libxev → event router only, chain worker thread → chain state owner, threadpool → crypto
> 2. `§Lock hierarchy`: simplifies significantly — chain worker owns its resources without sharing; only cross-thread state (network maps, public key cache) needs fine-grained locks
> 3. A new `§Chain worker thread` section explaining ownership model, queue design, and how FFI callbacks hand off to it
> 
> Happy to update the doc to reflect this as the explicit target. Should the chain worker thread be a separate new slice (say (f) or rename current (b/c)), or should it become the new target state for this PR, with the per-resource locks scoped only to the resources that genuinely need cross-thread sharing (network maps, key cache)?

I think so. In additional the libxev main thread most important work if clock advance which keep slot, interval match wall clock, if the peers/connection management spend more time, that should be also offload from libxev main thread, you can take account this.

GrapeBaBa:
@zclawz can you confirm current rust libp2p network thread model? how many threads actually used?

zclawz:
**Current rust libp2p thread model: 1 OS thread total.**

Concretely, from `ethlibp2p.zig` and `rust/libp2p-glue/src/lib.rs`:

```
std.Thread.spawn()              ← Zig spawns exactly one OS thread per network instance
  └─ create_and_run_network()   ← enters Rust; calls Builder::new_current_thread().build()
       └─ rt.block_on(async {   ← single-threaded Tokio runtime — no worker threads, no thread pool
             start_network()
             run_eventloop()    ← tokio::select! on: swarm events | command channel | shutdown notify
           })
```

Key facts:

- **`Builder::new_current_thread()`** — the Tokio runtime has **no worker threads at all**. The current OS thread _is_ the reactor. There is no multi-thread pool.
- **1 OS thread per network instance** (`rustBridgeThread` in `EthLibp2p.init`). In practice zeam runs one network instance, so 1 thread total.
- Everything that happens on the Rust side — TCP/QUIC I/O, DNS resolution, GossipSub mesh maintenance, peer scoring, req/resp streams, **and** the FFI callbacks back into Zig (`onGossip`, `onReqResp`, etc.) — all execute serially on this one thread.
- FFI call-outs to Zig (e.g. `beam_node_on_gossip`) **block** the entire Tokio event loop for their duration. While Zig is doing XMSS verify + STF, no other swarm event can be processed.

This is the fundamental problem you've been pointing at — the "how many threads" answer is: **one**, and it's doing everything.

GrapeBaBa:
> **Current rust libp2p thread model: 1 OS thread total.**
> 
> Concretely, from `ethlibp2p.zig` and `rust/libp2p-glue/src/lib.rs`:
> 
> ```
> std.Thread.spawn()              ← Zig spawns exactly one OS thread per network instance
>   └─ create_and_run_network()   ← enters Rust; calls Builder::new_current_thread().build()
>        └─ rt.block_on(async {   ← single-threaded Tokio runtime — no worker threads, no thread pool
>              start_network()
>              run_eventloop()    ← tokio::select! on: swarm events | command channel | shutdown notify
>            })
> ```
> 
> Key facts:
> 
> * **`Builder::new_current_thread()`** — the Tokio runtime has **no worker threads at all**. The current OS thread _is_ the reactor. There is no multi-thread pool.
> * **1 OS thread per network instance** (`rustBridgeThread` in `EthLibp2p.init`). In practice zeam runs one network instance, so 1 thread total.
> * Everything that happens on the Rust side — TCP/QUIC I/O, DNS resolution, GossipSub mesh maintenance, peer scoring, req/resp streams, **and** the FFI callbacks back into Zig (`onGossip`, `onReqResp`, etc.) — all execute serially on this one thread.
> * FFI call-outs to Zig (e.g. `beam_node_on_gossip`) **block** the entire Tokio event loop for their duration. While Zig is doing XMSS verify + STF, no other swarm event can be processed.
> 
> This is the fundamental problem you've been pointing at — the "how many threads" answer is: **one**, and it's doing everything.

what about libxev main thread @zclawz

zclawz:
Good catch @GrapeBaBa — the “1 thread total” framing in my previous comment was describing the **Rust libp2p bridge** side only. The full node thread inventory is:

```
1. libxev main thread       ← Zig event loop: onInterval (slot tick), validator client, pending-block drain
2. Rust bridge thread       ← std.Thread.spawn() → create_and_run_network() → single-threaded Tokio rt
3. ThreadPool workers       ← short-lived: parallel sig-verify, aggregation compaction (spawnWg)
4. (slice c, future)        ← processFinalizationFollowup offload
```

The libxev main thread is the Zig counterpart to the bridge thread. They share all of `BeamNode` and `BeamChain` state — currently serialised by the single coarse `BeamNode.mutex`. The contention point is that the bridge thread holds this mutex for the entire duration of `onGossip` (including XMSS verify + STF), which blocks the libxev thread from doing slot-tick work, and vice versa.

That is exactly what the design doc is trying to fix: slice (a) breaks the coarse mutex into per-resource locks so that the libxev and bridge threads can run concurrently on non-overlapping state (e.g. bridge thread appending to `pending_gossip_blocks` while libxev thread is emitting chain events).

GrapeBaBa:
> **Current rust libp2p thread model: 1 OS thread total.**
> 
> Concretely, from `ethlibp2p.zig` and `rust/libp2p-glue/src/lib.rs`:
> 
> ```
> std.Thread.spawn()              ← Zig spawns exactly one OS thread per network instance
>   └─ create_and_run_network()   ← enters Rust; calls Builder::new_current_thread().build()
>        └─ rt.block_on(async {   ← single-threaded Tokio runtime — no worker threads, no thread pool
>              start_network()
>              run_eventloop()    ← tokio::select! on: swarm events | command channel | shutdown notify
>            })
> ```
> 
> Key facts:
> 
> * **`Builder::new_current_thread()`** — the Tokio runtime has **no worker threads at all**. The current OS thread _is_ the reactor. There is no multi-thread pool.
> * **1 OS thread per network instance** (`rustBridgeThread` in `EthLibp2p.init`). In practice zeam runs one network instance, so 1 thread total.
> * Everything that happens on the Rust side — TCP/QUIC I/O, DNS resolution, GossipSub mesh maintenance, peer scoring, req/resp streams, **and** the FFI callbacks back into Zig (`onGossip`, `onReqResp`, etc.) — all execute serially on this one thread.
> * FFI call-outs to Zig (e.g. `beam_node_on_gossip`) **block** the entire Tokio event loop for their duration. While Zig is doing XMSS verify + STF, no other swarm event can be processed.
> 
> This is the fundamental problem you've been pointing at — the "how many threads" answer is: **one**, and it's doing everything.

That is the problem which one thread mix cpu and io works, lighthouse has router and taskexecutor to dispatch and handle cpu works out of swarm io thread https://hackmd.io/@JVtpwRK3SwmkRIFfF0Bmyg/rylVP_WY-g  @ch4r10t33r @g11tech @anshalshukla , the libexv main thread has the same problem

zclawz:
@GrapeBaBa — yes, you're right, and the libxev main thread has the same problem as the libp2p bridge thread. Both are event loops; running XMSS verify + STF on either one stalls every other connection / timer / mesh-maintenance event sharing that loop until the work returns. That's a correctness bug (peers dropped, slot ticks missed), not a perf nuance.

Pushed **r5 of the design doc** in `b36dde2` on the slice-(a-2) implementation branch (PR #805) absorbing this. Diff: https://github.com/blockblaz/zeam/blob/feat/agent-slice-a-2-chain-primitives/docs/threading_refactor_slice_a.md

### What r5 changes

**1. New §IO-thread non-blocking invariant (the rule that drives everything else):**

> Neither the libxev main thread nor the libp2p bridge thread may run XMSS verify, STF, or any FFI work whose worst case exceeds a per-interval budget (currently 4s slot / 5 intervals = 800ms, with headroom needed inside that). FFI work that touches `states_lock` for longer than the snapshot window must be marshalled to the chain worker thread (or, for parallel-friendly crypto, the `ThreadPool`).

Slice (a) on its own does not satisfy this invariant — it only shrinks the FFI critical sections enough that the marshalling step (slice c) can be a snapshot-then-enqueue rather than a stop-the-world. The invariant is satisfied end-to-end after slice (c).

**2. New §Chain-worker thread (target architecture):**

```
┌──────────────────┐    enqueue          ┌────────────────────────────┐
│  libxev main     │ ─────────────────▶  │                            │
│  (event router,  │                     │   Chain worker thread      │
│   slot tick,     │                     │   (single owner of chain)  │
│   timers)        │                     │                            │
└──────────────────┘                     │   - drains MPSC queue       │
                                          │   - runs STF / forkchoice  │
┌──────────────────┐    enqueue          │   - emits chain events     │
│  libp2p bridge   │ ─────────────────▶  │   - prunes states           │
│  (Tokio reactor, │                     │                            │
│   FFI entry pt)  │                     └────────────┬───────────────┘
└──────────────────┘                                  │ dispatch to
                                                       ▼
┌──────────────────┐                     ┌──────────────────────────────┐
│  HTTP API task   │ ─── snapshot read ─▶  ThreadPool workers        │
│  /eth/v1/*       │     via BorrowedState   (XMSS verify, BLS aggr,  │
└──────────────────┘                     │    sig batches)             │
                                          └──────────────────────────────┘
```

Mirrors lighthouse's three-tier model from the hackmd:
- **Tier 1** (router): `network` task → `router` task → `BeaconProcessor manager` task. Nothing in tier 1 may block.
- **Tier 2** (blocking pool): `BeaconProcessor manager` is the *only* tier 1 entry point that triggers tier 2 via `spawn_blocking`. BLS verify, slashing/exit checks, `BlobsByRoot` serving, etc. all live here.
- **Tier 3** (Rayon): backfill / column-reconstruction data parallelism.

Lighthouse's invariant — "Tier 1 is the only tier that can trigger other tiers; tier 2/3 only signal back via `idle_tx` / `oneshot`" — gives us the equivalent shape:

| zeam | lighthouse |
|---|---|
| libp2p bridge thread (FFI entry point) | network task |
| libxev main thread (event router target) | router task + sync manager |
| Chain worker thread (single owner of forkchoice + STF + states) | BeaconProcessor manager + spawned async/blocking workers |
| ThreadPool workers (XMSS verify, BLS aggr) | tokio blocking pool + Rayon pools |

Mutation-side: only the chain worker touches state. That collapses every deadlock class slice (a) carefully sequences (the lock hierarchy is enforced by program structure, not by careful ordering at every callsite) and zeros locking overhead on the mutation hot path.

**3. Slice (c) repositioned and rescoped:**

Was "followup off-thread" (just `processFinalizationFollowup`). Now: **off-IO-thread chain mutation** — spawns the chain worker thread, marshalls every `chain.onBlock` / `onGossip` / `onAttestation` / `processPendingBlocks` to it via a bounded MPSC queue, leaves only event routing on libxev main and the libp2p bridge. `processFinalizationFollowup` rides along (was the original (c) goal, now subsumed). Mirrors what PR #670 already did Rust-side for the `Swarm` (FFI → `SwarmCommand` enqueue → tokio task drains the queue).

LOC estimate: slice (c) is now ~1500–2000 LOC, likely splits into (c-1) chain-worker scaffold + queue + Arc-state and (c-2) per-handler migration.

**4. Refcounted state pointers promoted to a hard prerequisite of slice (c).**

Cross-thread readers (HTTP API `/eth/v1/*`, Prometheus `/metrics`, event broadcaster SSE consumers) can outlive the chain worker's current view of `states` — e.g. the worker prunes a finalised state while an HTTP request is mid-serialisation. With slice (a)'s `BorrowedState` holding `states_lock.shared` for the borrow's lifetime, any long-running cross-thread read would block the chain worker's next mutation.

Slice (c) therefore must land `Arc<BeamState>` (or hand-rolled equivalent) so cross-thread readers `cloneAndRelease` *or* take a refcounted pointer that survives independently. `BorrowedState`'s API stays correct — it just acquires a refcount instead of holding a shared lock for long-lived borrows.

**5. Bounded queue + backpressure + LIFO/FIFO discipline:**

- Bounded MPSC queue (size TBD, ~MAX_PENDING_BLOCKS-class).
- Producers `try_send`; full queue means: bridge drops + bumps `lean_chain_queue_drops_total{source="gossip"}` (same shape as the swarm-command-channel drop in #808), libxev main partial-drains and resumes next interval, HTTP returns 503 Retry-After.
- LIFO for attestations / aggregates (freshness > order, like lighthouse), FIFO for slashings / exits / blocks (ordering matters for safety).
- Backlog visibility: `lean_chain_queue_depth{queue="..."}` gauge per queue so devnet stress tests catch unbounded growth.

**6. §Long-term direction rewritten from suggestive to definitive.**

The chain-worker shape *is* the target. The fine-grained-lock alternative is now described as the failure mode if slice (c) never lands. Forkchoice + STF have almost no useful parallelism between them — they're inherently sequential — so per-resource locks pay ongoing complexity and deadlock risk to enable concurrency that doesn't materially exist on the mutation hot path. Inside the worker thread, those locks become uncontended and effectively no-ops; they remain useful only as the read-side serialisation for the few cross-thread readers that legitimately bypass the queue.

### What this means for slice (a-2) (PR #805)

**Nothing on the wire changes.** PR #805 ships exactly the primitives + migrations described in r3/r4 — `LockedMap`, `BlockCache`, `BorrowedState`, the per-resource locks, the legacy-metric shim, the lock hierarchy. The redesign is forward-looking: it reframes what slice (a-2) is *for* (chain-worker prerequisite, not standalone solution to the libxev/bridge blocking problem) and locks in slice (c)'s scope so it can be a single coordinated PR rather than a series of follow-ups.

Slice (a-2) is now the *prerequisite* for the chain-worker target, not the target itself. With per-resource locks + `BorrowedState` in place, FFI handlers can `cloneAndRelease` and hand owned snapshots to the worker queue without contending against the libxev thread. Lock-hierarchy work survives — it still describes how the chain worker thread's owned resources interact with the few legitimate cross-thread readers (HTTP API, metrics scrape, event broadcaster).

### Open asks

@ch4r10t33r @gr3999 @anshalshukla @GrapeBaBa — ack on r5 direction? Specifically:

1. The chain-worker thread as the explicit target (vs. fine-grained locks as the end state).
2. `Arc<BeamState>` (or equivalent refcount) as a hard slice-(c) prerequisite.
3. Slice (c) splitting into (c-1) scaffold + queue + Arc-state and (c-2) per-handler migration — or do you prefer one big PR?
4. Queue discipline: LIFO for attestations/aggregates, FIFO for blocks/slashings/exits — same call lighthouse makes, ack?

Once aligned I'll open the slice (c) tracker on #803 with the concrete scope so we can pre-agree on queue shape and migration order before the implementation PR lands.

Conflicts on PR #805 are now resolved + zig 0.16 migration done; build green 26/26, tests green 50/50. Status comment posted on #805.

PR #784, opened 2026-04-26
upgrade zig 0.16.0

anshalshukla:

PR #805, opened 2026-04-29
node,chain: per-resource locks + BorrowedState (slice a-2 of #803)

zclawz:
## Slice (a-2): per-resource locks + BorrowedState

Implements slice (a-2) of the BeamNode threading refactor per the design doc at `docs/threading_refactor_slice_a.md` (commit a76c274). Adds the locking primitives and migrates every `chain.zig` callsite to use them, **without** touching `BeamNode.mutex` itself — that drops in slice (a-3).

Tracking: #803. Design tracker: #804.

This PR is **draft on purpose** — please leave it draft for human review. ~1500 LOC changed.

---

#### `pkgs/node/src/locking.zig` (new file, ~744 LOC)

Three primitives + two debug helpers:

1. **`LockedMap(K, V)`** — `std.Thread.Mutex` + `std.AutoHashMap` bundle with the small set of methods we actually use (`get` / `put` / `remove` / `fetchRemove` / `count` / `iterateLocked`). The Network-side wiring lands in (a-3); the helper ships here with unit tests so its API contract is reviewed independently.

2. **`BlockCache`** — atomic triple-update of `fetched_blocks` + `fetched_block_ssz` + `fetched_block_children` under a single `block_cache_lock`. The wiring into `network.zig` lands in (a-3) per Partha #4. `removeChildrenOf` is bounded by `MAX_CACHED_BLOCKS = 1024`.

3. **`BorrowedState`** — RAII wrapper carrying `*const BeamState` + the backing lock. Two backing variants:
   - `states_shared_rwlock` — `BeamChain.states_lock` read side (the common case).
   - `events_mutex` — `BeamChain.events_lock` (used only by the cache / DB-load paths inside `getFinalizedState`).

   `cloneAndRelease(allocator)` consumes the borrow and produces an owned snapshot, with `errdefer self.deinit()` ensuring the lock is released even on OOM-mid-clone. `deinit` is idempotent; `released: bool` sentinel + `assertReleased` panics in debug builds if a borrow is dropped without release. Naming honesty per Partha r3 #1 — the previous draft called this `sszClone`, which read like a non-mutating helper.

4. **`LockTimer`** — per-resource lock observation. Records into the new `zeam_lock_{wait,hold}_seconds{lock,site}` histograms AND double-emits into the legacy `zeam_node_mutex_{wait,hold}_time_seconds` series so existing dashboards stay alive for one release (Partha #11, r3 #5 — code-side derived shim, not a Prometheus recording rule).

5. **`tier5_depth` thread-local + `assertNoTier5SiblingHeld`** — the 5a/5b/5c sibling locks must never be co-held. A debug-build TLS depth counter increments on every tier-5 acquire and decrements on release; an assert at every entry catches violations in tests (Partha r3 #4).

#### `pkgs/node/src/chain.zig` — new lock fields

Added to `BeamChain`:

```zig
states_lock: std.Thread.RwLock = .{},
pending_blocks_lock: std.Thread.Mutex = .{},
pubkey_cache_lock: std.Thread.Mutex = .{},
root_to_slot_lock: std.Thread.Mutex = .{},
events_lock: std.Thread.Mutex = .{},
```

Plus three internal helpers around `states`:
- `statesGet(root) -> ?BorrowedState`
- `statesPutExclusive(site, root, state)` — single-key write
- `statesCommitKeepExisting(site, root, state) -> { effective, kept_existing }` — atomic insert-or-keep that **never** invalidates an existing in-map pointer (the swap-overwrite would create a UAF for in-flight borrows).
- `statesFetchRemoveExclusivePtr(site, root) -> ?*BeamState`

#### `chain.zig` callsite migration

Every `states.{get,put,fetchRemove}` callsite from the design-doc inventory has been migrated. Test sites are commented `// SAFETY: test-only, single-threaded`.

| File:line (post-migration) | Function | Operation | Lifetime classification |
|---|---|---|---|
| `chain.zig:611` | `produceBlock` | `states.get(parent_root)` → `statesGet` | `cloneAndRelease` (FFI window via `forkChoice.getProposalAttestations`) |
| `chain.zig:725` | `produceBlock` | `states.put(block_root, post_state)` → `statesPutExclusive("produceBlock.commit")` | exclusive write |
| `chain.zig:732` | `produceBlock` errdefer | `states.fetchRemove(block_root)` → `statesFetchRemoveExclusivePtr("produceBlock.errdefer")` | exclusive write |
| `chain.zig:1112` | `onBlock` | `states.get(parent_root)` → `statesGet` | `cloneAndRelease` (verify + STF window) |
| `chain.zig:1356` | `onBlock` | `states.put(fcBlock.blockRoot, post_state)` → `statesCommitKeepExisting("onBlock.commit")` | exclusive write (with keep-existing) |
| `chain.zig:1556` | `pruneStates` | `states.fetchRemove(root)` (loop) | exclusive write (single critical section over the whole prune) |
| `chain.zig:1782` | `onGossipAttestation` | `states.get(target.root)` → `statesGet` | borrow-only |
| `chain.zig:1819` | `verifyAggregatedAttestation` (called from `onGossipAggregatedAttestation`) | `states.get(target.root)` → `statesGet` | borrow-only |
| `chain.zig:1881` | `aggregate` | `states.get(head_root)` → `statesGet` | `cloneAndRelease` (FFI window via `forkChoice.aggregate`) |
| `node.zig:1489` | `publishBlock` | `states.get(block_root)` shortcut | **dropped** — see "publishBlock" below |
| `chain.zig:2337` (test) | `process and add mock blocks…` | `states.get` | `// SAFETY: test-only, single-threaded` |
| `chain.zig:3117` (test) | `produceBlock — greedy selection…` | `states.get` | `// SAFETY: test-only, single-threaded` |

Every `pubkey_cache` / `root_to_slot_cache` access is wrapped in its respective tier-5 lock (with the sibling-rule assertion). Every `last_emitted_{justified,finalized}` and `cached_finalized_state` read-modify-write is wrapped in `events_lock`.

#### `pending_blocks` migration

`processPendingBlocks` is now a one-at-a-time `orderedRemove(0)` re-scan loop per Partha #7. Indices are never assumed stable across an unlock. The `lean_pending_blocks_drain_iters` histogram measures iteration count so we can spot the O(n²) tail in production before optimising (cap or cursor — both noted in design doc).

#### `getFinalizedState` API change

Now returns **`?BorrowedState`** instead of `?*const types.BeamState`. Backing lock depends on path:
- In-memory map hit → `states_lock.shared`.
- `cached_finalized_state` hit / DB-load path → `events_lock` (the field that owns the pointer is mutated under the same lock).

**Every caller migrated** (Partha r3 #6 — grepping `states.get` does not find them):

| File:line | Caller | Migration |
|---|---|---|
| `pkgs/cli/src/api_server.zig:274` | `handleFinalizedCheckpointState` | takes the borrow, serialises into an owned `ArrayList(u8)`, then `defer borrow.deinit()` releases the lock before HTTP I/O. |

Repo-wide grep confirmed `getFinalizedState` is the only function with this `*const BeamState` outward contract; only the one HTTP route consumes it today.

#### `publishBlock` simplification

Dropped the `postState = self.chain.states.get(block_root)` shortcut. It was safe under global `BeamNode.mutex` but creates a deadlock in the new world (caller holds `states_lock.shared` from a borrow, callee `chain.onBlock` takes `states_lock.exclusive` to commit). Recompute is the single path now; `statesCommitKeepExisting` keeps the original in-map pointer untouched on the duplicate-import path so locally produced blocks no longer leak their first post-state on the publish hop.

#### Per-lock metrics + legacy shim

New series:
- `zeam_lock_wait_seconds{lock,site}`
- `zeam_lock_hold_seconds{lock,site}`
- `lean_pending_blocks_drain_iters`

Legacy `zeam_node_mutex_{wait,hold}_time_seconds{site}` is double-emitted by `LockTimer` summing observations across all per-resource locks. Operators do **not** redeploy Prometheus or change recording rules — the legacy series stays alive automatically for one release, then is removed in the release after.

#### What's NOT in this PR (saved for a-3)

- Network maps (`pending_rpc_requests`, `pending_block_roots`, `timed_out_requests`, `connected_peers`) → `LockedMap` migration.
- `BlockCache` Network-side wiring (the helper itself ships here, with unit tests).
- `BeamNode.mutex` rename to `finalization_lock` and dropping it from `onGossip` / `onInterval` / `onReqRespResponse`.
- Lock-free `onReqRespRequest`.
- Stress test scenarios.

Slice (a-2) keeps `BeamNode.mutex` wrapping every chain entry point — the new chain locks are nested inside it (redundant under the outer mutex but correct). They become the actual synchronisation once (a-3) drops the outer mutex from those callbacks. There is **no** `external_mutex` parameter on `chain.onBlock` to drop because it was never landed on `main` between #798–#801; the design doc's "drop external_mutex" item is therefore a no-op against the current tree.

---

### Tests

| Test | Location | Covers |
|---|---|---|
| `LockedMap: ctor + get/put/remove + count` | `locking.zig` | basic lifecycle |
| `LockedMap: fetchRemove returns the prior entry` | `locking.zig` | KV-pair return shape |
| `LockedMap: iterator-while-locked sees every inserted entry` | `locking.zig` | iterator API |
| `LockedMap: deinit on empty map is a no-op` | `locking.zig` | empty deinit |
| `LockedMap: deinit on non-empty map frees internal storage` | `locking.zig` | non-empty deinit, leak-free |
| `BorrowedState: deinit is idempotent and releases the lock` | `locking.zig` | RwLock backing + double-deinit |
| `BorrowedState: events_mutex backing also releases` | `locking.zig` | Mutex backing variant |
| `BorrowedState: assertReleased fires only when not released` | `locking.zig` | debug-build assertion |
| `BorrowedState: cloneAndRelease releases lock on OOM-mid-clone` | `locking.zig` | FailingAllocator(`fail_index=0`) → lock released, error propagated |
| `BlockCache: helper init/deinit when empty` | `locking.zig` | testing.allocator leak detection |
| `BlockCache: removeChildrenOf is bounded by MAX_CACHED_BLOCKS` | `locking.zig` | BFS bound |
| `tier5 depth counter increments and decrements` | `locking.zig` | Debug-only TLS counter balance |
| `BorrowedState: cloneAndRelease success path against real BeamState` | `chain.zig` | Real BeamState clone via genMockChain — sanity-checks slot equality and lock release |
| `BlockCache: insert + get + removeChildrenOf bounded` | `chain.zig` | End-to-end with real SignedBlock + ssz buffer |
| `BlockCache: partial-state invariant (re-insert leaves no orphans)` | `chain.zig` | Repeated insert at same root frees prior block + ssz |

All existing chain.zig integration tests continue to pass — they cover the migrated states map / events_lock / pending_blocks paths through the produceBlock + onBlock + onBlockFollowup integration shape.

---

### Build + test status

```
$ zig fmt pkgs/        # clean
$ zig build -Doptimize=Debug
Finished `release` profile [optimized] target(s) in 0.36s
$ zig build --summary all test -Doptimize=Debug
Build Summary: 50/50 steps succeeded; 158/158 tests passed
```

🟢 Green. Ready for review.

---

### Spec link

Design: `docs/threading_refactor_slice_a.md` @ commit [`a76c274`](https://github.com/blockblaz/zeam/commit/a76c274) ("docs: r3 polish from Partha verification"). The PR implements every line item listed in §"Status (r4 — ready to cut code)".

### Notes on deviations from the design doc

- **`statesCommitKeepExisting` instead of pure `states.put`** — discovered while migrating `onBlock` that an unconditional overwrite invalidates in-flight borrows handed out by `statesGet`. The keep-existing semantics preserve pointer identity for any concurrent reader holding `states_lock.shared`. Equivalent rationale to the design doc's "never invalidate a pointer that other borrows might still observe."
- **`publishBlock` no longer passes a cached `postState`** — see §publishBlock simplification above. Deviates from the design-doc inventory which only said "migrate the callsite" without specifying the lock-dance shape; the recompute path is the only one that survives both the borrow contract and the deadlock constraint.
- **Tier-5 sibling depth counter is process-wide thread-local, not per-`BeamChain`** — matches the design doc but worth flagging: a process running multiple chains in parallel (today there is one) would need per-chain counters. Not a problem for slice (a).

`BlockCache` Network-side wiring is intentionally deferred to (a-3) per the design doc's "Use your judgment" clause; the helper + unit tests live in this PR so reviewers can review the API contract independently.


zclawz:
**Conflicts resolved + zig 0.16 API migration completed (HEAD `be54314`).**

The branch was conflicting with main after PR #784 (zig 0.16.0 upgrade, +1013 / −902 across 57 files). Resolution shipped in three commits on top of the previous tip:

1. **`5711654` — Merge branch `main` into the slice branch.**
   Two textual conflicts:
   - `pkgs/metrics/src/lib.zig`: histogram `init` calls now take an `io` parameter on main (zig 0.16 prometheus crate change). Applied the same `io` parameter to the three new histograms added on this branch (`zeam_lock_wait_seconds`, `zeam_lock_hold_seconds`, `lean_pending_blocks_drain_iters`); kept the `LEGACY` notes on the `zeam_node_mutex_*` series so the code-side derived double-emit shim retains its documentation.
   - `pkgs/node/src/lib.zig`: kept `_ = @import("./locking.zig");` from this branch.

2. **`b36dde2` — docs: r5 of the design doc** (chain-worker target absorbed per @GrapeBaBa / lighthouse hackmd; substantive write-up posted on #804).

3. **`be54314` — fix: zig 0.16 API migration for slice (a-2) primitives.**
   `std.Thread.Mutex` / `std.Thread.RwLock` no longer exist in zig 0.16 (replaced by `std.Io.{Mutex,RwLock}` with `io: Io` arg + cancellable semantics). Main introduced `zeam_utils.{SyncMutex,SyncRwLock}` as drop-in wrappers presenting the old API. Migrated `pkgs/node/src/{chain,locking}.zig` to use those wrappers.
   `std.time.Timer` was also removed in zig 0.16 — rewrote `LockTimer` to use `zeam_utils.monotonicTimestampNs()` with the same wait/hold semantics and the same legacy-shim emission into `zeam_node_mutex_{wait,hold}_time_seconds`.
   `refAllDeclsRecursive` does not exist in zig 0.16's `std.testing` — reverted to `refAllDecls`.

### Verification

Both run locally against zig 0.16.0 at HEAD `be54314`:

| | result |
|---|---|
| `zig build --summary all` | **26/26 steps succeeded** |
| `zig build --summary all test` | **50/50 steps succeeded**, including the new `LockedMap` / `BlockCache` / `BorrowedState` unit tests in `pkgs/node` |

GitHub now reports `mergeable: true` against current main.

### What did *not* change on the wire

The three primitives (`LockedMap`, `BlockCache`, `BorrowedState`) and every `chain.zig` migration point are byte-for-byte unchanged from the pre-merge state — only the underlying mutex/rwlock type names and the timer plumbing differ, and those are dictated by the zig 0.16 stdlib reshuffle, not by review feedback. `BorrowedState.cloneAndRelease` semantics, the `errdefer` lock-release on OOM-mid-clone, the `released: bool` sentinel, the tier-5 TLS depth assertion, and the legacy `zeam_node_mutex_*` shim all behave identically.

### Companion design-doc revision

`b36dde2` revises `docs/threading_refactor_slice_a.md` to **r5** — the chain-worker target architecture per @GrapeBaBa's hackmd reference. See the parallel comment on #804 for the substantive write-up and what slice (a-2) being the *prerequisite* (rather than the end-state) means for review.

Ready for human review.

ch4r10t33r (approved):
LGTM

PR #809, opened 2026-04-30
fix(network): relieve rust-bridge swarm command channel backpressure (#808)

### LINKED ISSUE #808: rust-bridge: swarm command channel saturating under devnet-4 load — silent attestation/reqresp drops causing fork-choice divergence
## Summary

On devnet-4 we are seeing the **per-network swarm command channel introduced in #789** fill up under steady-state validator load and silently drop attestation publishes / outbound `req-resp` requests. The dropped messages line up with multiple competing fork branches in the local fork-choice tree because gossiped attestations and parent-by-root fetches never make it onto the wire.

This is observed on the running devnet-4 deployment (post-#789 build, both `zeam_0` and `zeam_1`).

## Symptoms

```
Apr-30 16:49:01.813 [s=574 i=1] [info]  [validator] constructed attestation slot=574 validator=0
Apr-30 16:49:01.813 [s=574 i=1] [info]  [node] adding locally produced attestation to chain: slot=574 validator=0
Apr-30 16:49:01.815 [s=574 i=1] [error] [network] rust-bridge: send_swarm_command: command channel full, dropping command (slow drain or peer backpressure)
Apr-30 16:49:01.815 [s=574 i=1] [info]  [node] published attestation to network: slot=574 validator=0   <-- log says "published" but command was dropped
```

The `[node] published …` info log fires unconditionally even though the underlying `try_send` returned `Err(Full)` and the gossip never left the host. Same pattern for proposed blocks (slot 576 above) and `send_rpc_request: failed to enqueue request: command channel full` for outbound req-resp.

## Frequency (last 1 h on devnet-4, 2026-04-30 ~15:51 → 16:51 UTC)

| Container | `command channel full` | `Outbound error` (reqresp) | total `[error]` |
|-----------|-----------------------:|----------------------------:|----------------:|
| `zeam_0`  |                    317 |                         210 |             816 |
| `zeam_1`  |                    311 |                         193 |             784 |

So we are dropping ~5 commands/min per node, sustained. Many slots see drops on **both** the locally produced attestation and the published block — the chain head visibly stops advancing and the local fork choice grows extra branches:

```
CHAIN STATUS: Current Slot: 574 | Head Slot: 560 | Behind: 14
ForkChoice Tree:
5fa6(139)... ─ e845(144) - 2 branches
 ├── [ ]─ da92(146)... ─ 038f(168) - 3 branches
 ├── [3]─ 7f9f(172)... ─ b2ce(496)─ … ─ 235b(544)─ [15]─ fefc(560)
 ├── [2]─ 8f7f(171)─ [4]─ c517(176)─ [ ]─ 1325(178)─ [ ]─ 35d5(180)
 └── f8f8(169)
 └── 3d0d(145)
```

(Multiple sibling branches at slots 144, 146, 168, 171, 176 etc. that should have been resolved by attestation gossip.)

## Root cause (code pointers — `main` @ d822793)

`rust/libp2p-glue/src/lib.rs`

- L237 `const SWARM_COMMAND_CHANNEL_CAPACITY: usize = 1024;`
- L243 `const MAX_SWARM_COMMANDS_PER_TICK: usize = 32;`
- L485 `fn send_swarm_command(network_id, cmd) -> bool` — uses `tx.try_send(cmd)`; on `Full(_)` it logs the error and returns `false`. **No retry, no backpressure to the FFI caller.**
- L606 `send_rpc_request` does the same `try_send` against the same channel.
- L992 channel is constructed bounded with capacity 1024.

The drainer (`MAX_SWARM_COMMANDS_PER_TICK = 32`) inside the `tokio::select!` branch on `cmd_rx` only pulls 32 commands per loop iteration before yielding. When the swarm side is busy (gossip flood, slow peer, reqresp work), the producer side fills 1024 messages faster than the consumer drains and we start dropping.

The Zig caller is given `false` / `request_id == 0` but the **node-level** logs (`[node] published attestation to network`) don't observe the bool and unconditionally log success. So from the operator's point of view the message looks published even when the command was dropped.

## Impact

- Locally produced attestations are silently dropped → other clients don't see our votes.
- Locally proposed blocks get the same treatment if the block-publish lands during a burst.
- Outbound `req-resp` (parent-by-root, blocks-by-range) requests are dropped → our fork choice never converges on the canonical head, observed as multiple sibling branches and `Behind: 14` style head lag.
- The `[node] published …` info log lies about success.

## Suggested fixes

1. **Honour the `bool` / `request_id == 0` return at the callsite.** `pkgs/network/src/ethlibp2p.zig` should treat a dropped command as a publish failure: bump a metric, do not emit the "published …" info log, and let the validator/node decide whether to retry on the next slot.
2. **Increase `SWARM_COMMAND_CHANNEL_CAPACITY` and/or `MAX_SWARM_COMMANDS_PER_TICK`.** 1024 / 32 is conservative for a node validating at 1 attestation/slot/validator plus parent fetches. Worth tuning for devnet sizes (try 8192 / 256) and exposing as config.
3. **Distinguish user-initiated publish vs. forwarded gossip.** Right now both paths share one channel; under load gossip-forwarding can starve our own attestations. Two channels (or a small priority queue with our own publishes promoted) would prevent self-attestations from being the ones we drop.
4. **Add a metric** `zeam_libp2p_swarm_command_dropped_total{reason="full"|"closed"}` and a Grafana panel — would let us alert on this directly instead of grep-counting Loki.
5. **Optional**: when `try_send` fails with `Full`, fall back to `tx.send().await` with a short timeout (e.g. 50 ms) before giving up, since dropping our own attestation is worse than a 50 ms delay.

## Reproduction

Devnet-4, zeam_0 / zeam_1 containers, last hour of Loki at `http://46.225.10.32:3100`. Query:

```logql
{container="zeam_0"} |~ `command channel full`
```

## References

- PR #789 — refactor: use actor model for rust-libp2p bridge (introduced the bounded channel + log line)
- PR #796 — parallelise: review follow-ups (does not touch this path)

Reported by Partha (zeam community, devnet-4 op log).

zclawz:
## Summary

Fixes #808.

The actor-model rust-libp2p bridge introduced in #789 bounds the per-network swarm command channel to **1024 slots** and drains **32 commands per event-loop tick**. On devnet-4 this saturates under steady-state validator load — Loki shows ~5 `command channel full` errors / minute / node — and the dropped messages line up with multiple competing fork-choice branches in the local tree because outbound attestations and req-resp parent fetches never make it onto the wire.

To make matters worse the node-level publish path logged `published attestation/block to network` unconditionally, even when the underlying `try_send` returned `Err(Full)`, so operators couldn't tell from the logs that anything had been dropped.

This PR addresses both problems with the smallest reasonable diff (6 files, +136 / -28).

### Rust — `rust/libp2p-glue/src/lib.rs`
- `SWARM_COMMAND_CHANNEL_CAPACITY`: **1024 → 8192** (~8x headroom for the same workload, still bounded).
- `MAX_SWARM_COMMANDS_PER_TICK`: **32 → 256** to actually drain the new headroom without monopolizing the executor.
- `publish_msg_to_rust_bridge` now returns `bool` (`true` on enqueue, `false` when the command was dropped). Returns `false` for null topic and for uninitialized / full / closed channels.
- Added two regression tests covering the new return contract:
  - `test_publish_msg_to_rust_bridge_returns_false_when_uninitialized`
  - `test_publish_msg_to_rust_bridge_returns_false_on_null_topic`

### Zig — `pkgs/network/`, `pkgs/node/`
- Threaded the bool return up through `GossipSub.publishFn`, `NetworkBackend.publish`, and `Node.publishBlock` / `publishAttestation` / `publishAggregation`.
- The `[node] published … to network` info log only fires when the backend accepted the publish; otherwise we emit a new
  `[node] failed to publish … (backend dropped publish)` **warn** line so the situation is visible in Loki and to operators.
- Mock backend returns `true` unconditionally (synchronous, no command channel) and the existing `mock.zig` publish test now asserts it.

## What's intentionally NOT in this PR

Issue #808 lists three additional follow-ups that are out of scope here to keep the diff reviewable:

1. **Metric** `zeam_libp2p_swarm_command_dropped_total` — needs new Rust→Zig FFI plumbing into the metrics registry. Worth its own PR.
2. **Priority split** between own-publishes and forwarded gossip — needs a second channel + select biasing. Design discussion would help before coding.
3. **`send().await` fallback** with a short timeout before dropping our own attestations — possibly desirable, but changes the FFI contract from non-blocking to potentially-blocking which warrants a separate review.

The capacity bump alone should be sufficient to clear the symptom on devnet-4 (the channel was at most ~1.5x oversubscribed at peak).

## Test plan

```
zig build              # ✓ EXIT:0
zig build test         # ✓ EXIT:0
cd rust && cargo test -p libp2p-glue
# 7 passed; 0 failed (5 existing + 2 new)
```

## ABI / FFI impact

`publish_msg_to_rust_bridge` changes its return type from `void` to `bool`. No other FFI symbols are affected. Both sides are updated atomically in this PR — there is no mixed-build window.

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -60,6 +60,12 @@ const Metrics = struct {
     lean_connected_peers: LeanConnectedPeersGauge,
     lean_peer_connection_events_total: PeerConnectionEventsCounter,
     lean_peer_disconnection_events_total: PeerDisconnectionEventsCounter,
+    // Issue #808: per-reason count of swarm commands dropped before reaching
+    // the rust-libp2p event loop (channel full / closed / uninitialized).
+    // Refreshed from a Rust-side atomic on every scrape via a registered
+    // refresher — see `registerScrapeRefresher` and the network-layer
+    // implementation in `pkgs/network/src/ethlibp2p.zig`.
+    lean_libp2p_swarm_command_dropped_total: LibP2pSwarmCommandDroppedCounter,
     // Node lifecycle metrics
     lean_node_info: LeanNodeInfoGauge,
     lean_node_start_time_seconds: LeanNodeStartTimeGauge,
@@ -141,6 +147,7 @@ const Metrics = struct {
     const LeanConnectedPeersGauge = metrics_lib.GaugeVec(u64, struct { client: []const u8, client_type: []const u8 });
     const PeerConnectionEventsCounter = metrics_lib.CounterVec(u64, struct { direction: []const u8, result: []const u8 });
     const PeerDisconnectionEventsCounter = metrics_lib.CounterVec(u64, struct { direction: []const u8, reason: []const u8 });
+    const LibP2pSwarmCommandDroppedCounter = metrics_lib.CounterVec(u64, struct { reason: []const u8 });
     // Node lifecycle metric types
     const LeanNodeInfoGauge = metrics_lib.GaugeVec(u64, struct { name: []const u8, version: []const u8 });
     const LeanNodeStartTimeGauge = metrics_lib.Gauge(u64);
@@ -498,6 +505,7 @@ pub fn init(allocator: std.mem.Allocator) !void {
         .lean_connected_peers = try Metrics.LeanConnectedPeersGauge.init(allocator, "lean_connected_peers", .{ .help = "Number of connected peers" }, .{}),
         .lean_peer_connection_events_total = try Metrics.PeerConnectionEventsCounter.init(allocator, "lean_peer_connection_events_total", .{ .help = "Total number of peer connection events" }, .{}),
         .lean_peer_disconnection_events_total = try Metrics.PeerDisconnectionEventsCounter.init(allocator, "lean_peer_disconnection_events_total", .{ .help = "Total number of peer disconnection events" }, .{}),
+        .lean_libp2p_swarm_command_dropped_total = try Metrics.LibP2pSwarmCommandDroppedCounter.init(allocator, "lean_libp2p_swarm_command_dropped_total", .{ .help = "Total number of swarm commands dropped before reaching the rust-libp2p event loop, by reason (issue #808)" }, .{}),
         // Node lifecycle metrics
         .lean_node_info = try Metrics.LeanNodeInfoGauge.init(allocator, "lean_node_info", .{ .help = "Node information (always 1)" }, .{}),
         .lean_node_start_time_seconds = Metrics.LeanNodeStartTimeGauge.init("lean_node_start_time_seconds", .{ .help = "Start timestamp" }, .{}),
@@ -586,6 +594,19 @@ pub fn init(allocator: std.mem.Allocator) !void {
     g_initialized = true;
 }
 
+/// Optional pre-scrape refresher. Modules that own state outside the
+/// `Metrics` struct (e.g. a Rust-side atomic counter accessed via FFI) can
+/// register a callback here; it is invoked on every `writeMetrics` so the
+/// counter values reflect the latest source-of-truth at scrape time. Issue
+/// #808 (libp2p swarm command drops) is the first user.
+var g_scrape_refresher: ?*const fn () void = null;
+
+/// Register (or replace) a scrape refresher. Pass `null` to clear. Safe to
+/// call before `init()`; the registration sticks regardless of init order.
+pub fn registerScrapeRefresher(refresher: ?*const fn () void) void {
+    g_scrape_refresher = refresher;
+}
+
 /// Writes metrics to a writer (for Prometheus endpoint).
 pub fn writeMetrics(writer: *std.Io.Writer) !void {
     if (!g_initialized) return error.NotInitialized;
@@ -595,6 +616,10 @@ pub fn writeMetrics(writer: *std.Io.Writer) !void {
         try writer.writeAll("# Metrics disabled for ZKVM target\n");
         return;
     }
+
+    // Pull in any externally-owned counters (e.g. Rust-side libp2p drops)
+    // before serializing so each scrape returns up-to-date values.
+    if (g_scrape_refresher) |refresher| refresher();
 
     try metrics_lib.write(&metrics, writer);
 }
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -65,7 +65,7 @@ const Metrics = struct {
     // Refreshed from a Rust-side atomic on every scrape via a registered
     // refresher — see `registerScrapeRefresher` and the network-layer
     // implementation in `pkgs/network/src/ethlibp2p.zig`.
-    lean_libp2p_swarm_command_dropped_total: LibP2pSwarmCommandDroppedCounter,
+    zeam_libp2p_swarm_command_dropped_total: LibP2pSwarmCommandDroppedCounter,
     // Node lifecycle metrics
     lean_node_info: LeanNodeInfoGauge,
     lean_node_start_time_seconds: LeanNodeStartTimeGauge,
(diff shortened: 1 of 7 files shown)


ch4r10t33r:
Quick review — flagging edge cases before merge:

1. **`send_rpc_request` not threaded.** #808 names req-resp drops as half the fork-choice divergence; this PR only fixes `publish_msg_to_rust_bridge`. `send_rpc_request` still drops silently with `request_id == 0` and no caller audit.
2. **Drop branch untested on Zig side.** `mock.zig::publish` returns `true` unconditionally → new `else` arms in `publishBlock` / `publishAttestation` / `publishAggregation` have no test coverage. Suggest a `force_drop` knob on the mock.
3. **No saturation test on Rust side.** Both new tests cover null-topic / uninitialized; neither exercises the actual full-channel path. Please add: pause drainer, push 8193 commands, assert last returns `false`.
4. **No metric.** Without `zeam_libp2p_swarm_command_dropped_total` we can't verify the capacity bump is actually sufficient on devnet-4 post-merge — Loki grep is the only signal. Strong preference to land the counter here or immediately after.
5. **8192 sized by anecdote.** "1.5× peak oversubscription → 8× headroom" is steady-state math; burst peak unknown. Tied to (4).
6. **`Full` vs `Closed` collapsed to same `false`.** Future retry policy will need them split; doing it now avoids changing the warn-line contract later.
7. **Local-chain vs network divergence persists on drop.** `chain.onBlockFollowup` / `onGossipAggregatedAttestation` run regardless of publish bool — message is in local fork-choice but never gossiped, no retry path. Visibility improved, divergence not resolved.

Stop-gap is fine; suggest keeping #808 open until at least 1, 2, 3, 4 land.

zclawz:
Pushed `cd7a6a7` addressing review points 1-4 plus the lint failure. Diff for this round: 4 files, +266 / -10.

**Lint** — `cargo fmt` complaint on the multi-line FFI call. Now passes locally:
- `zig fmt --check .` ✓
- `cargo fmt --check` ✓
- `cargo clippy --all-targets -D warnings` ✓

**#1 — `send_rpc_request` not threaded.** Bumped the new per-reason drop counter on every `send_rpc_request` failure path (uninitialized / full / closed) so the same metric covers both publish and req-resp drops. Also added a Zig-side warn in `EthLibp2p.sendRPCRequest` at the `request_id == 0` callsite — operators correlating req-resp timeouts now see the dispatch-failure event in the Zig log stream alongside the preceding rust-bridge error (the only previous signal was a generic `error.RequestDispatchFailed` that the upper layers swallowed).

**#2 — drop branch untested on Zig side.** Added `Mock.setForcePublishDrop(bool)` knob; when set, every `publish` returns `false` without invoking subscribers. The existing mock gossip test now exercises both the normal and dropped paths back-to-back.

**#3 — no saturation test on Rust side.** New test `test_swarm_command_full_channel_drops_and_counts`: installs a bounded sender of capacity 4 into `COMMAND_SENDERS` without a drainer, fills it to capacity, then asserts the next 3 sends each return `false` and each bump the `Full` counter. Runs in microseconds vs. allocating 8192 commands.

**#4 — no metric.** Added `lean_libp2p_swarm_command_dropped_total{reason="full|closed|uninitialized"}`:
- Rust: `SwarmCommandDropReason` enum + `static SWARM_COMMAND_DROPPED_TOTAL: [AtomicU64; 3]` bumped on every drop path. New FFI getter `get_swarm_command_dropped_total(reason_tag: u32) -> u64` exposes the cumulative count. Unknown tags return 0 so a Zig build compiled against an older Rust glue can't panic.
- Zig: `pkgs/metrics` declares the labeled CounterVec and a new `registerScrapeRefresher` hook (`writeMetrics` calls it before serializing). `pkgs/network/src/ethlibp2p.zig` registers a refresher in `EthLibp2p.init` that polls all 3 reason tags and `incrBy`s the delta. Counter is monotonic + per-reason, so no double-counting if multiple `EthLibp2p` instances co-exist.

New Rust tests assert the counter increments on uninitialized + full paths and that unknown reason tags return 0. **10 Rust tests pass total** (5 existing + 5 new).

**Intentionally still deferred** to follow-up PRs (review points 5-7):
- (5) Capacity tuning numbers based on actual devnet-4 metric data once this lands and we can read `lean_libp2p_swarm_command_dropped_total` directly.
- (6) Splitting `Full` vs `Closed` in the warn-line contract — counter already splits them; widening the warn-line API is easier once retry policy is decided.
- (7) Local-chain vs network divergence on drop (no retry path) — needs design discussion (re-publish queue vs. validator-level retry vs. attestation aggregation timing). I'll spin up a separate issue to scope it.

Keeping #808 open per your suggestion until those land.

ch4r10t33r:
> * lean_libp2p_swarm_command_dropped_total

as this is a zeam specific metric, we should prefix it as zeam_ instead of lean_

zclawz:
Good catch — pushed `65bbf8c` renaming the metric to `zeam_libp2p_swarm_command_dropped_total`.

Fits the existing convention: `lean_*` for protocol-level metrics shared across clients, `zeam_*` for client-implementation specifics (alongside `zeam_chain_onblock_duration_seconds`, `zeam_compact_attestations_*`, `zeam_node_mutex_*`, `zeam_fork_choice_tick_interval_*`). The rust-libp2p command channel is purely a zeam concern, so `zeam_` is the right prefix.

Updated 3 files (rename + 5 doc-comment touch-ups). Build + fmt clean.

ch4r10t33r (approved):
Looks good.

PR #812, opened 2026-04-30
Subscribe gossipsub mesh to selective attestation subnets only (#811)

### LINKED ISSUE #811: Attestation subnet over-subscribe defeats bandwidth savings (deviation from leanSpec)
## Summary

Every node currently joins the gossipsub mesh for **every** attestation subnet at startup, regardless of which subnets its validators are assigned to. This defeats the bandwidth-saving purpose of attestation subnets and deviates from the [leanSpec reference implementation](https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/__main__.py).

## Where the bug lives

`pkgs/network/src/ethlibp2p.zig` — `EthLibp2p.run()` enumerates `0..attestation_committee_count` and joins each `attestation_<id>` topic into the comma-joined `topics_str` that gets pushed to the rust libp2p bridge:

```zig
for (std.enums.values(interface.GossipTopicKind)) |kind| {
    switch (kind) {
        .attestation => {
            const subnet_count = try getAttestationSubnetCount(self.params.attestation_committee_count);
            for (0..subnet_count) |i| {
                const subnet_id: types.SubnetId = @intCast(i);
                const gossip_topic = interface.GossipTopic{ .kind = .attestation, .subnet_id = subnet_id };
                ...
                try topics_list.append(self.allocator, topic_str);
            }
        },
        ...
    }
}
```

The rust side (`rust/libp2p-glue/src/lib.rs:1793-1799`) calls `swarm.behaviour_mut().gossipsub.subscribe(&topic)` for every entry, so the gossipsub mesh genuinely joins every subnet.

`pkgs/node/src/node.zig` `BeamNode.run()` does build the *correct* selective list (validator-derived subnets + explicit `aggregation_subnet_ids` for aggregators + fallback to subnet 0):

```zig
for (validator.ids) |validator_id| {
    const subnet_id = try types.computeSubnetId(@intCast(validator_id), committee_count);
    if (seen_subnets.contains(@intCast(subnet_id))) continue;
    try seen_subnets.put(@intCast(subnet_id), {});
    try topics_list.append(self.allocator, .{ .kind = .attestation, .subnet_id = @intCast(subnet_id) });
}
...
try self.network.backend.gossip.subscribe(topics_slice, handler);
```

But `gossip.subscribe()` here only registers in-process zig handler callbacks (`pkgs/network/src/interface.zig:939-949`); it never reaches the rust gossipsub subscribe call. So you have:

- Rust gossipsub mesh: **all N subnets joined** (uses the `EthLibp2p.run()` enumeration)
- Zig handler dispatch: **selective** (uses the `BeamNode.run()` list)

Inbound dispatch at `interface.zig:885-889` early-exits with no handlers when an attestation arrives on a subnet the BeamNode didn't register, so message validation CPU is partially saved. Bandwidth, decode, and decompress are not.

## Effect

| Layer | Filtered? |
|-------|-----------|
| libp2p gossipsub mesh | No — all N subnets |
| Inbound bandwidth | No |
| Decode/decompress CPU | No |
| Validation (STF) CPU | Yes — early-exit at handler-map miss |

A node with one validator on subnet 3 still receives every attestation gossiped on subnets 0, 1, 2, 4, 5, ... and pays full bandwidth for them. With \`attestation_committee_count\` = N, per-node attestation BW is N× the spec target.

## leanSpec for comparison

`leanSpec/src/lean_spec/__main__.py:515-549` builds exactly one selective list and calls `event_source.subscribe_gossip_topic(...)` per entry. \`subscribe_gossip_topic\` directly invokes \`_gossipsub_behavior.subscribe(topic)\` (`event_source.py:681-692`), which is the actual mesh join. Single layer, no parallel "join all" path. The comment block at \`__main__.py:517-519\` even spells out the design intent zeam loses:

> Subscribing to a subnet causes the p2p layer to receive all messages on it. Not subscribing saves bandwidth — messages arrive only if the node publishes to that subnet (via gossipsub fanout), not as a mesh member.

## Fix shape

Two viable options:

1. **Drive rust subscribes from BeamNode.run()'s selective list.** Either by making \`EthLibp2p.run()\` derive its topic set from \`gossipHandler.onGossipHandlers\` (which BeamNode populates), and reordering \`Node.run()\` so \`BeamNode.run()\` happens before \`EthLibp2p.run()\`. **(See associated PR — this is the minimal patch.)**

2. **Add a runtime subscribe API zig→rust** so subscriptions can be added/removed after the swarm is up, enabling runtime mesh reshape on validator add/remove.

## Other gaps (out of scope for the immediate fix)

- **No ENR \`attnets\` bitfield advertisement.** Discovery cannot route by subnet membership. leanSpec has the container plumbing (\`enr/eth2.py:13\`, \`enr/keys.py:52\`) but zeam has no equivalent.
- **No subnet rotation.** Ethereum's \`RANDOM_SUBNETS_PER_VALIDATOR\` (epoch-based subnet rotation per validator to backfill mesh) is absent.
- **No runtime subscribe/unsubscribe.** Validator hot-add/remove and admin-API aggregator toggle don't reshape the mesh; \`node.zig:1629-1634\` documents this as a fixed-startup decision.

ch4r10t33r:
Fixes #811.

## Summary
- `EthLibp2p.run()` enumerated every `attestation_<id>` topic up to `attestation_committee_count` and pushed them all to the rust libp2p bridge, joining the gossipsub mesh to every subnet on every node — irrespective of which subnets the node's validators were assigned to.
- `BeamNode.run()` already builds the spec-correct selective list (validator-derived subnets + explicit aggregator extras + fallback to subnet 0) and registers it via `gossip.subscribe(...)`, but that path only populates the in-process zig handler map; it never reached the rust mesh subscribe.
- Net effect: every node paid full attestation bandwidth for every subnet. CPU was partially saved (validation early-exits on unregistered subnets) but bandwidth, decode, and decompress all ran flat. Deviation from leanSpec (`src/lean_spec/__main__.py:515-549`).

## What changed
- `pkgs/network/src/ethlibp2p.zig` — `run()` no longer enumerates kinds. It iterates `self.gossipHandler.onGossipHandlers` (populated by `BeamNode.run()` via `gossip.subscribe`) and encodes each `GossipTopic` into the comma-joined `topics_str` for the rust bridge. One source of truth for both layers. Dead helper `getAttestationSubnetCount` removed.
- `pkgs/cli/src/node.zig` — `Node.run()` reorders `beam_node.run()` before `network.run()` so the handler map is populated before `EthLibp2p.run()` reads it.
- `pkgs/cli/src/main.zig` — `DelayedNodeRunner.onInterval` (multi-node simtest) reordered the same way.

A startup warning fires if `EthLibp2p.run()` runs with an empty handler map, which would indicate the call order regressed.

## Test plan
- [x] `zig build all` — passes
- [x] `zig build test` — 151/151 unit tests pass
- [x] `zig build simtest` — multi-node devnet integration test passes (including the delayed-sync node 3 path that exercises the reordered DelayedNodeRunner)

## Out of scope (tracked in #811)
- ENR `attnets` bitfield advertisement (discovery-side filtering).
- Subnet rotation (`RANDOM_SUBNETS_PER_VALIDATOR`).
- Runtime subscribe/unsubscribe on validator add/remove or admin-API aggregator toggle.


zclawz (commented):
Solid fix and a clear improvement for the devnet-4 subnet bandwidth/divergence issue I flagged earlier — replied to the underlying problem at the right layer (single source of truth = `BeamNode.run`'s spec-correct selection, with `EthLibp2p.run` consuming it instead of independently re-deriving). Diff is small, narrow, and the rationale comments are excellent.

Reproduced locally:
- `zig build` ✓
- `zig build test` ✓ (all tests pass)

## Things I checked and like

- **Topic-key correctness for the new iteration:** `GossipTopic` is `{kind, ?subnet_id}` used with `AutoHashMap`, so `attestation_0` and `attestation_1` get distinct keys; `block` and `aggregation` use `subnet_id: null` consistently between the old enumerate path and the new selective path, so there's no key-collision regression.
- **Empty topic list is safe:** if the handler map is empty for some reason, `joinZ` produces an empty C string, and the Rust glue's `CStr → split(",") → filter(!is_empty())` yields an empty `topics` Vec. Rust side won't panic; gossipsub mesh just starts empty. The new `[network] EthLibp2p.run() called before any gossip subscriptions were registered` warn covers operator visibility.
- **Order in `Node.run`:** swapping `beam_node.run()` before `network.run()` is correct, because `EthLibp2p.subscribe` is a pure Zig handler-map insertion (`gossipHandler.subscribe` → `onGossipHandlers.put`). It does **not** poke the Rust bridge, so it's safe to call before the Rust thread is spawned. The Rust thread subscribes to the topics list once, at startup.
- **Inbound dispatch:** `GenericGossipHandler.onGossip` already uses `onGossipHandlers.get(topic) orelse return` (debug-log, no crash), so any leftover wire traffic on a subnet we no longer subscribe to is dropped gracefully — no behavior change there.
- **Selective-subscribe log line:** the new `gossip subscriptions: block + aggregation + N attestation subnet(s) [0,1,…]` info log makes the regression I called out earlier directly observable in Loki on every restart. That alone is worth landing.

## Questions / things worth a note before merging

1. **Aggregator → subnet mapping when the aggregator's validator(s) hash to a *different* subnet.** The fallback at lines ~1625–1631 does:
   ```zig
   if (seen_subnets.count() == 0 and self.chain.isAggregator()) {
       try topics_list.append(self.allocator, .{ .kind = .attestation, .subnet_id = 0 });
   }
   ```
   This only fires when `seen_subnets` is empty. If an aggregator on subnet 1 has a validator whose `validator_id % committee_count == 0`, it ends up subscribed to subnet 0 (via the validator path) and never to subnet 1 — even though it is *the* aggregator for subnet 1. With the devnet-4 round-robin shuffle this is a real concern: nothing pins the aggregator role to validators whose subnet matches.

   I think the right shape is: **aggregator role implies subscription to that aggregator's subnet, independent of validator-derived subnets**. Either (a) require operators to set `aggregation_subnet_ids` for aggregator nodes (and validate at startup that it's non-empty when `isAggregator`), or (b) accept a `--aggregator-subnet` CLI arg and add it unconditionally when `isAggregator`. Status quo would silently misroute on subnet-1 aggregators.

   Not a blocker for this PR — the leanSpec-equivalent gap is also there — but worth tracking in #811's checklist or a follow-up issue. Could you call this out in the PR body or a `// TODO(#811)` comment near the fallback?

2. **Aggregator without validators *and* without explicit subnets** falls back to subnet 0. Agreed with the leanSpec parity, just wanted to confirm: if a future config has an aggregator on subnet 2 with no validators and no `aggregation_subnet_ids`, this still subscribes to subnet 0. Worth a comment that this is the intentional spec-parity behavior so a future reader doesn't "fix" it to subnet `aggregator_id % committee_count` or similar.

3. **`getOnGossipCbHandler` is called once and the same `handler` is registered for every topic** (block, aggregation, every attestation subnet we subscribe to). That's how `gossip.subscribe(topics, handler)` works today, but it means inbound dispatch always goes through one handler that switches on topic kind — fine, but if the dispatch ever needs per-topic handlers (e.g. different attestation processing per subnet), you'll need to re-register. Not in scope here; just noting.

4. **Empty-handler-map warn is currently `[network] EthLibp2p.run() called before any gossip subscriptions were registered`** — would suggest also logging `network_id` and the `topics_str` that ends up being passed (empty) so the asymmetry between the warn and the eventual mesh state is one log line apart. Minor.

## Nits (non-blocking)

- `pkgs/node/src/node.zig` — the `subnet_ids_buf` formatting builds a comma-list manually. Could be replaced with `std.mem.join` over a small temporary `[]u8` array for readability, but the current code is fine and avoids an extra allocation per startup. Up to you.
- The warn in `EthLibp2p.run` says "Make sure BeamNode.run() runs first" which is great, but the assertion is *operational* — would be even better as a hard error in test/debug builds (e.g. `if (builtin.mode == .Debug) return error.GossipSubscriptionsEmpty;`) so a CI regression on the call order is loud. Not a blocker.

## Verdict

**LGTM** with the aggregator-subnet question (item 1) addressed either inline as a `TODO` comment or in #811's checklist. The other items are non-blocking. Once CI on this commit goes green, happy to see it land — this directly addresses what we were seeing in the devnet-4 Loki analysis (every node subscribed to every subnet regardless of its `subnet:` config field).

For the record, the leanSpec ref in the comment is correct and the `BeamNode.run` selection mirrors `lean_spec/__main__.py:515-549` accurately.

g11tech (approved):

Issue #816, opened 2026-05-01
Add ENR attnets advertisement, discv5 attnets-aware peer filter, and backbone subnet subscriptions


ch4r10t33r:
Follow-up to #811. Now that #812 lands selective gossipsub subscribe, the mesh-connectivity problem flagged in #811's "Other gaps" section becomes the dominant remaining risk — covered here as its own issue.

## Problem

With selective subscribe, a node only joins the gossipsub mesh for subnets its validators are assigned to. That's correct for bandwidth, but it means a node now needs *enough peers on those specific subnets* for the mesh to reach `D_low`. Standard gossipsub has no way to find subnet-specific peers — it relies on either:

1. **Topic-aware peer discovery** (find peers on the subnet at the discovery layer), or
2. **Backbone density guarantees** (a fraction of nodes on every subnet regardless of duty).

zeam currently has neither. On a sparse network or an unfortunate connection topology, a subscriber can have zero same-subnet neighbours → mesh size 0 → attestations don't propagate. Diagrammatically this is the "checkerboard" failure mode.

## What the beacon spec mandates

Three normative pieces in the Ethereum consensus spec, all specifically designed to keep the mesh dense per subnet:

### A. ENR `attnets` field — ENR-based subnet discovery

[`specs/phase0/p2p-interface.md` — "Attestation subnet bitfield"](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#attestation-subnet-bitfield):

> An ENR `attnets` entry signifies the attestation subnet bitfield with the following form to more easily discover peers participating in particular attestation gossip subnets.
>
> | Key       | Value                                              |
> |:----------|:---------------------------------------------------|
> | `attnets` | SSZ `Bitvector[ATTESTATION_SUBNET_COUNT]`          |
>
> If a node's `MetaData.attnets` has any non-zero bit, the ENR MUST include the `attnets` entry with the same value as `MetaData.attnets`. … Bits SHOULD be set to true for a subnet where the validator is intending to continue collecting custodial signed aggregated attestations.

### B. Discv5-level subnet peer discovery

Same section:

> Nodes looking for peers participating in a particular subnet SHOULD search for peers via the discovery mechanism (e.g. ENR-based peer discovery via Discv5) using the `attnets` ENR entry to filter for peers that are participating in the desired subnet(s).

i.e. the spec REQUIRES that ENR `attnets` is populated AND that subnet-targeted peer search consults it. zeam does neither today.

[...]

PR #819, opened 2026-05-01
libp2p-glue: harden FFI surface (catch_unwind, drop static mut, CStr guards)

ch4r10t33r:
Addresses the critical/high findings from the libp2p-glue audit, plus a closely-related Tokio/FFI panic surfaced on devnet during sync (#838).

Closes #838.

## What it solves

- **Panics across FFI = UB.** Every `extern "C"` entry point is wrapped in `std::panic::catch_unwind` via a small `catch_ffi` helper. **Caveat:** `catch_unwind` is a no-op under the `risc0-release` and `openvm-release` profiles which set `panic = "abort"`. To cover that case, the dominant panic source — `Mutex::lock().unwrap()` on a poisoned mutex — is routed through a new `MutexExt::lock_recover` extension that accepts the poison and continues. Other `.unwrap()` sites are either at startup before any FFI call (`tokio::runtime::Builder::build`) or test-only.
- **Unsynchronised `static mut` globals + hardcoded 3-network ceiling.** The family of `SWARM_STATE0..2`, `ZIG_HANDLER0..2`, `SHUTDOWN_NOTIFY0..2`, plus the `NETWORK_READY_SIGNALS` 3-tuple, all replaced with a single `Mutex<HashMap<u32, NetworkSlot>>`. The Swarm itself no longer lives in a global — `start_network` returns it and `run_eventloop` takes it as a parameter.
- **`stop_network` race.** Drops the per-network command sender BEFORE signalling the shutdown notify, so FFI publish/RPC sites racing shutdown see a closed channel and bail cleanly instead of enqueueing work that will never drain.
- **Null-pointer UB in CStr / byte-slice inputs.** Every `*const c_char` and `*const u8 + len` FFI parameter (`publish_msg_to_rust_bridge`, `send_rpc_request`, `send_rpc_response_chunk`, `send_rpc_error_response`, `create_and_run_network`'s four string fields) is null-checked before any dereference. The `create_and_run_network_inner` null-validation branch no longer forwards null pointers back into Zig's `releaseStartNetworkParams` (which declares non-nullable args).
- **Tokio-required `delay_map` mutations were happening on the FFI thread (#838).** `HashMapDelay::insert` / `update_timeout` schedule a `tokio::time::sleep` and panic with *"there is no reactor running"* when called outside a Tokio runtime. Two `extern "C"` sites — `send_rpc_request` (`REQUEST_ID_MAP.insert`) and `send_rpc_response_chunk` (`RESPONSE_CHANNEL_MAP.update_timeout`) — were doing exactly that on Zig's libxev thread. Latent for a year; surfaced once `Node.refreshSyncFromPeers` (#820) and `blocksByRange` bulk sync (#824) started dispatching RPCs from the interval ticker. Under the prover `panic = "abort"` profiles `catch_unwind` cannot catch this, so the whole zeam process aborted mid-sync. Fix: both delay-map mutations are now performed inside the event loop's `SwarmCommand::SendRpcRequest` / `SwarmCommand::SendRpcResponseChunk` arms, where Tokio is live; the FFI body of `send_rpc_request` no longer touches the maps at all, and `send_rpc_response_chunk` keeps only the read-only lookup. Regression test (`test_send_rpc_request_does_not_panic_outside_tokio_runtime`) installs a sender on `COMMAND_SENDERS`, calls `send_rpc_request` from a freshly-spawned OS thread (no Tokio), and asserts non-zero id + that the command landed on the receiver — verified to fail on the un-fixed code with the exact production panic message at `delay_map-0.4.1/src/hashmap_delay.rs:94:46`.

Single file diff: `rust/libp2p-glue/src/lib.rs`.

## Test plan

- [x] `cargo test --release -p libp2p-glue` — 16/16 (10 original + 5 added on `main` since this branch forked + 1 new #838 regression)
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy -p libp2p-glue -- -D warnings` — clean
- [x] `zig fmt --check .` — clean
- [x] `zig build test` — all suites green
- [x] Confirmed the `#838` regression test reproduces the production panic message verbatim when reverted to the pre-fix code

## Out of scope (separate PRs from the same audit)

- Split 2183-line `lib.rs` into modules (`swarm.rs`, `event_loop.rs`, `ffi.rs`, `handlers.rs`).
- Peer scoring (gossipsub `PeerScoreParams`).
- Discv5 + ENR `attnets` (tracked under #816).
- Concurrent stress tests (loom / proptest).
- The unrelated STF `error.DuplicateAttestationData` wedge (#837) — same incident window on devnet but a separate code path in `pkgs/chain/src/...`. Not addressed here; tracked as its own issue.

Also closes #840 (same root cause as #838 — same `delay_map-0.4.1/src/hashmap_delay.rs:94:46` panic from FFI thread without a Tokio reactor; see https://github.com/blockblaz/zeam/issues/840#issuecomment-4397627112 for the audit).


zclawz:
## Adversarial review against `14e7e0d`

I reviewed each of the four claims in the PR description against the actual diff and validated the behaviour locally where it mattered. Two findings; one is blocking, the other is a follow-up.

### What I verified

- **All exported FFI entry points are wrapped in `catch_ffi`.** Counted 9 `#[no_mangle] pub extern "C"` functions in `rust/libp2p-glue/src/lib.rs` (`get_swarm_command_dropped_total`, `stop_network`, `wait_for_network_ready`, `create_and_run_network`, `publish_msg_to_rust_bridge`, `send_rpc_request`, `send_rpc_response_chunk`, `send_rpc_end_of_stream`, `send_rpc_error_response`); each goes through `catch_ffi`. ✅
- **`static mut` family is gone.** `git show main:rust/libp2p-glue/src/lib.rs | grep static.mut` shows 6 `static mut` slots + the `0/1/2 => …, _ => panic!` ceiling on `main`; the new file has zero `static mut`s and replaces them with `lazy_static! { static ref NETWORK_SLOTS: Mutex<HashMap<u32, NetworkSlot>> }`. The 3-network ceiling is genuinely lifted. ✅
- **`stop_network` ordering.** Sender is removed, receiver is removed, then the shutdown notify is fired — in that order. Note the receiver-removal is a no-op once `run_eventloop` has taken ownership of it (it does `COMMAND_RECEIVERS.lock().remove(&self.network_id)` itself), but that's harmless; the comment "removes the receiver too — its drop signals to the event loop" is slightly misleading because the eventloop holds the receiver locally, but functionally fine. ✅
- **Null-pointer guards in CStr / byte-slice FFI inputs.** Verified for `publish_msg_to_rust_bridge` (topic + message), `send_rpc_request` (peer_id + request_data), `send_rpc_response_chunk` (response_data), `send_rpc_error_response` (message_ptr), and `create_and_run_network_inner` (all four string fields). ✅
- **`cargo test --release` → 10/10.** Reproduced locally on this branch. ✅

### 🔴 Blocking — `catch_unwind` is a no-op under the prover release profiles

This is the central UB-avoidance claim of the PR, and it does not hold for the artifacts the Zig side actually links in production.

`rust/Cargo.toml` defines two profiles that explicitly set `panic = "abort"`:

```toml
[profile.openvm-release]
inherits = "release"
…
panic = "abort"      # Remove unwinding code for smaller binary

[profile.risc0-release]
inherits = "release"
…
panic = "abort"      # Remove unwinding code for smaller binary
```

`build.zig` uses these for the OpenVM and RISC0 builds:

```
"--profile", "risc0-release",  -p zeam-glue …
"--profile", "openvm-release", -p zeam-glue …
```

I confirmed on this branch that `cargo build --profile risc0-release -p libp2p-glue -vv` invokes `rustc … -C panic=abort`. With `-C panic=abort`, `std::panic::catch_unwind` becomes effectively a no-op — the panic runtime calls `abort()` directly, no unwinding occurs, `catch_unwind` never gets the chance to translate the panic into `Err(_)` and `catch_ffi` never returns `None`.

To make sure I wasn't misreading the docs, I built a minimal repro under the same nightly toolchain (`rustc 1.97.0-nightly`) with a `panic = "abort"` profile:

```rust
fn main() {
    let r = std::panic::catch_unwind(|| { panic!("boom"); });
    println!("recovered: {:?}", r.is_err());
}
```

```
thread 'main' (15570) panicked at src/main.rs:3:9:
boom
note: run with `RUST_BACKTRACE=1` …
Aborted (core dumped)
EXIT=134
```

i.e. the process aborts with SIGABRT before `catch_unwind` returns. The "recovered: true" line never prints.

Net effect on this PR: in the **risc0** and **openvm** prover builds (which is `prebuilt-image`-shipped devnet binary), every panic that originates anywhere inside an FFI call — `lock().unwrap()` on a poisoned `Mutex`, an out-of-range index, a libp2p builder `.unwrap()`, etc. — will still abort the entire zeam process and unwind across the FFI boundary the same way it did before. The 10 new tests pass because `cargo test --release` uses the default `release` profile which inherits the workspace default of `panic = "unwind"`, but the behaviour they assert is **not** the behaviour the Zig binary will see when something panics in production.

Three options, in increasing order of effort:

1. **Honest scope reduction.** Keep `catch_ffi` as belt-and-suspenders for the dummy-prover/dev path (it is genuinely useful there) and update the doc-comment + PR description to say so explicitly: *"under `panic = abort` profiles this is a no-op; the goal there is for `panic-abort` to convert UB into a deterministic process abort, which it already does."* That at least stops a future reader from believing they have a defence they don't.

2. **Set `panic = "unwind"` on the libp2p crate (or zeam-glue) only.** Cargo enforces a single panic strategy per final binary, so this only works if you don't mind the rest of the prover code also unwinding (i.e. losing the binary-size benefit of abort across the whole process). Probably not desirable for the prover binaries.

3. **Replace `.unwrap()` and arithmetic that could panic with explicit error returns** at every FFI entry point so panics simply cannot originate inside the wrapped closure. That's the only fix that genuinely buys what the PR description claims under `panic = abort`. Some `.unwrap()`s on poisoned mutexes (e.g. `NETWORK_SLOTS.lock().unwrap()`, `COMMAND_SENDERS.lock().unwrap()`, `RESPONSE_CHANNEL_MAP.lock().unwrap()`, `REQUEST_ID_MAP.lock().unwrap()`) are particularly load-bearing — once one mutex is poisoned, every subsequent FFI call into that area aborts the process.

Either way I think this needs to be addressed before merge — ideally option (3) on the hottest paths plus a realistic comment on `catch_ffi` itself, otherwise option (1) at minimum so the PR description is accurate.

### 🟡 Follow-up — `release_params()` forwards null pointers into Zig that assumes non-null

In `create_and_run_network_inner`, the new null-pointer check is:

```rust
if local_private_key.is_null()
    || listen_addresses.is_null()
    || connect_addresses.is_null()
    || topics_str.is_null()
{
    logger::rustLogger.error(network_id, "create_and_run_network: null pointer in CreateNetworkParams string fields");
    release_params();   // <-- forwards the nulls back into Zig
    return;
}
```

`release_params` calls into Zig:

```zig
export fn releaseStartNetworkParams(zig_handler: *EthLibp2p,
    local_private_key: [*:0]const u8,
    listen_addresses:  [*:0]const u8,
    connect_addresses: [*:0]const u8,
    topics:            [*:0]const u8,
) void {
    const listen_slice = std.mem.span(listen_addresses);
    zig_handler.allocator.free(listen_slice);
    …
}
```

`[*:0]const u8` is non-nullable on the Zig side, and `std.mem.span` walks the pointer looking for the sentinel — passing `null` in is UB / an immediate segfault when it dereferences. So the new defensive null-check in Rust converts what would have been a `CStr::from_ptr(null)` UB into a different UB on the cleanup path. In practice this should be unreachable (the Zig caller never hands us nulls), but the whole reason the check exists is for the case where the Zig side has a bug — and in that case we now crash inside `releaseStartNetworkParams` instead of inside `CStr::from_ptr`.

Two simple fixes, either is fine:
- Skip `release_params()` on the null-pointer branch — leak the (presumably also null) buffers, log loudly. The Zig allocator wasn't going to be able to free a null pointer anyway.
- Or null-check each pointer inside `releaseStartNetworkParams` on the Zig side and skip the `free` for that field.

I'm OK with this being a follow-up rather than a blocker — but please file it before merge.

### Suggested next step

If you take option (3) above on the high-value paths (mutex `.unwrap()` and the libp2p `.unwrap()` chain in `Behaviour::new` / `build_transport`), I'm happy to re-review. Otherwise, please soften the PR description and the `catch_ffi` doc-comment so they accurately describe what you get under `panic = abort`, and I'll approve the scope-reduced version.

I'm explicitly **not** approving until at least the doc/PR-description fix lands.

Tested on `14e7e0dc95088567e9ab614e7b9d78d1db2da5ed`:
- `cargo test --release -p libp2p-glue` → 10 passed
- `cargo build --profile risc0-release -p libp2p-glue -vv` → confirms `-C panic=abort`
- minimal `catch_unwind` reproducer under `panic = "abort"` → process aborts, `catch_unwind` does not run

— @zclawz (review delegated by @ch4r10t33r)

ch4r10t33r:
@zclawz thanks for the deep review — both findings are valid. Addressed in `a05c85a`.

**Blocking — `catch_unwind` no-op under `panic = \"abort\"`**

You're right: `cargo build --profile risc0-release -p libp2p-glue -vv` does emit `-C panic=abort`, and the prover binaries shipped via `prebuilt-image` therefore wouldn't get any benefit from `catch_unwind`. Took a hybrid of options 1 and 3:

- Added a `MutexExt::lock_recover` extension trait and bulk-replaced every `.lock().unwrap()` (29 sites) with `.lock_recover()`. That converts the dominant panic source — `lock()` on a poisoned mutex — into a guard that just reuses the inner data. None of the Mutex-protected items in this crate (HashMaps of channel senders, request-id maps, response-channel maps, reconnect queues, the `NetworkSlot` map) have invariants that a mid-update panic would invalidate, so accept-and-continue is the right semantic.
- Applied the same recovery pattern to the `NETWORK_READY_CONDVAR.wait_timeout(...)` site that was the one remaining `.unwrap()` on a `LockResult` outside of `Mutex::lock`.
- Updated `catch_ffi`'s doc-comment to spell out the `panic = \"abort\"` caveat and what `lock_recover` does instead.

The `tokio::runtime::Builder::build().unwrap()` site is genuinely startup-only (before any Zig FFI call has registered the network). I left it as-is for now; if you'd like that covered too I can return `Result` from `create_and_run_network_inner` and surface a logged error to Zig.

**Follow-up — `release_params()` forwarding nulls into Zig**

Also fixed in the same commit. The null-validation branch in `create_and_run_network_inner` now logs and returns *without* calling `releaseStartNetworkParams`. Reasoning matches your suggestion: Zig declares those params as `[*:0]const u8` (non-nullable), `std.mem.span` would UB on null, and the branch is already unreachable in practice — leak + log is the safer recovery for the impossible case.

**Verification on `a05c85a`:**
- `cargo test --release -p libp2p-glue` → 10/10
- `cargo fmt --check` → clean
- `zig build test` → 151/151
- `zig build simtest` → same pre-existing `admin aggregator endpoint` flake as main (re-ran on a clean checkout of main at the same commit; both failed identically on that one integration test, all other suites passed).

PR description also updated to reflect what `catch_ffi` does and doesn't cover under each profile.

zclawz:
## Re-review of `a05c85a` — partial fix, one issue remains

Verified locally on `fix/libp2p-glue-critical-hardening` @ `a05c85a`.

### ✅ release_params null-forwarding (the 🟡 finding) — fixed
`create_and_run_network_inner`'s null branch no longer calls `release_params()`, with an inline comment explaining why (`releaseStartNetworkParams` declares `[*:0]const u8`, non-nullable). Leak-and-log is the right call here. 👍

### ✅ Tests + format
- `cargo test --release` (libp2p-glue) → 10/10 ✅
- `cargo fmt --check` → clean ✅

### ⚠️ `lock_recover` rollout is incomplete — 16 sites missed

The commit message says *"Bulk-replaces every `.lock().unwrap()` in the crate (29 sites)"*. Every **single-line** `.lock().unwrap()` was indeed converted (0 remaining), but the bulk replacement missed **16 multi-line sites** where `.lock()` and `.unwrap()` are on separate lines. Verified with:

```
python3 -c "import re; print(len(list(re.finditer(r'\.lock\(\)\s*\.unwrap\(\)', open('rust/libp2p-glue/src/lib.rs').read()))))"
# → 16
```

Sample of the missed sites (all production FFI-hot paths, not tests):

| Line | Function | Mutex |
|---|---|---|
| 174 | `get_zig_handler` | `NETWORK_SLOTS` |
| 195 | `get_shutdown_notify` | `NETWORK_SLOTS` |
| 208 | `is_network_ready` | `NETWORK_SLOTS` |
| 835 | `send_rpc_request_inner` | `REQUEST_PROTOCOL_MAP` |
| 1120 | reconnect path | `RECONNECT_ATTEMPTS` |
| 1235, 1239 | network startup | `COMMAND_SENDERS`, `COMMAND_RECEIVERS` |
| 1296 | inbound RPC dispatch | `REQUEST_PROTOCOL_MAP` |
| 1346, 1370, 1458, 1594 | swarm event loop | `RECONNECT_ATTEMPTS` |
| 1482 | swarm event loop | `CONNECTION_DIRECTIONS` |
| 1746, 1807 | response chunking | `REQUEST_PROTOCOL_MAP` |
| 1794 | response chunking | `RESPONSE_CHANNEL_MAP` |

These are exactly the paths that fire **inside** FFI handlers, so under `risc0-release`/`openvm-release` (`panic = "abort"`) a poisoned mutex at any of these 16 sites still aborts the prover binary — which is the failure mode this PR is supposed to eliminate.

Suggested follow-up commit: convert the remaining 16 sites to `.lock_recover()`. The grep above gives you a complete list. After that, `grep -nE '\.lock\(\)\s*$' rust/libp2p-glue/src/lib.rs` should return nothing followed by `.unwrap()` on the next non-blank line.

Once those are converted, this is good to merge from my side.

zclawz:
Thanks for the thorough response @ch4r10t33r — the `lock_recover` approach for the mutex sites makes sense given none of those protected types have invalidation-sensitive invariants. The null-guard fix on `release_params` looks correct too.

For the `tokio::runtime::Builder::build().unwrap()` site — leaving it as-is is fine; startup-only panic is acceptable there and propagating a `Result` just adds noise.

LGTM on the `a05c85a` fixes 👍

ch4r10t33r:
@zclawz fixed in `b0e2359`. The previous commit's bulk replacement was a single-line sed; missed every multi-line variant exactly as you showed. Re-ran with a regex that matches `\.lock\(\)\s*\.unwrap\(\)` (whitespace-aware), then `cargo fmt` to clean up.

Verified zero remaining sites:

```
python3 -c "import re; print(len(re.findall(r'\.lock\(\)\s*\.unwrap\(\)', open('rust/libp2p-glue/src/lib.rs').read())))"
0
```

41 `.lock_recover()` callsites in total now. Covers every map you listed: `NETWORK_SLOTS`, `REQUEST_PROTOCOL_MAP`, `COMMAND_SENDERS`, `COMMAND_RECEIVERS`, `RECONNECT_ATTEMPTS`, `CONNECTION_DIRECTIONS`, `RESPONSE_CHANNEL_MAP`.

`cargo test --release -p libp2p-glue` 10/10, `cargo fmt --check` clean, `zig build test` 151/151.

zclawz:
Thanks @ch4r10t33r — 41 sites converted, zero remaining `.lock().unwrap()` calls, all tests green.

Reviewed `b0e2359`: the whitespace-aware regex catch and `cargo fmt` cleanup is exactly the right approach. `lock_recover()` coverage across all seven maps (`NETWORK_SLOTS`, `REQUEST_PROTOCOL_MAP`, `COMMAND_SENDERS`, `COMMAND_RECEIVERS`, `RECONNECT_ATTEMPTS`, `CONNECTION_DIRECTIONS`, `RESPONSE_CHANNEL_MAP`) is complete.

No further concerns — PR is ready for merge from my side. ✅

zclawz:
## Re-verification of `b0e2359` — fixes confirmed ✅

Verified locally on the latest tip of `fix/libp2p-glue-critical-hardening` (`b0e2359cc0e954de437ad5a7c79509319a23f544`, fast-forwarded from `a05c85a`).

### What I checked

| Check | Result |
|---|---|
| `.lock().unwrap()` sites remaining | **0** (confirmed with whitespace-aware regex `\.lock\(\)\s*\.unwrap\(\)`) |
| `.lock_recover()` callsites | **41**, covering `NETWORK_SLOTS`, `REQUEST_ID_MAP`, `REQUEST_PROTOCOL_MAP`, `RESPONSE_CHANNEL_MAP`, `COMMAND_SENDERS`, `COMMAND_RECEIVERS`, `RECONNECT_QUEUE`, `RECONNECT_ATTEMPTS`, `CONNECTION_DIRECTIONS` |
| Every line I called out in the previous review (174, 195, 208, 835, 1120, 1235, 1239, 1296, 1346, 1370, 1458, 1482, 1594, 1746, 1794, 1807) | All converted to `lock_recover` |
| `MutexExt::lock_recover` impl | `Err(poisoned) => poisoned.into_inner()` — the right semantic for these maps (no invariants that a mid-update panic could violate) |
| `catch_ffi` doc-comment | Now includes the `panic = "abort"` caveat and explains the role of `lock_recover` as the production-build mitigation |
| `release_params()` on the null branch | Removed; comment explains the Zig-side `[*:0]const u8` non-nullability |
| `cargo test --release -p libp2p-glue` | **10 / 10 passed** |
| `cargo fmt --check` | clean |
| `cargo build --profile risc0-release -p libp2p-glue` | clean (panic=abort confirmed via `-vv`) |
| `cargo build --profile openvm-release -p libp2p-glue` | clean |

CI on `b0e2359` is green where it has reported so far: `lint`, `build` (macos+ubuntu), `test (macos-latest)`, `Dummy prove` (macos+ubuntu), `build-all-provers (macos-latest)` — that last one is the meaningful one for the panic-abort path. `build-all-provers (ubuntu-latest)` and `test (ubuntu-latest)` were still running when I finished verifying; assuming they land green, no concerns from my side.

### Tiny nit (non-blocking)

In the `MutexExt` doc-comment:

> Using `lock_recover` everywhere a previous version called `.lock_recover()` lets us avoid the dominant source of panic in this crate

Should be `.lock().unwrap()` (not `.lock_recover()`). Self-referential doc-bug only — ignore or fold into the next touchup.

### Decision

**LGTM 👍** Both findings from the original review (`catch_unwind` no-op under `panic = "abort"`, and `release_params` null-forwarding) are correctly addressed. The PR description and `catch_ffi` doc now accurately describe what the abstraction does and doesn't cover under each profile, and the `lock_recover` rollout is complete.

I'm leaving this as a comment rather than a formal "Approve" review, in line with our standing convention that bots don't gate consensus-critical merges — any owner approval still required for the branch protection rule.

— @zclawz (review delegated by @ch4r10t33r)

rust/libp2p-glue/src/lib.rs
```diff
@@ -355,35 +391,35 @@ pub unsafe extern "C" fn stop_network(network_id: u32) {
...
+            }
+            let now = std::time::Instant::now();
+            if now >= deadline {
+                return false;
+            }
+            let remaining = deadline - now;
+            let (g, timeout_result) = match NETWORK_READY_CONDVAR.wait_timeout(guard, remaining) {
+                Ok(result) => result,
+                Err(poisoned) => poisoned.into_inner(),
+            };
```
anshalshukla, line 511: Why are we using a poisoned guard here? Since we will have different network_ids it will be poisoned so why not have a separate condvar for each network id. @zclawz

zclawz: Good question! Two things going on here:

**1. Why `lock_recover()` on the dummy?**

The `dummy` mutex is a **local stack variable** — created fresh on every call to `wait_for_network_ready` and never shared with any other thread. A mutex can only become poisoned if another thread panics while holding it. Since no other thread ever holds this mutex, it can never actually be poisoned. `lock_recover()` here is purely defensive/idiomatic consistency with the rest of the codebase (the PR establishes `lock_recover()` as the universal way to acquire any mutex, whether or not it's actually at risk of poisoning).

**2. Why a single shared `NETWORK_READY_CONDVAR` instead of per-network condvars?**

The single condvar approach is intentional: `notify_all()` wakes all waiting callers, and each checks `is_network_ready(network_id)` for their own slot. This is a classic "broadcast condvar" pattern — correct and simple, just with potential spurious wake-ups when multiple networks exist simultaneously.

Per-network condvars (e.g. one `Condvar` per `NetworkSlot`) would eliminate spurious wakes, but since `wait_for_network_ready` is called at startup and not in a hot path, the overhead is negligible.

Happy to refactor to per-network condvars if you'd prefer — can do it inline here or as a follow-up.

g11tech (dismissed):

rust/libp2p-glue/src/lib.rs
```diff
@@ -600,10 +676,20 @@ pub unsafe extern "C" fn create_and_run_network(params: *const CreateNetworkPara
 
     rt.block_on(async move {
         let mut p2p_net = Network::new(network_id, zig_handler);
-        p2p_net
+        let swarm = p2p_net
             .start_network(local_key_pair, listen_multiaddrs, connect_multiaddrs)
             .await;
-        p2p_net.run_eventloop().await;
+        if let Some(swarm) = swarm {
+            p2p_net.run_eventloop(swarm).await;
```
anshalshukla, line 685: we mark network as ready in `start_network` and `run_eventloop` post that. `install_shutdown_notify` is called in `run_eventloop` so if in a scenario a shutdown is issued before the run_eventloop is completed than it will lead to unintended results

rust/libp2p-glue/src/lib.rs
```diff
@@ -2195,6 +2374,56 @@ mod tests {
...
+        // calls in. Under the panic="abort" prover profiles this aborts the
+        // whole process; here it would surface as a panicked test thread.
+        //
+        // We install a per-network command sender (so the function actually
+        // gets past `get_command_sender` and reaches the formerly-panicking
+        // code path) and call `send_rpc_request` from a freshly-spawned
+        // OS thread that is guaranteed to have no Tokio context attached.
+        // The fix moves both delay-map inserts to the event-loop side, so
+        // this call must now succeed (non-zero id, no panic).
+        let network_id = 101;
```
anshalshukla: we should use different network_id in all the tests as `cargo test` runs tests in parallel and there can be conflicts between the test

rust/libp2p-glue/src/lib.rs
ch4r10t33r: Good catch — the race is real. Trace:

`start_network` ends with `mark_network_ready(...)` while `shutdown_notify` is still `None` on the slot — `run_eventloop` installed it. A `stop_network` racing the `start → eventloop` handoff (e.g. Zig's `wait_for_network_ready` unblocks → caller hits Ctrl-C immediately) sees `ready == true` but `get_shutdown_notify == None`, so its `notify_one()` is silently dropped.

The cmd_rx-remove guard at the top of `run_eventloop` keeps this from hanging the thread on the *first* stop (the loop short-circuits because `cmd_rx` is gone), but the slot is left in a half-cleaned `ready=true / no event loop` state that subsequent start/stop cycles on the same `network_id` can observe.

Fixed in 80c3a30 by moving `install_shutdown_notify` into `start_network`, just before `mark_network_ready`. The invariant `ready ⇒ shutdown_notify present` now holds atomically from any concurrent observer's point of view; `run_eventloop` just picks the handle up via `get_shutdown_notify`. `notify_one` stores a permit if no waiter is parked yet, so a `stop_network` landing between the install and the first `notified().await` is observed on the first poll.

Added `test_shutdown_notify_installed_before_mark_ready` exercising the ordering directly + the permit-before-waiter property. Full `libp2p-glue` suite (17 tests) green.

rust/libp2p-glue/src/lib.rs
ch4r10t33r: Good catch — there were two real flake-prone collisions hidden in the existing ids:

- `network_id = 101`: `test_send_rpc_request_does_not_panic_outside_tokio_runtime` installs a sender, `test_publish_msg_to_rust_bridge_returns_false_when_uninitialized` asserts no sender.
- `network_id = 104`: `test_swarm_command_full_channel_drops_and_counts` installs a sender, `test_subscribe_gossip_topic_to_rust_bridge_returns_false_when_uninitialized` asserts no sender.

Either pair would flake depending on scheduler order.

Fixed in `ec10781`: introduced a `TEST_NETWORK_ID_COUNTER` (`AtomicU32` starting at 1000) and replaced every hardcoded `network_id` in the test module with `next_test_network_id()`. New tests just call the helper and can't possibly collide with anything else, including each other. Ran the suite 5× in a row — 17/17 every run.

anshalshukla (approved):

PR #820, opened 2026-05-04
node,network: drop BeamNode.mutex + BlockCache wiring (slice a-3 of #803)

zclawz:
## Scope

Slice **(a-3)** of the BeamNode threading refactor (#803) — drops the coarse `BeamNode.mutex` from every libxev/libp2p hot callback and wires the slice (a-2) primitives (`LockedMap`, `BlockCache`, `LockTimer`, `BorrowedState`) into `Network` + `BeamChain.connected_peers`. Builds on PR #805 (slice a-2).

Design doc: [`docs/threading_refactor_slice_a.md`](https://github.com/blockblaz/zeam/blob/main/docs/threading_refactor_slice_a.md)
Tracking: #803

### Network — every map gets its own per-resource lock

- `pending_rpc_requests` → `LockedMap(u64, PendingRPCEntry)`
- `pending_block_roots` → `LockedMap(types.Root, u32)`
- `timed_out_requests` → `ArrayList(u64)` + dedicated `timed_out_requests_lock`. `getTimedOutRequests` now returns an **owned slice** the caller frees, so the lock is never held across the retry loop.
- `fetched_blocks` / `fetched_block_ssz` / `fetched_block_children` → consolidated into a single `BlockCache` field under `block_cache_lock`. `BlockCache` gets two new helpers for the network's "block + parent link, ssz attached later" shape:
  - `insertBlockPtr(root, *SignedBlock, parent_root)` — atomic block + children-list update.
  - `attachSsz(root, []u8)` — separate ssz update; errors `BlockNotCached` if the root was pruned in between.
  - Plus `getBlock`, `getSsz`, `getChildrenCopy`, `forEachBlock`, `hasChildren` so external callers no longer reach into the underlying maps.
- `connected_peers` → new `ConnectedPeersImpl(PeerInfo)` wrapper around `StringHashMap + RwLock + std.atomic.Value(usize)`. Logger fast-path reads the atomic; iterators take the shared lock; adds/removes take the exclusive lock and update the atomic alongside the map mutation.

`BeamChain.connected_peers` now takes `*ConnectedPeers` (was `*const StringHashMap(PeerInfo)`). `count()` is lock-free; `iterateLocked()` provides the few cross-thread iteration sites the chain still needs.

### BeamNode — mutex dropped from the hot paths

The chain's per-resource locks (slice a-2) and the network's per-resource locks (this PR) make the outer mutex redundant on the gossip / interval / req-resp paths. The remaining lock is renamed and demoted:

| Callback | Before (a-2) | After (a-3) |
|---|---|---|
| `onGossip` | `BeamNode.mutex` | per-resource only (chain.\* + network.\*) |
| `onReqRespResponse` | `BeamNode.mutex` | snapshot pending entry → chain.onBlock with no outer lock |
| `onReqRespRequest.blocks_by_root` | `BeamNode.mutex` | **fully lock-free** (DB-internal sync + forkchoice RwLock shared) |
| `onReqRespRequest.status` | (none) | unchanged — already used `chain.getStatus()` |
| `onInterval` | `BeamNode.mutex` (around the whole inner loop) | per-resource only |
| `BeamNode.mutex` | sole synchroniser | **renamed to `finalization_lock`**; no users on hot callbacks; reserved for future multi-resource paths (e.g. an off-IO `processFinalizationFollowup`). |

`acquireMutex` / `MutexGuard` are renamed to `acquireFinalizationLock` / `FinalizationLockGuard`. Wait+hold metrics flow through `LockTimer` so the legacy `zeam_node_mutex_{wait,hold}_time_seconds{site=...}` series keeps getting double-emitted for one more release.

### `batch_pending_parent_roots` gets its own lock

Previously it relied on `BeamNode.mutex` for all access. With the global mutex gone, both the libxev tick path (`flushPendingParentFetches` after `processPendingBlocks`) and the libp2p bridge path (`cacheBlockAndFetchParent`) can now contend on this map; a dedicated `batch_pending_parent_roots_lock` is held only across the `put` / drain critical section.

### `handleReqRespResponse` snapshot pattern

The old code dereferenced a `*PendingRPCEntry` borrowed from the map and dispatched into `processBlockByRootChunk` (which calls `chain.onBlock`) with the borrow still live. Under per-resource locks that would mean holding `pending_rpc_requests` across the entire STF window. The new `Network.snapshotPendingRequest(request_id)` copies the per-request strings + roots out under the lock, then the response dispatch runs against the snapshot. `sweepTimedOutRequests` uses the same primitive.

## Callsite migration table

| File | Site | Before | After |
|---|---|---|---|
| `network.zig` | `Network.connected_peers` | `*StringHashMap(PeerInfo)` | `*ConnectedPeers` (= `ConnectedPeersImpl(PeerInfo)`) |
| `network.zig` | `selectPeer()` | borrowed `?[]const u8` | owned `?[]u8` (caller frees) |
| `network.zig` | `ensureBlocksByRootRequest()` | `BlocksByRootRequestResult{ peer_id: []const u8 }` | `BlocksByRootRequestResult{ peer_id: []u8 }` + `.deinit(allocator)` |
| `network.zig` | `getTimedOutRequests()` | borrowed `[]const u64` | owned `[]u64` (caller frees) |
| `network.zig` | `getPendingRequestPtr()` | returns `?*PendingRPCEntry` | replaced by `snapshotPendingRequest(id)` returning owned `PendingRequestSnapshot` |
| `network.zig` | `getChildrenOfBlock()` | borrowed slice | owned `[]types.Root` (caller frees) |
| `network.zig` | `cacheFetchedBlock()` | direct map insert | `BlockCache.insertBlockPtr` + cleanup |
| `network.zig` | `pruneCachedBlocks()` | walked maps directly | uses `BlockCache.getBlock` / `getChildrenCopy` |
| `network.zig` | block-collection sweeps | direct map iteration | `collectCachedBlocksAtOrBelowSlot()` / `collectReadyCachedBlocks()` (snapshot-then-process) |
| `chain.zig` | `connected_peers.iterator()` | direct | `iterateLocked()` (RAII shared-lock guard) |
| `chain.zig` | `connected_peers.count()` | hashmap-internal | atomic load (`ConnectedPeers.count_atomic`) |
| `node.zig` | `acquireMutex` / `MutexGuard` | global serialise | renamed to `acquireFinalizationLock` / `FinalizationLockGuard`, no users on hot callbacks |
| `node.zig` | `onGossip` / `onReqRespResponse` / `onReqRespRequest.*` / `onInterval` | took `BeamNode.mutex` | dropped |
| `node.zig` | `batch_pending_parent_roots` | covered by `BeamNode.mutex` | own `batch_pending_parent_roots_lock` |
| `node.zig` | `handleReqRespResponse` | borrow PendingRPCEntry across `chain.onBlock` | snapshot via `Network.snapshotPendingRequest` |
| `node.zig` | `pruneCachedBlocksCallback` / `processReadyCachedBlocks` | direct map iter | `Network.collectCachedBlocksAtOrBelowSlot` / `collectReadyCachedBlocks` |
| `node.zig` | `processCachedDescendants` | borrowed children slice | owned `getChildrenOfBlock` slice + free |
| `node.zig` | `refreshSyncFromPeers` | iterated `connected_peers` map directly | snapshot peer ids under shared lock, then dispatch |

## Tests

Unit tests in `pkgs/node/src/locking.zig`:
- `LockedMap: concurrent put/get/remove smoke` — 4 threads × 256 ops each, checks no deadlock and bounded final count.
- `ConnectedPeers: connect / disconnect / count atomic` — connect/disconnect/replace/iterator/selectPeerCopy semantics.
- `ConnectedPeers: concurrent connect/disconnect keeps count consistent` — 3 threads stress the atomic count + RwLock; final atomic count must equal the actual map size.

Integration tests added in `pkgs/node/src/node.zig`:
- `Network: BlockCache wiring smoke (slice a-3)` — exercises `cacheFetchedBlock` (insertBlockPtr path), duplicate handling, `getChildrenOfBlock`, `storeFetchedBlockSsz`, `collectCachedBlocksAtOrBelowSlot`, `collectReadyCachedBlocks`, `removeFetchedBlock` parent-link cleanup.
- `Network: ConnectedPeers integration with selectPeer (slice a-3)` — connectPeer / hasPeer / count / selectPeer (owned-copy) / disconnectPeer.

Existing tests preserved (cache pruning chain tests, peer connect/disconnect, full publish/onBlock flow).

### Stress test

The single-node ingestion stress harness called for in §"Stress test plan" of the design doc lives outside the unit-test pkg — it needs a real Network backend / mock libp2p / forked-chain seed and is best run as a `pkgs/sim` scenario. The unit-tested `LockedMap` / `ConnectedPeers` concurrency smokes here are the merge gate for the *lock-correctness* side; a separate sim run is the merge gate for the *throughput* side. Tracked in #803.

**Dev test run on this branch:** `zig build -Doptimize=Debug` and `zig build --summary all test -Doptimize=Debug` both green (see status comment for the latest run).

## Build status

```
$ zig fmt pkgs/        # clean
$ /opt/zig-x86_64-linux-0.16.0/zig build -Doptimize=Debug --summary all
Build Summary: 26/26 steps succeeded
$ /opt/zig-x86_64-linux-0.16.0/zig build --summary all test -Doptimize=Debug
EXIT:0  (all suites pass — params, network, configs, utils, database, api, xmss, types, state-transition, node, tools, …)
```

Run on zig 0.16.0 (matches the project's Zig pin after #784). `zig build` exit 0, `zig build test` exit 0. (Re-running on the new commit; will refresh the status comment.)

## What is NOT in this PR

- **Slice (b)** — split `events_lock` for `last_emitted_*` + `cached_finalized_state`, and the lifetime-tightening on the long-hold FFI paths. Slice (a-3) only relocates the lock boundary; it does not change `chain.zig` mutation semantics (out of scope per the slice-discipline rule in the design doc).
- **Slice (c)** — chain-worker thread + `processFinalizationFollowup` move-off-IO. `finalization_lock` survives as a labelled placeholder that slice (c) is expected to reach for; it has no users on the hot callbacks today.
- **`Arc<BeamState>` / refcounted state pointers** — slice (c) blocker; not introduced here.
- **HTTP API `/eth/v1/*`** — left at status quo per the design doc r3 ("no `/eth/v1/*` prototype branch").
- The sim-based ingestion stress harness (see "Stress test" above).

## Deviations / open questions

- `BlockCache` previously assumed a `block + ssz + parent-link` triple insert was atomic. The pre-slice network code stored ssz **separately** from the block (only `cacheFutureBlock` paired them). I matched the existing behaviour by adding `insertBlockPtr` (block + parent link, no ssz) + `attachSsz` (later, optional). The triple-`insert` API is preserved for any future caller that wants atomic ssz inclusion.
- `selectPeer` now returns an owned dup'd slice. The previous borrowed-slice contract was unsafe across the unlock and would have raced with `disconnectPeer`. Caller-frees keeps the boundary clean.
- `finalization_lock` has no users on the gossip / interval / req-resp paths today. It is *kept* (renamed, not deleted) because the design doc names it as the survivor for the small set of paths that legitimately need a multi-resource view; deleting it would force slice (c) to re-introduce the same primitive.

cc @ch4r10t33r @gr3999 @GrapeBaBa

### DIFF
--- a/pkgs/cli/src/api_server.zig
+++ b/pkgs/cli/src/api_server.zig
@@ -283,6 +283,11 @@ pub const ApiServer = struct {
             _ = request.respond("Not Found: Finalized checkpoint lean state not available\n", .{ .status = .not_found }) catch {};
             return;
         };
+        // assertReleasedOrPanic registered FIRST so it runs LAST (LIFO):
+        // by the time it runs, the deinit defer has already set
+        // `released = true`. Catches a future helper that bypasses
+        // deinit. PR #820 / issue #803.
+        defer finalized_borrow.assertReleasedOrPanic();
         defer finalized_borrow.deinit();
 
         // Serialize lean state (BeamState) to SSZ
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -365,13 +365,32 @@ pub const BeamChain = struct {
     }
 
     /// Insert under `states_lock.exclusive` if the entry is not already in
-    /// the map; otherwise keep the existing pointer untouched. Returns:
-    ///   * `effective` — the pointer the caller should USE for any
-    ///     subsequent reads (DB writes, forkchoice updates, etc.). When
-    ///     the entry already existed this is the in-map pointer, NOT the
-    ///     `state_ptr` argument; the in-map pointer outlives the borrow
-    ///     handed out by `statesGet`, so other readers don't observe a
-    ///     freed pointer.
+    /// the map; otherwise keep the existing pointer untouched. The exclusive
+    /// lock is HANDED OFF to the returned `BorrowedState` and is held until
+    /// the caller invokes `borrow.deinit()` — callers MUST keep the borrow
+    /// alive across any subsequent deref of `borrow.state` (DB writes,
+    /// forkchoice updates, etc.).
+    ///
+    /// Why hold the exclusive lock across the borrow? See PR #820 / issue
+    /// #803: with `BeamNode.mutex` removed from the gossip / req-resp /
+    /// interval paths, two threads can be inside `chain.onBlock`
+    /// concurrently. Thread A's `onBlockFollowup -> processFinalizationAdvancement
+    /// -> pruneStates` (also exclusive on `states_lock`) can `fetchRemove`
+    /// + free the very entry Thread B is about to deref for the post-commit
+    /// DB write + `forkChoice.confirmBlock`. Holding the exclusive lock
+    /// across that whole window is the fix — we cannot downgrade-then-
+    /// reacquire-shared because the unlock/reacquire gap is exactly the UAF
+    /// race we are closing. (A native RwLock downgrade primitive would let
+    /// us hold shared instead; that is a separate change.)
+    ///
+    /// Returns:
+    ///   * `borrow` — RAII wrapper. `borrow.state` is the pointer the
+    ///     caller should USE for any subsequent reads. When the entry
+    ///     already existed this is the in-map pointer, NOT the `state_ptr`
+    ///     argument; the in-map pointer outlives the borrow handed out by
+    ///     `statesGet`, so other readers don't observe a freed pointer.
+    ///     Caller MUST keep `borrow` alive until done dereferencing
+    ///     `borrow.state`, then call `borrow.deinit()`.
     ///   * `kept_existing` — true when the entry already existed and
     ///     `state_ptr` was discarded (caller is responsible for freeing
     ///     `state_ptr` if it owns it). False when `state_ptr` was inserted.
@@ -380,24 +399,39 @@ pub const BeamChain = struct {
         comptime site: []const u8,
         root: types.Root,
         state_ptr: *types.BeamState,
-    ) !struct { effective: *types.BeamState, kept_existing: bool } {
+    ) !struct { borrow: BorrowedState, kept_existing: bool } {
         var t = LockTimer.start("states", site);
         self.states_lock.lock();
         t.acquired();
+        // NOTE: hold-time histogram closes here at function return; the
+        // borrow's deinit-site is responsible for any per-callsite
+        // observation of the extended hold window. Mirrors the pattern in
+        // `statesGet`.
         defer t.released();
-        defer self.states_lock.unlock();
+        // NOTE: NO `defer self.states_lock.unlock()` here — the exclusive
+        // lock is owned by the returned BorrowedState and released by its
+        // deinit. errdefer below covers the OOM path on `getOrPut`.
+        errdefer self.states_lock.unlock();
         const gop = try self.states.getOrPut(root);
-        if (gop.found_existing) {
+        const effective_ptr: *types.BeamState = if (gop.found_existing) blk: {
             // Decision policy: keep the existing pointer (it's referenced
             // elsewhere — e.g. produceBlock just inserted it before
             // publishBlock landed here) and tell the caller to free the
             // freshly-computed copy. We never want to invalidate a pointer
             // that other borrows might still observe through
             // `states_lock.shared`.
-            return .{ .effective = gop.value_ptr.*, .kept_existing = true };
-        }
-        gop.value_ptr.* = state_ptr;
-        return .{ .effective = state_ptr, .kept_existing = false };
+            break :blk gop.value_ptr.*;
+        } else blk: {
+            gop.value_ptr.* = state_ptr;
+            break :blk state_ptr;
+        };
+        return .{
+            .borrow = BorrowedState{
+                .state = effective_ptr,
+                .backing = .{ .states_exclusive_rwlock = &self.states_lock },
+            },
+            .kept_existing = gop.found_existing,
+        };
     }
 
     /// Take the exclusive side of `states_lock` and remove the entry.
@@ -1364,21 +1398,31 @@ pub const BeamChain = struct {
         // already exists (locally produced block path), keep the existing
         // pointer and free the freshly-computed one — borrows already
         // observe the existing pointer through `states_lock.shared`, so
-        // overwriting would create a UAF window. Returns the new pointer
-        // when the existing one was kept; null when it was inserted.
-        const commit = try self.statesCommitKeepExisting("onBlock.commit", fcBlock.blockRoot, post_state);
+        // overwriting would create a UAF window.
+        //
+        // The exclusive lock is HELD across the post-commit deref window
+        // (DB write + forkchoice confirm) via the returned BorrowedState.
+        // This blocks `pruneStates` (also exclusive on `states_lock`) from
+        // racing in via another thread's `onBlockFollowup` and freeing the
+        // entry under us. See PR #820 / issue #803.
+        var commit = try self.statesCommitKeepExisting("onBlock.commit", fcBlock.blockRoot, post_state);
+        // Release the exclusive lock on every exit path (success or error).
+        defer commit.borrow.deinit();
         if (commit.kept_existing and post_state_owned) {
             // Existing entry kept — free the freshly-computed (and now
-            // redundant) post_state. Caller-supplied post-states (i.e.
-            // `post_state_owned == false`) belong to the caller; we don't
-            // touch them.
+            // redundant) post_state. The borrow points at the in-map
+            // pointer (a different allocation), so this free does not
+            // invalidate `commit.borrow.state`. Caller-supplied post-states
+            // (i.e. `post_state_owned == false`) belong to the caller; we
+            // don't touch them.
             post_state.deinit();
             self.allocator.destroy(post_state);
         }
-        // From here on use the effective pointer (the in-map one): if the
+        // From here on use the borrow's pointer (the in-map one): if the
         // commit kept an existing entry, our `post_state` is freed and
-        // unsafe to deref.
-        const effective_post_state: *types.BeamState = commit.effective;
+        // unsafe to deref. The borrow keeps the in-map pointer alive for
+        // the duration of this scope.
+        const effective_post_state: *const types.BeamState = commit.borrow.state;
         // Past this point post_state ownership is settled — either the
         // pointer is in the states map (insert path) or it was explicitly
         // freed above (existing-kept path). The top-level errdefer must
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -1801,7 +1801,10 @@ pub const BeamChain = struct {
         _ = is_from_gossip;
 
         const current_slot = self.forkChoice.fcStore.slot_clock.timeSlots.load(.monotonic);
-        const finalized_slot = self.forkChoice.fcStore.latest_finalized.slot;
+        // latest_finalized is a multi-field Checkpoint written under
+        // forkChoice.mutex (exclusive). Take the shared lock via the
+        // accessor to avoid a torn (slot, root) pair. PR #820 / #803.
+        const finalized_slot = self.forkChoice.getLatestFinalized().slot;
 
         // 1. Future slot check - reject blocks too far in the future
         // Allow a small tolerance for clock skew, but reject clearly invalid future slots
@@ -2062,7 +2065,11 @@ pub const BeamChain = struct {
     }
 
     pub fn aggregate(self: *Self) ![]types.SignedAggregatedAttestation {
-        const head_root = self.forkChoice.head.blockRoot;
+        // forkChoice.head is a multi-field ProtoBlock written under
+        // forkChoice.mutex (exclusive). Snapshot once via the shared-
+        // locked accessor; reading `.blockRoot` directly would tear
+        // against a concurrent updateHead. PR #820 / #803.
+        const head_root = self.forkChoice.getHead().blockRoot;
         // Snapshot-then-release: forkChoice.aggregate runs an FFI window
         // (~700ms) over `state.validators`. Holding `states_lock.shared`
         // for that window would force any STF commit to wait. Clone first,
@@ -2147,7 +2154,11 @@ pub const BeamChain = struct {
     /// PR description (a-2) enumerates every caller migrated under this
     /// API change — grepping `states.get` will not find them.
     pub fn getFinalizedState(self: *Self) ?BorrowedState {
-        const finalized_checkpoint = self.forkChoice.fcStore.latest_finalized;
+        // latest_finalized is a multi-field Checkpoint written under
+        // forkChoice.mutex (exclusive); use the shared-locked accessor
+        // so we don't pair `slot` with a different update's `root`.
+        // PR #820 / #803.
+        const finalized_checkpoint = self.forkChoice.getLatestFinalized();
 
         // First try the in-memory states map under `states_lock.shared`.
         if (self.statesGet(finalized_checkpoint.root)) |borrow| {
@@ -2234,10 +2245,12 @@ pub const BeamChain = struct {
         };
     }
 
-    /// Get the latest justified checkpoint
-    /// Returns the checkpoint with slot and root of the most recent justified checkpoint
+    /// Get the latest justified checkpoint.
+    /// Returns the checkpoint with slot and root of the most recent
+    /// justified checkpoint, snapshotted under forkChoice.mutex.lockShared
+    /// so callers see a coherent (slot, root) pair. PR #820 / #803.
     pub fn getJustifiedCheckpoint(self: *Self) types.Checkpoint {
-        return self.forkChoice.fcStore.latest_justified;
+        return self.forkChoice.getLatestJustified();
     }
 
     pub const SyncStatus = union(enum) {
@@ -2268,8 +2281,12 @@ pub const BeamChain = struct {
             return .no_peers;
         }
 
-        const our_head_slot = self.forkChoice.head.slot;
-        const our_finalized_slot = self.forkChoice.fcStore.latest_finalized.slot;
+        // forkChoice.head and fcStore.latest_finalized are multi-field
+        // structs written under forkChoice.mutex.lock(). Snapshot via the
+        // shared-locked accessors so callers see coherent values.
+        // PR #820 / #803.
+        const our_head_slot = self.forkChoice.getHead().slot;
+        const our_finalized_slot = self.forkChoice.getLatestFinalized().slot;
 
         // Find the maximum finalized slot reported by any peer
         var max_peer_finalized_slot: types.Slot = our_finalized_slot;
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -3303,9 +3303,11 @@ test "BlockCache: insert + get + removeChildrenOf bounded" {
 
     // Every inserted root should be retrievable.
     for (1..mock_chain.blocks.len) |i| {
-        const got = bc.getBlockAndSsz(mock_chain.blockRoots[i]);
-        try std.testing.expect(got != null);
-        try std.testing.expect(got.?.ssz != null);
+        const got_opt = try bc.cloneBlockAndSsz(mock_chain.blockRoots[i], std.testing.allocator);
+        try std.testing.expect(got_opt != null);
+        var got = got_opt.?;
+        defer got.deinit(std.testing.allocator);
+        try std.testing.expect(got.ssz != null);
     }
 
     // removeChildrenOf the genesis root should drop the entire chain.
@@ -3348,17 +3350,20 @@ test "BlockCache: partial-state invariant (re-insert leaves no orphans)" {
     // `ssz_bytes`, and the children list under the parent has the root
     // listed twice (since we appended on both inserts — documented
     // behaviour: `insert` does not de-dup, callers manage that). Either
-    // way `getBlockAndSsz` must return the latest entry.
-    const got = bc.getBlockAndSsz(mock_chain.blockRoots[1]);
-    try std.testing.expect(got != null);
-    try std.testing.expect(got.?.ssz != null);
+    // way `cloneBlockAndSsz` must return the latest entry.
+    const got_opt = try bc.cloneBlockAndSsz(mock_chain.blockRoots[1], std.testing.allocator);
+    try std.testing.expect(got_opt != null);
+    var got = got_opt.?;
+    defer got.deinit(std.testing.allocator);
+    try std.testing.expect(got.ssz != null);
     try std.testing.expect(bc.count() == 1);
 }
 
 test "BlockCache: insertBlockPtr+ssz atomic visibility" {
-    // Triple-atomic invariant: a reader using getBlockAndSsz must observe
-    // either both-Some or both-null, never a partial state. With the
-    // ssz-arg variant of insertBlockPtr the atomic case is the new path.
+    // Triple-atomic invariant: a reader using cloneBlockAndSsz must
+    // observe either both-Some or both-null, never a partial state.
+    // With the ssz-arg variant of insertBlockPtr the atomic case is
+    // the new path.
     var arena_allocator = std.heap.ArenaAllocator.init(std.testing.allocator);
     defer arena_allocator.deinit();
     const allocator = arena_allocator.allocator();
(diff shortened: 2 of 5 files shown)


zclawz:
**Status:** ✅ all green at head `a225cc2` after a substantial round of correctness fixes — thanks to @ch4r10t33r for the deep review. 11 bugs fixed across 11 commits.

## Bugs fixed (all credited to @ch4r10t33r)

### 1. UAF: `statesCommitKeepExisting` released `states_lock.exclusive` before caller derefed in-map pointer (commit `60761c9`)
With `BeamNode.mutex` dropped, two threads can be inside `chain.onBlock` concurrently. Thread A's `onBlockFollowup → processFinalizationAdvancement → pruneStates` (also exclusive on `states_lock`) was free to `fetchRemove`+free the very entry Thread B was about to deref via `updateBlockDb` + `forkChoice.confirmBlock`. Fixed by handing the exclusive lock off to the returned `BorrowedState` (new `states_exclusive_rwlock` backing variant) so the caller in `onBlock` keeps it alive across the post-commit deref window via `defer commit.borrow.deinit()`. Cost: pruner blocks for the `updateBlockDb` + `confirmBlock` window — acceptable for correctness.

### 2. Torn struct reads of `forkChoice.head` / `fcStore.latest_finalized` in `node.zig` (commit `6a07f74`)
Multi-field struct reads paired `slot` with a `blockRoot` from a different update. Migrated to existing `getHead()` / `getLatestFinalized()` helpers (which take `mutex.lockShared`) and snapshot once per logical decision block.

### 3. `BlockCache` atomic-triple invariant broken by split-insert (commit `ad59a9c`)
`insertBlockPtr` (no ssz) + later `attachSsz` left a window where `getBlock(root)` returned Some and `getSsz(root)` returned None. Fixed by adding an `ssz: ?[]u8` param to `insertBlockPtr` (atomic when non-null) and a new `getBlockAndSsz` returning a single coherent `BlockAndSsz` struct. Network gains `getFetchedBlockWithSsz`; `node.zig:processCachedDescendants` collapses two-lock dance into one call. `attachSsz` retained for the legacy block-first-ssz-later shape with documented partial-state window.

### 4. `removeFetchedBlock` TOCTOU between `getBlock` and `removeOne` (commit `45687c7`)
Two-step pattern leaked a window where another thread could remove the entry or its parent's children list, leaving `parent_root` stale. Fixed by adding `BlockCache.removeFetchedBlock(root)` doing the lookup + remove + parent-link cleanup all under one critical section. Network wrapper becomes a one-liner.

### 5. `ConnectedPeers.connect` non-atomic count + permanent inconsistency on error (commit `15bc0ef`)
Legacy `fetchSub` → dupe → put → `fetchAdd` shape would leak the count permanently low if either dupe failed (errdefer unwinds put but not the prior `fetchSub`). Fixed by treating `connect` as idempotent on the replace path: `count_atomic` is left untouched; only `fetchAdd(1)` on fresh insert.

### 6. `snapshotPendingRequest` O(n) walk → sweep O(n²) (commit `b110cbe`)
Replaced `iterateLocked + key compare` with `LockedMap.get(request_id)` — single hash lookup under the mutex.

### 7. `tier5_depth` decremented while `events_lock` still held (commit `4334a61`)
`chain.getFinalizedState`'s cache-hit and DB-load paths called `leaveTier5()` BEFORE returning the borrow, so HTTP / event-broadcaster paths silently bypassed the sibling-rule assertion for the entire borrow lifetime. Fixed by adding `tier5_held: bool` on `BorrowedState`; `deinit` calls `leaveTier5()` AFTER unlocking.

### 8. `assertReleased` never wired up — silent drop in Zig 0.16 (commit `1b2d580`)
Renamed to `assertReleasedOrPanic` with a richer panic message that names the backing variant. Wired up at every callsite in `chain.zig` and `api_server.zig` as:
```zig
defer borrow.assertReleasedOrPanic();   // registered FIRST, runs LAST (LIFO)
defer borrow.deinit();                    // registered SECOND, runs FIRST
```
On normal exit `deinit` runs first (sets `released=true`), then the assert validates. Catches future helpers that bypass `deinit`. Compiled out in release.

### 9. `cloneAndRelease` double-call is UAF (commit `1b2d580`, same commit as #8)
After the first call the lock is released; the in-map state may be freed by a concurrent thread. Added `std.debug.assert(!self.released)` at the top.

### 10. `BlockCache.get` returned synthetic zero `parent_root` (commit `0a9b00e`)
The helper claimed to return the cached triple but `parent_root` was always zero — callers that trusted the field were silently misled. Deleted the helper and the now-unused `CachedBlock` type. Existing test usages migrate to `getBlockAndSsz` (the honest replacement).

### 11. `LockTimer` hold-time wrong on borrow-handoff paths (commit `0a9b00e`, same commit as #10)
`statesGet` / `statesCommitKeepExisting` / `getFinalizedState` closed the LockTimer at the helper's return — but the borrow's true hold span extends to its `deinit` site. Fixed by adding a `timer: ?LockTimer` field on `BorrowedState`; the helper acquires + `t.acquired()`, then moves the timer into the borrow. `deinit` calls `t.released()` AFTER unlocking. Hold-span histograms now correctly attribute the full borrow lifetime.

### Bonus: same torn-struct-read sweep extended to `chain.zig` (commit `a225cc2`)
The bug-2 pattern existed in `chain.zig` too: `validateBlock`, `aggregate`, `getJustifiedCheckpoint`, `getSyncStatus`, `getFinalizedState`. Migrated each to the shared-locked accessors.

## Sweep before push
```
$ grep -nE "(getBlock|getSsz|removeOne|count_atomic|leaveTier5|enterTier5|forkChoice\.head|forkChoice\.fcStore\.latest_(finalized|justified))" pkgs/node/src/{node,network,chain,locking}.zig pkgs/cli/src/api_server.zig
```
Confirmed no remaining unguarded multi-field struct reads, iterator-walks where `get` suffices, tier-5 holes, or synthetic-parent uses outside init-time / single-threaded paths.

## Build & tests
```
$ zig fmt pkgs/                                                      # clean
$ /opt/zig-x86_64-linux-0.16.0/zig build -Doptimize=Debug            # EXIT:0
$ /opt/zig-x86_64-linux-0.16.0/zig build --summary all test \
    -Doptimize=Debug                                                  # EXIT:0
  Build Summary: 50/50 steps succeeded
  pkgs/node/src/lib.zig: 82/82 OK   (slice a-3 baseline 73 + 9 new)
```

## Lock-hold cost note (re bug 1)
The `states_lock.exclusive` window in `onBlock` grew by `updateBlockDb` + `forkChoice.confirmBlock`. Forkchoice has its own RwLock so `confirmBlock` is short; `updateBlockDb` is a serialised write to the DB layer. The pruner blocks for that window — acceptable cost for correctness. A native RwLock downgrade primitive would let us hold shared instead; that is a separate change.

Head: `a225cc2`. Branch: `feat/agent-slice-a-3-drop-node-mutex`. Ready for re-review.

zclawz:
### Bugs 12–14 (round 2 of post-review fixes) — landed on `feat/agent-slice-a-3-drop-node-mutex`

Three more bugs flagged by @ch4r10t33r have been fixed on top of the prior 11. Each is a separate commit on the existing branch. HEAD is now `961fa4a`.

| # | Bug | Commit | Summary |
|---|-----|--------|---------|
| 12 | `finalization_lock` is dead code (the rename adds review surface for nothing — zero callers in this PR) | `9250032` | Deleted `BeamNode.finalization_lock`, `acquireFinalizationLock`, the `FinalizationLockGuard` wrapper, and the related comment in `chain.zig`. Slice (c) (`processFinalizationFollowup` move-off-IO-thread) will reintroduce a multi-resource lock here when its first real user lands — with a real call site and metric labels chosen at that point. |
| 13 | `BlockCache.forEachBlock` long-hold-abuse footgun (handed `*const SignedBlock` map pointers; a callback could stash them past return and observe a rehash UAF — same shape as the `iterateLocked` issue from a-2) | `5aefe1c` | Changed signature to take `types.SignedBlock` by value (struct copy under the lock). Updated both callsites in `network.zig` (`pruneAtOrBelowEach`, `collectReadyEach`). Added a docstring documenting the lock-held-across-callback contract and the slice-fields-only-valid-inside-callback rule. Bounded by `MAX_CACHED_BLOCKS=1024` × `sizeOf(SignedBlock)` per sweep — small. |
| 14 | Concurrent test coverage still partial — needed unit-level repro for the original UAF surface so future regressions surface here without needing the sim package | `961fa4a` | Three new tests in `pkgs/node/src/chain.zig`:<br/>**A.** `BlockCache: 3-thread stress` — insert / read / remove with N=1500 iters; asserts triple-atomic invariant + no orphan parent-link entries after cleanup.<br/>**B.** `BorrowedState: cloneAndRelease vs concurrent statesFetchRemoveExclusivePtr` — start-barrier'd two-thread race over a real BeamChain with 64 unrelated states + 1000 cloneAndRelease iters; asserts coherent slot every time, no UAF.<br/>**C.** `chain.onBlock: two-thread concurrent import of same block` — start-barrier'd two-thread race calling `onBlock(blocks[1])` on a fresh BeamChain per iteration (30 iters, marked SLOW; STF + sig-verify dominates wall-clock). Exercises the `statesCommitKeepExisting` `kept_existing` path that the slice (a-2) `cloneAndRelease` and slice (a-3) hold-exclusive-until-deref fixes are guarding. |

### Build/test status

- `zig build -Doptimize=Debug` — `EXIT:0`
- `zig build --summary all test -Doptimize=Debug` — `EXIT:0`, all 85 chain/locking tests + every other test suite green.

### Deviations

- **Test C iter count.** Originally targeted 100; reduced to **30** because each iteration runs two parallel STF + XMSS verifies, and even with the existing test fixtures wall-clock would have crowded out the rest of the node test suite. The test still genuinely exercises both threads concurrently (start-barrier + ready-counter) and reproduces the UAF surface deterministically under `-Doptimize=Debug`. Documented inline; bump to 100 once an XMSS-bypass option lands.
- **Test B "unrelated" roots** are synthesized by encoding `k+1` into the first 8 bytes of the root (zeroed elsewhere) so they cannot collide with the genesis root. Each carries a fresh `sszClone` of the genesis state.

cc @ch4r10t33r — flagging for confirmation that bugs 12–14 are resolved as intended.

zclawz:
## Batch 2 follow-up — N1 + N3 + N2 tracking issue

Two more items from @ch4r10t33r's review on top of `961fa4a`. Pushed two new commits to this branch.

---

### N1 — UAF regression in `snapshotPendingRequest` (FIXED, present-tense bug) — credit @ch4r10t33r

`3aa3866` `fix(network): close UAF in snapshotPendingRequest by dupe'ing under LockedMap lock`

The batch-1 O(1) fix in commit [`60761c9`](https://github.com/blockblaz/zeam/pull/820/commits/60761c929eae471c2042d97399e4cdef3d708ec4) replaced the old `iterateLocked + key compare` scan with `pending_rpc_requests.get(request_id) + dupe`. `LockedMap.get` returns the value **by value** and releases the lock before returning — so the returned `PendingRPCEntry` struct's slice headers (`peer_id: []const u8`, `requested_roots: []types.Root`) still aliased the in-map allocator-owned storage. A concurrent `finalizePendingRequest(id)` could `fetchRemove` + `rpc_entry.deinit(allocator)` between `get` returning and the dupes running — UAF.

Fix shape (option (a) from the review):

- Added `LockedMap.withValueLocked(key, ctx, each)` in `pkgs/node/src/locking.zig`. Runs the callback while the map's mutex is still held; the callback receives `?*const V` pointing at the in-map storage and threads the snapshot back through `ctx`.
- Migrated `snapshotPendingRequest` to do **all** dupes inside that callback. No `allocator.dupe` calls outside the closure (verified by inspection + grep).
- Three new unit tests for the helper:
  - `LockedMap: withValueLocked holds the lock across the callback` — proves the lock is held during the callback via a second-thread `tryLock` that must report contended (same shape as the existing `BorrowedState` pruner-shaped test).
  - `LockedMap: withValueLocked invokes callback with null on missing key`.
  - `LockedMap: withValueLocked propagates callback errors` — confirms the lock is released on the error path.

**Audit for the same pattern in the codebase**: only two `LockedMap` types exist (`PendingRPCMap = LockedMap(u64, PendingRPCEntry)` — fixed here; `PendingBlockRootMap = LockedMap(types.Root, u32)` — value is plain `u32` with no slice fields, no symmetric bug).

---

### N3 — split `insertBlockPtr(null)+attachSsz` concurrent test (gap closed) — credit @ch4r10t33r

`42b4566` `test(locking): split insertBlockPtr(null)+attachSsz concurrent race coverage`

The existing 3-thread BlockCache stress test only exercises `insertBlockPtr(ssz=non-null)` (atomic triple in one call). Production `Network.cacheFetchedBlock` passes `ssz=null` and a later `storeFetchedBlockSsz` does `attachSsz` — that race surface (the bounded partial-state window where block-Some + ssz-null is observable) was documented but had no test.

Three-thread harness in `pkgs/node/src/locking.zig`:

- Thread A (writer): 1500 iters of `insertBlockPtr(ssz=null)` followed by `attachSsz`, distinct root per iter.
- Thread B (reader): spins on `getBlockAndSsz`, asserts the inner block slices are readable (`slot` range-checked; `parent_root == ZERO_HASH`) and SSZ is either `null` (transient) or the 4-byte fixture — never garbage. Counts `partial_observed` / `full_observed`.
- Thread C (remover): drains attached-roots queue, calls `removeFetchedBlock`. No double-free.

Critical assertion: `try testing.expect(partial_observed > 0)` AND `try testing.expect(full_observed > 0)` at the end — fails LOUDLY if the test ever degenerates into a smoke test that doesn't actually exercise the partial-state surface.

Synthetic `SignedBlock` fixtures (zeroed body, no XMSS, no STF). Test runtime <1s. Ran the full test suite **5 times in a row** post-implementation — all 5 runs `EXIT:0` with the test passing.

---

### N2 — perf concern, NOT a correctness bug — tracking issue filed

`onBlock` keeps `states_lock.exclusive` across `updateBlockDb` + `forkChoice.confirmBlock` (the trade-off taken in `60761c9` to fix a different UAF). This blocks the attestation hot path (`onGossipAttestation`, `verifyAggregatedAttestation`, `chain.aggregate` — all take `states_lock.shared`) for the DB write window (5–50 ms).

Filed as **#821** with two follow-up shapes (sszClone before DB write, or refcount on `*BeamState` in slice c) and a request to instrument `zeam_lock_hold_seconds{lock="states", site="onBlock.commit"}` for a devnet measurement first.

---

### Build/test gates

- `zig fmt pkgs/` clean.
- `zig build -Doptimize=Debug` → `EXIT:0`.
- `zig build --summary all test -Doptimize=Debug` → `EXIT:0`. The two new tests show as:
  - `60/89 locking.test.LockedMap: withValueLocked holds the lock across the callback...OK`
  - `61/89 locking.test.LockedMap: withValueLocked invokes callback with null on missing key...OK`
  - `62/89 locking.test.LockedMap: withValueLocked propagates callback errors...OK`
  - `63/89 locking.test.BlockCache: split insertBlockPtr(null)+attachSsz race vs concurrent reader...OK`
- Full-suite test loop (5 runs) all `EXIT:0`, partial-state assertion fired every run.

Branch HEAD: `42b4566`.

zclawz:
## CI fix pushed (commit 9c9c21a)

**Failure:** `locking.test.BlockCache: split insertBlockPtr(null)+attachSsz race vs concurrent reader` panicking on macOS with `bs.ssz.?[0] == 0xDE` assertion.

**Root cause:** `getBlockAndSsz` returned a `?[]const u8` borrow pointing into cache-owned storage. The Remover thread can call `removeFetchedBlock` after the mutex releases, freeing those bytes while the reader still holds a slice header — UAF. macOS allocator surfaces freed memory as garbage bytes, triggering the assert.

**Fix (4 files):**
- `locking.zig`: Remove `getBlockAndSsz` (borrow-shape), replace with `cloneBlockAndSsz` (deep-clone under lock, caller owns). Race test updated to use clone + reduced ITER 1500→300 (clone is heavier: full SSZ round-trip per probe).
- `network.zig`: Remove `getFetchedBlockWithSsz`, add `cloneFetchedBlockAndSsz` delegating to `cloneBlockAndSsz`.
- `node.zig`: Use `cloneFetchedBlockAndSsz` + `defer cached.deinit(self.allocator)` so block + ssz are freed after `chain.onBlock`.
- `chain.zig`: All tests updated to `cloneBlockAndSsz` + proper deinit.

CI re-run triggered automatically.

zclawz:
### macOS CI failure → real UAF in `BlockCache.getBlockAndSsz` — fixed in 9c9c21a

Thanks @ch4r10t33r for catching this on macOS CI. **The new N3 race test (`BlockCache: split insertBlockPtr(null)+attachSsz race vs concurrent reader`) was designed to surface exactly this class of bug, and its assertion was correctly tripped by the underlying API rather than a test issue.** Linux happens to keep the freed buffer intact long enough for the assertion to pass; macOS frees + reuses with different bytes, so the assertion fires.

#### Root cause

`BlockCache.getBlockAndSsz` returned a `BlockAndSsz { block: SignedBlock, ssz: ?[]const u8 }` whose internal slices (block body / signatures, ssz bytes) pointed into **cache-owned memory**. Once the mutex released, a concurrent `removeFetchedBlock(root)` could `fetchRemove + bv.deinit() + allocator.free(ssz)` — the reader's slice headers then alias freed memory. UAF.

#### Production exposure (more important than the test)

`pkgs/node/src/node.zig:processCachedDescendants` (gossip-replay path) was the live exposure:

```zig
if (self.network.getFetchedBlockWithSsz(descendant_root)) |cached| {
    const cached_block = cached.block;
    // ...
    const block_ssz = cached.ssz;
    const missing_roots = self.chain.onBlock(cached_block, .{ .sszBytes = block_ssz }) catch ...;
    // ...
    self.chain.onBlockFollowup(true, &cached_block);
    _ = self.network.removeFetchedBlock(descendant_root);
}
```

`chain.onBlock` runs STF + XMSS verify (hundreds of ms). During that window, another thread (gossip ingestion / RPC / this very loop on a sibling root) could call `removeFetchedBlock` on a different block whose `removeChildrenOf` cascade frees the descendant — or simply on the same root from a different replay path — freeing both the SignedBlock interior storage AND the ssz bytes mid-STF.

#### Fix (commit 9c9c21a)

Holding the cache mutex across `chain.onBlock` would block every reader for hundreds of ms — regression worse than the bug. Correct shape is **clone-then-release**: dupe block + ssz under the cache mutex, return owned copies, caller takes ownership and frees.

- **Removed** `BlockCache.getBlockAndSsz` (borrow-shape) and `Network.getFetchedBlockWithSsz`. Slice discipline: don't ship footguns even if documented.
- **Added** `BlockCache.cloneBlockAndSsz(root, allocator) -> ?OwnedBlockAndSsz` — deep-clones the SignedBlock via `types.sszClone` (the existing codebase deep-clone idiom — round-trip through ssz bytes) and dupes the ssz bytes via `allocator.dupe`, both **inside** the critical section. Caller owns the result and `deinit(allocator)`s it.
- **Added** `Network.cloneFetchedBlockAndSsz(root, allocator)` thin wrapper.
- **Updated** `processCachedDescendants` to use the cloning variant with `defer cached.deinit(self.allocator)` covering every exit path (including the `chain.onBlock` error branches via `catch ... continue`).
- **Migrated** all chain.zig + locking.zig tests to `cloneBlockAndSsz` (with proper `deinit`).
- **Reduced** N3 test ITER 1500 → 300 (and the chain.zig 3-thread stress N 1500 → 200): each `cloneBlockAndSsz` probe is now a full SSZ round-trip under the mutex, much heavier than the borrow-getter; lower N keeps wall time reasonable while still exercising the 3-thread interleaving.
- **Added** a slice-discipline header comment on `BlockCache` warning that direct access to `self.blocks`/`self.ssz_bytes` from outside the type bypasses the clone discipline.

#### Sweep audit

Searched for symmetric LockedMap-with-slice-fields patterns (same UAF class as N1 / `snapshotPendingRequest` from 3aa3866). Findings:

```
$ grep -rn "getBlockAndSsz\|getFetchedBlockWithSsz\|BlockAndSsz\b" pkgs/ --include="*.zig"
# only docstring/historical references remain; all callsites migrated.
```

`Network.getFetchedBlock` (without ssz) still exists at network.zig:481, 507 in `pruneCachedBlocks`. Audited: both callers only read `block.block.parent_root` (fixed-size `[32]u8` array) and `block.block.slot` (u64) — POD subfields that copy by value into the returned struct. No slice-field access. **Safe by construction**, but the function remains a footgun for future callers; that's a separate cleanup.

`PendingBlockRootMap = LockedMap(Root, u32)` — value is plain `u32`, no slice fields. Safe.

`PendingRPCMap` already migrated to `withValueLocked` in 3aa3866 (snapshotPendingRequest UAF fix).

#### Verification

- `zig build -Doptimize=Debug` → EXIT:0
- `zig build --summary all test -Doptimize=Debug` → EXIT:0
- macOS CI on this commit: `test (macos-latest) → success` ✅ (https://github.com/blockblaz/zeam/actions/runs/25342321960/job/74302503762)
- 5× isolated full-suite runs locally; N3 test passes deterministically (the clone owns the bytes, the assertion can't race a `removeFetchedBlock` anymore).

#### Post-merge regression vector

Any future caller that touches `BlockCache.blocks` / `ssz_bytes` directly bypasses the clone discipline. The new header comment on `BlockCache` flags this. Worth keeping in mind for slices (b)/(c) wiring: anyone adding a new accessor that returns slice-bearing data from this cache MUST clone under the lock.

ch4r10t33r (approved):
LGTM

PR #824, opened 2026-05-05
feat: implement blocksByRange RPC for efficient bulk sync

zclawz:
Closes #823

## Summary

Implements the `blocksByRange` RPC protocol as added in leanSpec PR https://github.com/leanEthereum/leanSpec/pull/691.

### Server side
- New `BlocksByRangeRequest` type with `start_slot` and `count` fields.
- Server walks the canonical chain backward from the current head via `forkChoice.getBlock`, collects every block whose slot falls in `[start_slot, start_slot + count)`, then streams them in slot-ascending order.
- Count is capped at `MAX_REQUEST_BLOCKS` (1024) to bound work per request.

### Client side
- `requestBlocksByRange` / `sendBlocksByRangeRequest` on `Network`.
- New `BlockByRangeContext` + `PendingRPC` variant for tracking.
- `processBlockByRangeChunk` reuses `chain.onBlock` and the existing parent-caching path, so out-of-order or missing-parent chunks fall back to the by-root recovery flow.

### Sync trigger
- When a status response arrives showing the peer is far ahead (`gap > BLOCKS_BY_RANGE_SYNC_THRESHOLD`, currently 64 slots), the node now issues a single bulk `blocks_by_range` request from `our_head_slot + 1` instead of the recursive head-by-root walk. Falls back to head-by-root if the bulk request fails.

### Wiring
- Added `blocks_by_range` to `LeanSupportedProtocol` (both the Zig enum in `pkgs/network/src/interface.zig` and the Rust enum in `rust/libp2p-glue/src/req_resp/protocol_id.rs`).
- Pinned the enum ordinals so the `protocol_tag: u32` that crosses the FFI boundary stays consistent: `blocks_by_root=0, status=1, blocks_by_range=2`. Matches `TryFrom<u32>` on the Rust side.
- Registered the new protocol in the `ReqResp` behaviour vector in `rust/libp2p-glue/src/lib.rs`.
- Updated `mock.zig` and every exhaustive switch on `LeanSupportedProtocol` / `request_kind` (handler, response handler, failure path, timeout sweep, snapshot, finalize, mock fixtures).

## Testing

`zig build test` passes locally (zig 0.16.0): all 11 test suites green (13 + 0 + 4 + 0 + 7 + 8 + 18 + 3 + 18 + 89 + 11 tests).

No new dedicated unit test for the range RPC — leaning on the existing protocol-wiring tests + the fact that the change reuses `processBlock` for chunk handling. Happy to add a focused mock-based round-trip test if reviewers want one.

cc @g11tech

### DIFF
--- a/pkgs/network/src/interface.zig
+++ b/pkgs/network/src/interface.zig
@@ -35,6 +35,7 @@ pub const DisconnectionReason = enum(u32) {
 
 const topic_prefix = "leanconsensus";
 const lean_blocks_by_root_protocol = "/leanconsensus/req/blocks_by_root/1/ssz_snappy";
+const lean_blocks_by_range_protocol = "/leanconsensus/req/blocks_by_range/1/ssz_snappy";
 const lean_status_protocol = "/leanconsensus/req/status/1/ssz_snappy";
 
 fn unionPayloadType(comptime UnionType: type, comptime tag: anytype) type {
@@ -363,13 +364,16 @@ pub const GossipMessage = union(GossipTopicKind) {
     }
 };
 
-pub const LeanSupportedProtocol = enum {
-    blocks_by_root,
-    status,
+pub const LeanSupportedProtocol = enum(u32) {
+    // Ordinals must match `rust/libp2p-glue/src/req_resp/protocol_id.rs::TryFrom<u32>`.
+    blocks_by_root = 0,
+    status = 1,
+    blocks_by_range = 2,
 
     pub fn protocolId(self: LeanSupportedProtocol) []const u8 {
         return switch (self) {
             .blocks_by_root => lean_blocks_by_root_protocol,
+            .blocks_by_range => lean_blocks_by_range_protocol,
             .status => lean_status_protocol,
         };
     }
@@ -395,6 +399,10 @@ pub const LeanSupportedProtocol = enum {
             return .blocks_by_root;
         }
 
+        if (std.mem.eql(u8, protocol_id, lean_blocks_by_range_protocol)) {
+            return .blocks_by_range;
+        }
+
         return error.UnsupportedProtocol;
     }
 };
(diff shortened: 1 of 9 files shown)


zclawz:
Pushed `a7776e9` — addresses the two review comments above.

**Comment 1: Rust enum `repr` asymmetry — fixed.**
- `rust/libp2p-glue/src/req_resp/protocol_id.rs`: added `#[repr(u32)]`, reordered + explicit discriminants `BlocksByRootV1 = 0, StatusV1 = 1, BlocksByRangeV1 = 2`. Kept the manual `TryFrom<u32>` as a sanity wrapper (no new `num_enum` dep).
- 3 unit tests in `mod tests`: `try_from_round_trip_matches_repr` (the foot-gun pin), `discriminants_match_zig_side`, `try_from_rejects_out_of_range`. All pass.
- Updated Zig comment at `pkgs/network/src/interface.zig::LeanSupportedProtocol` to call out the bidirectional FFI invariant + the Rust-side guard test.

**Comment 2: mock round-trip tests — partially landed.**
Added 3 mock-network round-trip tests in `pkgs/network/src/mock.zig` covering the wire-contract scenarios that are testable without a full BeamNode harness:
- `multi-chunk in slot-ascending order` — 3 chunks at slots 7/8/9, asserts order + single `completed` + zero failures.
- `empty stream + clean finish (start_slot past head)` — server `finish()` with no responses; asserts zero chunks + `completed`.
- `RESOURCE_UNAVAILABLE error path` — server `sendError(3, ...)`; asserts failure event with code 3, no chunks, no bare `completed`.

Scenario 1 from the review (`count == 0`) is already exercised by the client-side guard at `pkgs/node/src/network.zig:668` (`error.NoBlocksRequested`); didn't add a redundant test for it.

**Deferred to follow-up issues** (filing shortly, will link here):
- Scenarios 5 (gaps preserved via empty-slot skip in finalized walk) and 6 (sync-trigger gap > 64 → range vs head-by-root) — both require the full BeamNode harness, not just mock-network.
- The three additional issues you flagged in your second comment: (a) request-ID-vs-pending-state race in `sendBlocksByRangeRequest`/`ensureBlocksByRootRequest`, (b) per-peer rate limit / DoS, (c) `@min(gap, MAX_REQUEST_BLOCKS)` cycle-gated catch-up cadence.

Race fix specifically I want to land in its own PR since it's cross-cutting (the same pattern lives in `ensureBlocksByRootRequest` and the fix should land both paths in one diff with a designed locking story).

Test status:
- `cargo test --package libp2p-glue --lib req_resp::protocol_id::tests` — 3 pass.
- `zig build test` — all 11 network/mock tests pass, all 89 forkchoice+locking, no regressions across the rest of the suite.

zclawz:
Follow-up tracking issues filed:

- #833 — request-ID vs pending-state race in `sendBlocksByRangeRequest` / `ensureBlocksByRootRequest` (cross-cuts `blocks_by_root` too; will land both in one PR)
- #834 — per-peer rate-limit / DoS handling for the `blocks_by_range` server
- #835 — `@min(gap, MAX_REQUEST_BLOCKS)` cycle-gated catch-up cadence

Ready for re-review.

ch4r10t33r (approved):

PR #827, opened 2026-05-05
node: introduce RcBeamState refcounted state pointer (slice c-2a of #803)

zclawz:
## Slice (c-2a) — RcBeamState refcounted state pointer

Implements item 3 of `docs/threading_refactor_slice_a.md` §"Slice (c) scope (revised)" — refcounted state pointers so cross-thread readers don't block worker mutations. Per the c-2 plan posted at https://github.com/blockblaz/zeam/issues/803#issuecomment-4381027301, c-2 is split into three reviewable chunks:

- **c-2a (this PR)** — `RcBeamState` type + tests.
- **c-2a follow-up** — migrate `BeamChain.states` map storage to `*RcBeamState` (no behaviour change, all current synchronous semantics preserved).
- **c-2b** — wire `chain_worker` field, per-handler migration, drop write locks.
- **c-2c** — queue-depth saturation stress + devnet4 burn-in.

This first PR is pure-additive: it lands the type and its tests in isolation. No existing code is modified except `pkgs/node/src/lib.zig` to expose the new module. The follow-up PR will land the chain.zig migration on top.

### Why split this from the chain.zig migration?

`BeamChain.states` is touched at ~9 production callsites + ~30 test callsites (per the slice-(a-2) inventory). Bundling the type introduction with the storage migration would obscure the refcount semantics review — reviewers would have to track the type's correctness AND the migration safety in one diff. After the c-1 parallel-agent collisions, smaller per-PR diffs are clearly worth it.

### What's in this PR

New file `pkgs/node/src/rc_beam_state.zig` (291 LOC including tests):

```zig
pub const RcBeamState = struct {
    allocator: Allocator,
    refcount: std.atomic.Value(u32),
    state: types.BeamState,

    pub fn create(allocator: Allocator, state: types.BeamState) !*Self { ... }
    pub fn acquire(self: *Self) *Self { ... }    // refcount.fetchAdd(1, .acq_rel)
    pub fn release(self: *Self) void { ... }     // refcount.fetchSub(1, .acq_rel); free at 0
    pub fn count(self: *const Self) u32 { ... }  // .monotonic; observational only
};
```

### Design choices (per the c-2 plan)

1. **Storage layout (a)** — single allocation with `BeamState` embedded inline alongside the refcount. Pros: one allocation per state, one `release` frees the whole header + state, no double-pointer chase on the read path. Cons: cannot share a refcount across multiple distinct `BeamState` pointers (e.g. shadowing). No caller currently needs this; option (b) split layout can be added later if a use case appears.

2. **Atomic refcount with `.acq_rel` ordering** on bumps and decrements so the freeing thread observes all prior writes to the wrapped state. The `.monotonic` load in `count()` is observational-only (tests + metrics).

3. **Ownership transfer at `create()`** — takes a freshly-allocated `BeamState` by value, transfers it into the heap allocation. Initial refcount = 1 (the creator's reference). Returns `error.OutOfMemory` on allocation failure with the input state NOT consumed.

4. **Double-release is a debug-build assert** (refcount underflow). Release-build behaviour is silent UB — the map's invariant must guarantee callers hold a valid acquire before calling release. Slice (a-2)'s `BorrowedState` already enforces this for the lock path; under c-2b it will enforce it for the refcount path too.

### Tests (6 new, all pass)

- `create + release frees the state` — most basic round-trip; `testing.allocator`'s leak detector verifies clean teardown.
- `acquire + release pair is balanced` — count goes 1 → 2 → 1.
- `multiple acquires and releases keep state alive until last` — 4 references; release in arbitrary order.
- `acquire() returns the same pointer` — chained call sites.
- `concurrent acquire/release stress (40 threads × 10k iters)` — high-contention test; each thread takes acquire, touches `state.slot` via `doNotOptimizeAway`, releases. Final refcount must be exactly 1 (creator's reference) — proves no acquire/release pair lost an increment.
- `release order doesn't matter` — two threads each acquire+release independently, final refcount = 1 regardless of interleaving.

### Verification

```
...
83/114 rc_beam_state.test.RcBeamState: create + release frees the state...OK
84/114 rc_beam_state.test.RcBeamState: acquire + release pair is balanced...OK
85/114 rc_beam_state.test.RcBeamState: multiple acquires and releases keep state alive until last...OK
86/114 rc_beam_state.test.RcBeamState: acquire returns the same pointer (cheap to chain)...OK
87/114 rc_beam_state.test.RcBeamState: concurrent acquire/release stress (40 threads × 10k iters)...OK
88/114 rc_beam_state.test.RcBeamState: release order doesn't matter (acquire-then-release vs release-then-acquire)...OK
...
All 114 tests passed.
EXIT:0
```

All 6 new tests pass, all 108 prior tests pass (was 108 with c-1's chain_worker; now 114 with the 6 RcBeamState tests), EXIT:0.

### Branch lineage

- Off `main` post-#826 merge (`f2b8735`).
- Tip: `509df11`.
- c-2a follow-up + c-2b will branch from this once it merges.

cc @ch4r10t33r @GrapeBaBa — design questions on the c-2 plan posted at https://github.com/blockblaz/zeam/issues/803#issuecomment-4381027301. Default proposals are reflected in this PR; happy to revisit (a)-vs-(b) layout if there's a future-state shadowing case I missed.

### DIFF
--- a/pkgs/node/src/lib.zig
+++ b/pkgs/node/src/lib.zig
@@ -22,6 +22,9 @@ pub const chain_worker = @import("./chain_worker.zig");
 pub const ChainWorker = chain_worker.ChainWorker;
 pub const ChainWorkerMessage = chain_worker.Message;
 
+pub const rc_beam_state = @import("./rc_beam_state.zig");
+pub const RcBeamState = rc_beam_state.RcBeamState;
+
 const networks = @import("@zeam/network");
 pub const NodeNameRegistry = networks.NodeNameRegistry;
 
@@ -31,5 +34,6 @@ test "get tests" {
     _ = @import("./utils.zig");
     _ = @import("./locking.zig");
     _ = @import("./chain_worker.zig");
+    _ = @import("./rc_beam_state.zig");
     @import("std").testing.refAllDecls(@This());
 }
--- a/pkgs/node/src/rc_beam_state.zig
+++ b/pkgs/node/src/rc_beam_state.zig
@@ -56,8 +56,11 @@ const types = @import("@zeam/types");
 ///
 /// Lifecycle:
 ///
-///     // Creator path: takes ownership of `state` (which must be
-///     // a freshly-allocated, fully-initialised BeamState).
+///     // Creator path: `create` takes ownership of `state`
+///     // unconditionally. On success, the heap allocation embeds
+///     // `state` and the matching `release` frees it; on OOM,
+///     // `state.deinit()` runs before the error returns. Either
+///     // way the caller does NOT call `state.deinit()` after.
 ///     const rc = try RcBeamState.create(allocator, state);
 ///     defer rc.release();              // initial reference
 ///
@@ -81,16 +84,33 @@ pub const RcBeamState = struct {
 
     const Self = @This();
 
-    /// Create a new refcounted state wrapping `state` (transferred by
-    /// value into the heap allocation; caller must not retain a
+    /// Create a new refcounted state wrapping `state` (transferred
+    /// by value into the heap allocation; caller must not retain a
     /// pointer to the old `state` location). Initial refcount is 1
     /// — the creator is responsible for the matching `release`.
     ///
-    /// Returns `error.OutOfMemory` if the allocation fails. On
-    /// failure, `state` is NOT consumed — the caller still owns it
-    /// and is responsible for its cleanup.
+    /// **Always-consume contract.** This function takes ownership of
+    /// `state` unconditionally. On the success path, `state` is
+    /// embedded into the heap allocation and the matching `release`
+    /// frees it. On `error.OutOfMemory`, `state.deinit()` is called
+    /// before the error is returned, so the caller never has to
+    /// remember an `errdefer state.deinit()` between
+    /// `buildState(...)` and `RcBeamState.create(...)`.
+    ///
+    /// Earlier versions of this function returned without consuming
+    /// `state` on the OOM path. That asymmetry forced every caller
+    /// to write a fragile cleanup pattern; symmetric ownership is
+    /// the lower-footgun shape for an API used at every chain.zig
+    /// state-write site.
     pub fn create(allocator: Allocator, state: types.BeamState) !*Self {
-        const self = try allocator.create(Self);
+        const self = allocator.create(Self) catch |err| {
+            // OOM on the wrapper allocation — we still consumed
+            // `state` per the always-consume contract, so drop its
+            // owned interior allocations before returning.
+            var consumed = state;
+            consumed.deinit();
+            return err;
+        };
         self.* = .{
             .allocator = allocator,
             .refcount = std.atomic.Value(u32).init(1),
@@ -219,12 +239,22 @@ test "RcBeamState: acquire returns the same pointer (cheap to chain)" {
     reader.release();
 }
 
-test "RcBeamState: concurrent acquire/release stress (40 threads × 10k iters)" {
-    // High-contention test: 40 threads each take an acquire, do a
+test "RcBeamState: acquire/release pair accounting under producer race (40 threads × 10k iters)" {
+    // Pair-accounting test: 40 threads each take an acquire, do a
     // tiny no-op load, then release. The creator's refcount stays
     // at 1 throughout (each thread's acquire/release pair is
     // balanced). At the end, refcount must be exactly 1, then the
     // creator's release frees.
+    //
+    // What this test does NOT prove: the freeing race. Because the
+    // creator's reference is held throughout, every fetchSub call
+    // observes `prev > 1` and the freeing branch is never taken
+    // under contention — the acq_rel ordering on the freeing
+    // thread is not exercised. The next test
+    // ("freeing race — last release wins under contention") fills
+    // that gap by dropping the creator's reference partway and
+    // verifying that whichever thread brings refcount to 0 frees
+    // cleanly under testing.allocator's leak detection.
     const state = try makeState();
     const rc = try RcBeamState.create(testing.allocator, state);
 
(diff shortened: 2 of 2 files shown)


zclawz:
Both review issues addressed in `88f56b6`. Each fix is in code (compile-time/runtime enforced) plus a matching test that fires the path the review called out.

### 1. `create()` ownership asymmetry → always-consume

Switched to option (a) per the review. `create()` now consumes `state` unconditionally:

```zig
pub fn create(allocator: Allocator, state: types.BeamState) !*Self {
    const self = allocator.create(Self) catch |err| {
        // OOM on the wrapper allocation — we still consumed
        // `state` per the always-consume contract, so drop its
        // owned interior allocations before returning.
        var consumed = state;
        consumed.deinit();
        return err;
    };
    self.* = .{ ... .state = state };
    return self;
}
```

Lifecycle doc + function docstring updated; the previous "caller still owns state on OOM" footgun is gone. Locked the contract in code with a new test:

```zig
test "RcBeamState.create: OOM path consumes state (no leak under FailingAllocator)" {
    const state = try makeState();
    var failing = std.testing.FailingAllocator.init(testing.allocator, .{ .fail_index = 0 });
    const result = RcBeamState.create(failing.allocator(), state);
    try testing.expectError(error.OutOfMemory, result);
    // Caller does NOT call state.deinit() here. If create() failed
    // to consume, testing.allocator's leak detector trips.
}
```

The test passes ONLY if `create()` consumed `state` on the OOM path — exactly the assertion the review wanted.

### 2. Freeing-race test → renamed + 2 real tests added

(a) Renamed the existing test to `"acquire/release pair accounting under producer race (40 threads × 10k iters)"` — its actual proof is pair-balance under contention, not the freeing race. Added a header comment that points at the new tests for the freeing path.

(b) Added `"freeing race — last release wins under contention"`:

```zig
// Pre-bump refcount once per worker BEFORE spawning (the spawn
// point still holds the creator's reference, so the bump is
// safe), hand each worker its pre-acquired ref, then drop the
// creator's ref WHILE workers are still running. Whichever
// thread brings refcount to 0 — worker or creator depending on
// interleaving — must free cleanly. OUTER_ITERS=50 to sweep
// interleavings. testing.allocator's leak/UAF detector is the
// assertion.
```

The pre-bump is critical: my first version had workers call `acquire()` as their first action, but spawn doesn't hand a valid acquire to the worker — by the time the worker ran, the rc could already have been freed by another worker that finished first. UAF on the acquire itself. CI caught it via segfault on test 88. Pre-bumping on the spawning thread (which holds the creator's reference at that moment) closes the hole. Documented in the test header so future contributors don't repeat the bug.

(c) Added `"freeing race — worker takes the final release"`:

```zig
// Sequence:
//   1. create(state)            → refcount = 1 (creator)
//   2. handed = rc.acquire()    → refcount = 2 (creator + handed)
//   3. rc.release()             → refcount = 1 (handed only)
//   4. spawn(worker, handed)
//   5. join
// The worker is now the sole reference holder; its release
// brings refcount to 0 and frees. Exercises .acq_rel on a
// non-creator thread.
```

Deterministically forces the WORKER (not the creator) to hit `prev == 1` on `fetchSub` and free.

### Verification

```
...
83/117 rc_beam_state.test.RcBeamState: create + release frees the state...OK
84/117 ... acquire + release pair is balanced...OK
85/117 ... multiple acquires and releases keep state alive until last...OK
86/117 ... acquire returns the same pointer...OK
87/117 ... acquire/release pair accounting under producer race (40 threads × 10k iters)...OK
88/117 ... freeing race — last release wins under contention...OK
89/117 ... freeing race — worker takes the final release...OK
90/117 ... release order doesn't matter...OK
91/117 ... create: OOM path consumes state (no leak under FailingAllocator)...OK
...
All 117 tests passed.
EXIT:0
```

9 rc_beam_state tests pass (was 6: 3 added, 1 renamed). All 108 prior tests pass. EXIT:0.

Branch tip: `88f56b6`. Diff vs `509df11`: +209 / -10 in `pkgs/node/src/rc_beam_state.zig`.

zclawz:
r2 review fixes pushed in `8ba88f3`. Two issues addressed in code; three c-2b follow-ups filed on #803.

### 1. `acquire()` returned mutable `*Self` for every reader → split into two entry points

```zig
pub fn acquire(self: *Self) *Self            // chain worker (writer path)
pub fn acquireConst(self: *Self) *const Self // every cross-thread reader
```

`acquireConst` is a thin wrapper around `acquire` (refcount semantics identical) but the return type makes the c-2b sole-writer invariant compile-checked: only the chain worker can get a `*Self` (mutable); every cross-thread reader (HTTP API, metrics scrape, event broadcaster) gets `*const Self` and physically can't mutate `state.slot` to corrupt the worker's view.

`release` was retyped to take `*const Self` so readers holding `*const Self` can release directly without casting at the call site. Internally it uses `@constCast` for the destructive ops — release IS a destructor and is allowed to mutate even on a const view (matches how Rust `Arc`'s `Drop` impl works on a shared reference).

Two new tests:
- `acquireConst: returns *const Self that releases cleanly` — single-threaded round-trip
- `acquireConst: concurrent readers free cleanly when creator drops` — 8 readers × 25 outer iters using `acquireConst` exclusively, exercising the const-pointer + `release(*const Self)` path under contention

### 2. `acquire()` used `.acq_rel` → switched to `.monotonic`

Standard refcount pattern (Rust `Arc::clone`, C++ `shared_ptr(const&)`) uses `.monotonic` on the increment because the caller already holds a valid acquire — there is nothing for the increment to synchronise against. The freeing thread synchronises via the `.acquire` half of its own `.acq_rel` `fetchSub` on `release`, not against any acquire-side fence.

```zig
pub fn acquire(self: *Self) *Self {
    const prev = self.refcount.fetchAdd(1, .monotonic);  // was .acq_rel
    std.debug.assert(prev < std.math.maxInt(u32));
    return self;
}
```

`release` stays at `.acq_rel` — that's where the actual synchronisation happens. File-header comment + refcount-field doc rewritten to spell out the rules:

> * `acquire`: `.monotonic` fetchAdd. The caller already holds a valid acquire (that's the contract); incrementing the count cannot race with a free because no thread can free until refcount reaches 0, which can't happen while we hold a reference. No fence needed on this side.
> * `release`: `.acq_rel` fetchSub. The freeing thread (the one that observes `prev == 1`) needs `.acquire` to synchronise with all prior `.release` decrements from other threads, so it observes every prior write to the state before calling `state.deinit()`.

Hot-path win for the c-2b reader path where every API request does an `acquireConst`.

### 3-5. c-2b follow-ups → filed on issue #803

The other three review points are c-2b shape decisions, not c-2a fixes. Posted as a tracking comment on the parent issue: https://github.com/blockblaz/zeam/issues/803#issuecomment-4381905197

- **`tryAcquire()` for upgrade-from-weak**: needed when c-2b drops the rwlock from `BorrowedState`. Standard `cmpxchg(N → N+1, reject N=0)` pattern; freeing thread does `cmpxchg(1 → 0)`. Includes a sketch of the implementation.
- **Double-release assert is best-effort**: already documented honestly in the doc + commit; flagged for c-2b reviewers so they don't rely on it as a debug crutch.
- **Refcount-distribution histogram metric**: `lean_chain_state_refcount_distribution{le="1|2|4|8|16|32|+Inf"}` sampled by metrics scrape, populated by iterating `BeamChain.states`. c-2b job alongside the per-handler migration.

### Verification

```
...
83/119 rc_beam_state.test.RcBeamState: create + release frees the state...OK
84/119 ... acquire + release pair is balanced...OK
85/119 ... multiple acquires and releases keep state alive until last...OK
86/119 ... acquire returns the same pointer...OK
87/119 ... acquire/release pair accounting under producer race (40 threads × 10k iters)...OK
88/119 ... freeing race — last release wins under contention...OK
89/119 ... freeing race — worker takes the final release...OK
90/119 ... release order doesn't matter...OK
91/119 ... create: OOM path consumes state (no leak under FailingAllocator)...OK
92/119 ... acquireConst: returns *const Self that releases cleanly...OK
93/119 ... acquireConst: concurrent readers free cleanly when creator drops...OK
...
All 119 tests passed.
EXIT:0
```

11 rc_beam_state tests pass (was 9 in r1: 2 added). All 108 prior tests pass. EXIT:0.

Branch tip: `8ba88f3`. Diff vs `88f56b6`: +160 / -19 in `pkgs/node/src/rc_beam_state.zig`.

zclawz:
r3 review fixes pushed in `a9fd5e8`. Both points correct — fixed in code + doc.

### 1. Overclaim corrected: `acquire` → `acquireWriter`, doc rewritten honestly

You're right that the previous wording overstated the compile-time guarantee. The type system enforces the **consequence** (a `*const Self` holder cannot mutate `state.<field>`), not the **cause** (who holds the mutable ref). Anyone could call `acquire()` and get `*Self` — there's no compile-time gate.

Picked option (a) per the review:

```zig
pub fn acquireWriter(self: *Self) *Self        // chain worker (convention)
pub fn acquireConst(self: *Self) *const Self   // every cross-thread reader
```

The asymmetric names are loud at every call site; code review enforces the convention. Updated the doc to spell out exactly that:

> Convention, NOT compile-time enforcement: the type system does not gate `acquireWriter` to a particular thread — anyone can call it and get `*Self`. What the type system DOES enforce is the consequence: a holder of `*const Self` (from `acquireConst`) cannot mutate `state`. The asymmetric names (`acquireWriter` vs `acquireConst`) make the convention loud at every call site, and code review (not the compiler) keeps non-worker code paths off this entry point. If convention proves leaky in c-2b, the next step is two distinct types (`RcBeamStateWriter` / `RcBeamStateReader`).

`acquireConst` delegates to `acquireWriter` (refcount semantics identical). All test/doc call sites renamed. No external callers exist yet — c-2a is pure-additive; the `BeamChain.states` migration is c-2b's job, so c-2b can choose `acquireWriter`/`acquireConst` per call site cleanly.

### 2. `@constCast` UB-on-true-const caveat documented

You're right — silent foot-gun. Today every `RcBeamState` goes through `create()` (heap-allocated, so `@constCast` is safe), but a future contributor writing a test fixture or example might do:

```zig
const rc_global: RcBeamState = ...; // BUG
rc_global.release();                  // UB in optimised build
```

…and the bug manifests only as silent miscompilation. Added an explicit MUST NOT clause to `release`'s docstring:

> MUST NOT be called on a `*const Self` derived from a true const value (a global `const RcBeamState`, or a stack-local declared `const`). The `@constCast` is only safe because every `RcBeamState` is heap-allocated by `create()` — the underlying memory is mutable, the `*const` qualifier is purely a view restriction. Mutating through a pointer derived from a true-const value is undefined behaviour in optimised builds (the compiler is allowed to assume the memory does not change). Future contributors writing test fixtures or examples MUST go through `create()`; do not fabricate an `RcBeamState` value as a `const` decl and release it.

### Verification

```
...
83/119 rc_beam_state.test.RcBeamState: create + release frees the state...OK
84/119 ... acquire + release pair is balanced...OK
85/119 ... multiple acquires and releases keep state alive until last...OK
86/119 ... acquire returns the same pointer (cheap to chain)...OK
87/119 ... acquire/release pair accounting under producer race (40 threads × 10k iters)...OK
88/119 ... freeing race — last release wins under contention...OK
89/119 ... freeing race — worker takes the final release...OK
90/119 ... release order doesn't matter...OK
91/119 ... create: OOM path consumes state (no leak under FailingAllocator)...OK
92/119 ... acquireConst: returns *const Self that releases cleanly...OK
93/119 ... acquireConst: concurrent readers free cleanly when creator drops...OK
...
All 119 tests passed.
EXIT:0
```

11 rc_beam_state tests still pass; test count unchanged (rename is mechanical). All 108 prior tests pass.

Branch tip: `a9fd5e8`. Diff vs `8ba88f3`: +56 / -26 in `pkgs/node/src/rc_beam_state.zig`.

zclawz:
r4 review fixes pushed in `fed2f34` — doc-only changes addressing both points.

### 1. Negative-case comment in the const test

You're right, the test framing oversold what the read demonstrated. Kept the positive read, added an explicit negative-case comment showing the line a contributor would write if they forgot the const contract:

```zig
const reader: *const RcBeamState = rc.acquireConst();
// Read-only deref through the const view. The negative case
// — the line that would fail to compile if a contributor
// tried to mutate — is shown explicitly below as a comment
// (uncommenting it surfaces the type-system guarantee at
// build time):
//
//     reader.state.slot = 5;
//     // error: cannot assign to constant
//
// We don't actually compile-fail in this test (Zig has no
// stable in-source @compileError fixture for negative tests),
// but the commented line documents what the type system
// catches. The positive case is below: a plain read.
std.mem.doNotOptimizeAway(reader.state.slot);
```

Cheapest path that makes the guarantee visible at human-review time. (Zig 0.16 still doesn't have a stable in-source `@compileError` fixture for negative tests — the closest pattern is `std.testing.refAllDecls` + a separate file with `comptime` blocks, which is heavier than the bug warrants.)

### 2. Lifetime contract on `acquireConst`

You're right that the previous doc didn't spell out the contract. Added a `## Lifetime contract (READ THIS BEFORE c-2b)` section covering the c-2b reader rules. Highlights:

> * Drop the borrow as soon as the read is done. Acquire, read what you need (snapshot fields, copy slices), call release. Treat it like a Rust scoped borrow, not a cached handle.
>
> * Do NOT cache the `*const Self` in a long-lived struct field. That is the obvious footgun; it pins the state across STF advances and blocks the freeing path on stale heads. Each consumer takes a fresh acquire per unit of work.
>
> * Do NOT hold the borrow across an STF FFI window or any long-running call. The chain worker (sole writer) may be ready to retire that state; pinning it serialises the worker against the slowest reader.
>
> * The borrow does NOT pin `state.<field>` against in-place mutation. The chain-worker's `acquireWriter` view of the same `RcBeamState` can mutate `state.<field>` concurrently. Today this is fine because the c-2b design promotes-then-releases (writer never mutates a state that has live readers), but the type system does NOT enforce that — see the `acquireWriter` docstring on the convention vs enforcement point.

The "I'll cache `acquireConst()` in a struct field" footgun is now closed before the c-2b per-handler migration starts.

### Verification

```
...
83/119 rc_beam_state.test.RcBeamState: create + release frees the state...OK
...
93/119 rc_beam_state.test.RcBeamState.acquireConst: concurrent readers free cleanly when creator drops...OK
...
All 119 tests passed.
EXIT:0
```

11 rc_beam_state tests still pass; test count unchanged (doc-only). All 108 prior tests pass.

Branch tip: `fed2f34`. Diff vs `a9fd5e8`: +50 / -2 in `pkgs/node/src/rc_beam_state.zig`.

ch4r10t33r (approved):
Looks good.

PR #828, opened 2026-05-05
node: chain-worker routing + state-rwlock drop + queue-saturation stress (slice c-2b + c-2c of #803)

zclawz:
**Slice c-2b + c-2c of #803** — clubbed per @ch4r10t33r's call.

Branch off `main` @ `f76e762` (= c-2a's merge commit `node: introduce RcBeamState refcounted state pointer (slice c-2a of #803) (#827)`).

## Plan

Detailed plan: https://github.com/blockblaz/zeam/issues/803#issuecomment-4382830836

6 commits, each independently reviewable. **All six landed.**

| # | Status | SHA | Commit |
|---|---|---|---|
| 1 | ✅ landed | `5bf12dd` | `tryAcquire` upgrade-from-weak primitive on `RcBeamState` |
| 2 | ✅ landed | `9b3b036` | `BeamChain.states` map value type: `*BeamState` → `*RcBeamState` |
| 3 | ✅ landed | `4709e7b` | Producer-side handlers route through chain_worker queues; `--chain-worker=on\|off` CLI flag |
| 4 | ✅ landed | `cf12167` | Drop `BorrowedState` rwlock when chain-worker is sole writer |
| 5 | ✅ landed | `72d0244` | `lean_chain_state_refcount_distribution` histogram metric |
| 6 | ✅ landed | `df81046` | Queue-saturation stress mode (c-2c part 1) |

Plus three review-fix commits picked up during review rounds:

| SHA | What |
|---|---|
| `1835008` | `zig fmt` (trailing spaces in `rc_beam_state.zig` comment block) |
| `fa7cce1` | Fix `post_state_rc` leak on `statesCommitKeepExisting` OOM (errdefer + `rc_owned` flag, mirrors `statesPutExclusive`) + `FailingAllocator` regression test |
| `c19e5f2` | `wrapOwnedStateIntoRc` helper unifying `produceBlock`/`onBlock` post_state→rc dance + rename misleadingly-titled `tryAcquire: race against release-to-zero` test (reflects what the test actually proves; null-path-under-contention is unreachable by safe code per the doc'd safety preconditions) |

**Burn-in (c-2c part 2)** is NOT gated on this PR — devnet4 burn-in runs from `main` post-merge, then a separate one-line PR flips the `--chain-worker` default from `off` to `on`.

### Commit 1 (`5bf12dd`) — `tryAcquire` primitive

Pure-additive. Standard CAS pattern (Rust `Arc::upgrade` / C++ `weak_ptr::lock`):

```zig
pub fn tryAcquire(self: *Self) ?*Self {
    var current = self.refcount.load(.monotonic);
    while (true) {
        if (current == 0) return null;
        if (self.refcount.cmpxchgWeak(current, current + 1, .monotonic, .monotonic)) |actual| {
            current = actual;
        } else {
            return self;
        }
    }
}
```

The freeing thread's claim is the existing `.acq_rel` `fetchSub` in `release()`; once it commits 1 → 0, every subsequent `tryAcquire` observes 0 and returns null. `.monotonic` on both load + cmpxchg is correct because the increment side has nothing to synchronize against.

4 new tests: basic, producer race (32 threads), null-after-zero (white-box fixture), and the pre-bumped concurrency test (renamed in `c19e5f2` to accurately reflect what it proves vs the unreachable null-under-contention path).

### Commit 2 (`9b3b036`) — `BeamChain.states` migration

Mechanical map type swap. Functionally a no-op:
- Refcount stays at 1 throughout (no other holder takes acquires yet)
- `states_lock` rwlock continues to gate concurrent map access exactly as before
- `BorrowedState` public shape unchanged: still exposes `state: *const BeamState`. Internally points at `&rc.state`.

Updated:
- `init`: stack-build via sszClone + `RcBeamState.create` (replaces create-then-clone-into-heap)
- `deinit`: `.release()` per entry instead of `.deinit() + .destroy()`
- `statesGet`: `BorrowedState{ .state = &rc.state }`
- `statesPutExclusive` / `statesCommitKeepExisting` / `statesFetchRemoveExclusivePtr`: signatures take/return `*RcBeamState`
- `produceBlock.commit` / `onBlock.commit`: wrap heap-allocated `post_state` into rc before commit (with care to flip `post_state_settled` BEFORE `create()` so the upstream errdefer doesn't deref freed memory on OOM)
- `pruneStates`: `release()` per pruned root
- 3 test sites updated

### Commit 3 (`4709e7b`) — chain-worker routing + `--chain-worker` CLI flag

`processBlockJob` / gossip-attestation handlers can now route through the c-1 chain_worker queue under a new `--chain-worker=on|off` CLI flag (default `off`). When on, the chain worker thread is the sole consumer for gossip blocks and gossip attestations; producer-side wrappers `submitBlock` / `submitGossipAttestation` enqueue on the worker queues. When off, handlers run synchronously on caller threads exactly as today.

Ownership contract on success: the caller transfers ownership of any heap-bearing payload (`SignedBlock`, `SignedAggregatedAttestation`) to the worker; the worker calls `Message.deinit` after dispatch. On `error.QueueFull` / `error.QueueClosed` / `error.ChainWorkerDisabled` the caller retains ownership and is responsible for cleanup.

End-to-end test wires the producer thread → BlockQueue → ChainWorker.runLoop → vtable thunk → `BeamChain.onBlock` + `onBlockFollowup`, polls until the post-state appears in the states map.

### Commit 4 (`cf12167`) — drop `BorrowedState` rwlock under `--chain-worker=on`

When chain-worker is the sole writer, cross-thread readers no longer need the shared rwlock to gate borrow lifetime — the `RcBeamState` refcount IS the lifetime mechanism.

`BorrowedState` extended:
- New variant `Backing.none` (lock-free borrow, lifetime gated by `acquired_rc` only).
- New optional field `acquired_rc: ?*const RcBeamState`. `deinit` releases it AFTER dropping any backing lock so `release()` (which may run the freeing path inline if refcount → 0) is OUTSIDE the critical section.

`statesGet` and `getFinalizedState` (cache-hit and DB-load paths) switch to `tryAcquire`-then-drop-lock when `chain_worker != null`, returning a `Backing.none` borrow whose deinit releases the rc. Under `--chain-worker=off` the lock-based path is unchanged (kill-switch backward-compat).

`statesCommitKeepExisting` INTENTIONALLY keeps holding the exclusive lock across its borrow even under `chain_worker_enabled` — the PR #820/#803 cross-call write-barrier rationale (DB write + `forkChoice.confirmBlock` must observe the in-map pointer atomically) still applies. Documented inline.

`cached_finalized_state` migrated from raw `*BeamState` to `*RcBeamState` (Option B from the PR review thread) for uniformity with the in-memory states map.

New tests cover: `Backing.none` + `acquired_rc` deinit (rc released, no leak), `cloneAndRelease` against a `Backing.none` borrow (rc released as part of the success-path deinit), `statesGet` returning `Backing.none` borrow under `chain_worker_enabled`, concurrent `statesGet` while a writer takes the exclusive lock (would deadlock under the old shared-held-for-borrow shape).

### Commit 5 (`72d0244`) — `lean_chain_state_refcount_distribution` histogram

Scrape-time observer iterates `BeamChain.states` under shared lock and samples `rc.count()` for each entry. Buckets `[1, 2, 4, 8, 16, 32, +Inf]`. Typical value 1 (writer-only); transient 2-4 under reader concurrency; values >16 indicate leaked acquires.

Wired via a new context-bearing scrape refresher slot `registerScrapeRefresherCtx(ctx, fn(?*anyopaque) void)` parallel to the existing `void → void` slot. Trampoline casts back to `*BeamChain` so the observer can iterate the chain's in-memory map without forcing the chain into a global.

Lock-hold minimization: stack-buffered snapshot (size 128) for the iteration; observe outside the lock. On overflow falls back to observing under the lock (graceful degradation — observe is just a bucket increment).

Lifecycle: registered in `BeamNode.init` after `startChainWorker` (chain at final heap address); cleared in `BeamChain.deinit` BEFORE any chain state is torn down so the metrics endpoint cannot call back into freed memory mid-deinit.

Audit-pattern test: 3 rounds of N observations (writer-only → +N readers → back to writer-only) → scrape `/metrics` → assert the `_count` line grew by exactly `3*N`. Includes a tiny Prometheus-text parser for the count assertion.

### Commit 6 (`df81046`) — queue-saturation stress mode (c-2c part 1)

New mode in `pkgs/node/src/stress.zig` driven by `ZEAM_STRESS_MODE=saturation`. 4 block-producer threads + 4 attestation-producer threads slam `submitBlock` / `submitGossipAttestation` against a single chain-worker drainer.

Producers correctly handle `error.QueueFull` (block path retains ownership and deinits the clone per the c-2b commit-3 contract; attestation path is value-typed today so QueueFull just drops the value).

Assertions at run end:
1. `attempts == ok + queue_full + other_err` per queue (no lost producer-side accounting).
2. `queue_full > 0` on both queues (otherwise we never saturated — the test would pass trivially).
3. `send_ok > 0` on both queues (otherwise the worker is deadlocked behind something other than the queue).
4. `other_err == 0` (no unexpected `submitBlock` / `submitGossipAttestation` error tag).

Plus a saturation-specific watchdog that catches producer-side deadlock (no progress in either send_ok OR queue_full counters).

New build steps:
- `zig build stress-saturation` — operator-driven full run (~30s default).
- `zig build stress-quick-saturation` — 10s, **wired into `zig build test`** alongside the existing `stress-quick`.

Verified locally with a 10s `zig build stress-quick-saturation`:

```
=== saturation run summary ===
duration: 10.0s
total attempts: 7746 (773 attempts/sec)
  block: attempts=3016 ok=455 queue_full=2561 other_err=0
  attn : attempts=4730 ok=1037 queue_full=3693 other_err=0
errs g=0 a=0 b=0 queue_full_b=2561 queue_full_a=3693
clean exit — chain-worker queues saturated and drained as expected, no UAF/deadlock observed
```

## Verification

```
$ zig build test
...
All 132 tests passed.
clean exit — no panics, no UAFs, no deadlocks observed
```

Test count: 119 (slice c-2a base) → 132 in this PR (+13 across all 6 commits). `stress-quick` (30s) and `stress-quick-saturation` (10s) both wired into `zig build test` and green. CI green on `df81046`.

## Out of scope (deferred follow-ups)

- **Burn-in (c-2c part 2):** devnet4 ≥24h with `--chain-worker=on` from `main` post-merge. NOT gated on this PR.
- **Default flip:** separate one-line PR after burn-in flips `--chain-worker` default from `off` to `on`.
- **`TODO(slice-c-2c)` lock-hold option (b):** marker in `statesCommitKeepExisting` for moving the kept-existing `rc.release()` outside the exclusive lock if devnet shows a `zeam_lock_hold_seconds{site="onBlock.commit"}` p99 regression vs slice (b) baseline. Currently option (a) — accept the inline release, observe via the existing metric — per the PR review thread.
- Double-release `prev > 0` assert remains best-effort (doc-only; flagged in c-2a).
- `RcBeamStateWriter` / `RcBeamStateReader` two-distinct-types refactor stays as escalation path.

Closes/relates to #803 (slice c-2b + c-2c part 1).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -509,6 +509,45 @@ pub fn build(b: *Builder) !void {
     stress_quick_step.dependOn(&run_stress_quick.step);
     test_step.dependOn(&run_stress_quick.step);
 
+    // -----------------------------------------------------------------
+    // `stress-saturation` and `stress-quick-saturation`: chain-worker
+    // queue saturation harness (slice c-2c commit 6 of #803).
+    //
+    // The full `stress-saturation` step is operator-driven (~30s
+    // default). The quick variant is wired into `zig build test` so
+    // CI catches:
+    //   * Producer-side accounting drift (attempts != ok+qfull+err).
+    //   * Backpressure regression (queue never fills — either the
+    //     producers are too slow or the queue capacity got bumped
+    //     without a corresponding bump to producer count).
+    //   * Worker-drain regression (queue fills but never drains —
+    //     classic worker-thread deadlock).
+    //   * Any unexpected `submitBlock` / `submitGossipAttestation`
+    //     error tag (today only `QueueClosed` and
+    //     `ChainWorkerDisabled` are non-`QueueFull`).
+    //
+    // Both steps reuse the same `stress_exe` artifact — the harness
+    // dispatches on `ZEAM_STRESS_MODE=saturation` set here.
+    const run_stress_saturation = b.addRunArtifact(stress_exe);
+    run_stress_saturation.setEnvironmentVariable("ZEAM_STRESS_MODE", "saturation");
+    if (b.args) |args| run_stress_saturation.addArgs(args);
+    const stress_saturation_step = b.step(
+        "stress-saturation",
+        "Run the chain-worker queue saturation harness (issue #803 slice c-2c)",
+    );
+    stress_saturation_step.dependOn(&run_stress_saturation.step);
+
+    const run_stress_quick_saturation = b.addRunArtifact(stress_exe);
+    run_stress_quick_saturation.setEnvironmentVariable("ZEAM_STRESS_MODE", "saturation");
+    run_stress_quick_saturation.setEnvironmentVariable("ZEAM_STRESS_DURATION_SECS", "10");
+    run_stress_quick_saturation.setEnvironmentVariable("ZEAM_STRESS_WATCHDOG_SECS", "15");
+    const stress_quick_saturation_step = b.step(
+        "stress-quick-saturation",
+        "Run a 10s chain-worker queue saturation harness (CI gate, slice c-2c)",
+    );
+    stress_quick_saturation_step.dependOn(&run_stress_quick_saturation.step);
+    test_step.dependOn(&run_stress_quick_saturation.step);
+
     // CLI integration tests (separate target) - always create this test target
     const cli_integration_tests = b.addTest(.{
         .root_module = b.createModule(.{
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -75,6 +75,10 @@ pub const NodeCommand = struct {
     @"attestation-committee-count": ?u64 = null,
     @"aggregate-subnet-ids": ?[]const u8 = null,
     @"db-backend": database.Backend = .rocksdb,
+    /// Slice c-2b commit 3 of #803: route producer-side gossip
+    /// handlers through the chain-worker queue. Default `false`
+    /// preserves slice-(b) synchronous behavior.
+    @"chain-worker": bool = false,
 
     pub const __shorts__ = .{
         .help = .h,
@@ -98,6 +102,7 @@ pub const NodeCommand = struct {
         .@"attestation-committee-count" = "Number of attestation committees (subnets); overrides config.yaml ATTESTATION_COMMITTEE_COUNT",
         .@"aggregate-subnet-ids" = "Comma-separated list of subnet ids to additionally subscribe and aggregate gossip attestations (e.g. '0,1,2'); adds to automatic computation from validator ids",
         .@"db-backend" = "Database backend to use for on-disk state: 'rocksdb' (default) or 'lmdb'",
+        .@"chain-worker" = "Route gossip block + attestation handlers through the dedicated chain-worker thread. Off by default; the synchronous path stays in place when disabled.",
         .help = "Show help information for the node command",
     };
 };
--- a/pkgs/cli/src/node.zig
+++ b/pkgs/cli/src/node.zig
@@ -94,6 +94,11 @@ pub const NodeOptions = struct {
     attestation_committee_count: ?u64 = null,
     max_attestations_data: ?u8 = null,
     db_backend: database.Backend = .rocksdb,
+    /// Slice c-2b commit 3 of #803: route producer-side gossip
+    /// handlers through the chain-worker queue. Default `false`
+    /// preserves slice-(b) synchronous behavior. Surfaced as
+    /// `--chain-worker` on the `zeam node` CLI.
+    chain_worker_enabled: bool = false,
 
     pub fn deinit(self: *NodeOptions, allocator: std.mem.Allocator) void {
         for (self.bootnodes) |b| allocator.free(b);
@@ -360,6 +365,7 @@ pub const Node = struct {
             .is_aggregator = options.is_aggregator,
             .aggregation_subnet_ids = options.aggregation_subnet_ids,
             .thread_pool = self.thread_pool,
+            .chain_worker_enabled = options.chain_worker_enabled,
         });
         errdefer self.beam_node.deinit();
 
@@ -755,6 +761,7 @@ pub fn buildStartOptions(
     opts.hash_sig_key_dir = hash_sig_key_dir;
     opts.checkpoint_sync_url = node_cmd.@"checkpoint-sync-url";
     opts.is_aggregator = node_cmd.@"is-aggregator";
+    opts.chain_worker_enabled = node_cmd.@"chain-worker";
 
     // Parse --aggregate-subnet-ids (comma-separated list of subnet ids, e.g. "0,1,2")
     // Require --is-aggregator to be set when --aggregate-subnet-ids is provided.
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -145,6 +145,11 @@ const Metrics = struct {
     lean_chain_queue_dropped_total: LeanChainQueueDroppedCounter,
     lean_chain_queue_depth: LeanChainQueueDepthGauge,
     lean_chain_worker_loop_iters_total: LeanChainWorkerLoopItersCounter,
+    // Slice c-2b commit 5 of #803: distribution of refcount values across
+    // map-resident BeamState entries at scrape time. Sampled by the chain
+    // via `recordChainStateRefcountDistribution` registered as a
+    // context-bearing scrape refresher (see `registerScrapeRefresherCtx`).
+    lean_chain_state_refcount_distribution: LeanChainStateRefcountDistributionHistogram,
 
     const ChainHistogram = metrics_lib.Histogram(f32, &[_]f32{ 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10 });
     const StateTransitionHistogram = metrics_lib.Histogram(f32, &[_]f32{ 0.25, 0.5, 0.75, 1, 1.25, 1.5, 2, 2.5, 3, 4 });
@@ -193,6 +198,14 @@ const Metrics = struct {
     const LeanChainQueueDroppedCounter = metrics_lib.CounterVec(u64, struct { queue: []const u8 });
     const LeanChainQueueDepthGauge = metrics_lib.GaugeVec(u64, struct { queue: []const u8 });
     const LeanChainWorkerLoopItersCounter = metrics_lib.Counter(u64);
+    // Refcount-distribution buckets [1, 2, 4, 8, 16, 32, +Inf]. Typical
+    // value is 1 (writer-only); transient 2-4 under reader concurrency;
+    // values >16 indicate leaked acquires (a reader forgot to release
+    // its borrow). The +Inf bucket is implicit in the Histogram's tail.
+    const LeanChainStateRefcountDistributionHistogram = metrics_lib.Histogram(
+        f32,
+        &[_]f32{ 1, 2, 4, 8, 16, 32 },
+    );
     // Fork-choice store gauge types
     const LeanGossipSignaturesGauge = metrics_lib.Gauge(u64);
     const LeanLatestNewAggregatedPayloadsGauge = metrics_lib.Gauge(u64);
@@ -609,6 +622,7 @@ pub fn init(allocator: std.mem.Allocator) !void {
         .lean_chain_queue_dropped_total = try Metrics.LeanChainQueueDroppedCounter.init(allocator, io, "lean_chain_queue_dropped_total", .{ .help = "Producer trySend rejections on the chain-worker queues, labeled by queue (block|attestation)." }, .{}),
         .lean_chain_queue_depth = try Metrics.LeanChainQueueDepthGauge.init(allocator, io, "lean_chain_queue_depth", .{ .help = "Instantaneous depth of the chain-worker queues, labeled by queue (block|attestation)." }, .{}),
         .lean_chain_worker_loop_iters_total = Metrics.LeanChainWorkerLoopItersCounter.init("lean_chain_worker_loop_iters_total", .{ .help = "Cumulative chain-worker loop iterations. External watchdogs use the delta between scrapes to detect worker stalls." }, .{}),
+        .lean_chain_state_refcount_distribution = Metrics.LeanChainStateRefcountDistributionHistogram.init("lean_chain_state_refcount_distribution", .{ .help = "Distribution of refcount values across map-resident BeamState entries at scrape time. Typical value 1 (writer-only); transient 2-4 under reader concurrency; values >16 indicate leaked acquires." }, .{}),
     };
 
     // Initialize validators count to 0 by default (spec requires "On scrape" availability)
@@ -669,6 +683,29 @@ pub fn registerScrapeRefresher(refresher: ?*const fn () void) void {
     g_scrape_refresher = refresher;
 }
 
+/// Optional context-bearing pre-scrape refresher. Slice c-2b commit 5 of
+/// #803 (lean_chain_state_refcount_distribution) needed a callback that
+/// receives a `*BeamChain` so the observer could iterate the chain's
+/// in-memory states map under its shared lock and sample each rc's
+/// refcount. The first-form `g_scrape_refresher` slot above is
+/// `void → void` (used for FFI-backed atomic counters that need no
+/// context); rather than coerce the chain into a global, we expose a
+/// parallel slot that takes an opaque pointer back to the caller. The
+/// two slots are independent and both run on every scrape (the FFI
+/// refresher first, then the context-bearing one).
+var g_scrape_refresher_ctx: ?*const fn (?*anyopaque) void = null;
+var g_scrape_refresher_ctx_ptr: ?*anyopaque = null;
+
+/// Register (or replace) a context-bearing scrape refresher. Pass `null`
+/// for `refresher` to clear (the ctx pointer is also cleared).
+pub fn registerScrapeRefresherCtx(
+    ctx: ?*anyopaque,
+    refresher: ?*const fn (?*anyopaque) void,
+) void {
+    g_scrape_refresher_ctx = refresher;
+    g_scrape_refresher_ctx_ptr = if (refresher == null) null else ctx;
+}
+
 /// Writes metrics to a writer (for Prometheus endpoint).
 pub fn writeMetrics(writer: *std.Io.Writer) !void {
     if (!g_initialized) return error.NotInitialized;
@@ -682,6 +719,10 @@ pub fn writeMetrics(writer: *std.Io.Writer) !void {
     // Pull in any externally-owned counters (e.g. Rust-side libp2p drops)
     // before serializing so each scrape returns up-to-date values.
     if (g_scrape_refresher) |refresher| refresher();
+    // Context-bearing refresher (slice c-2b commit 5 of #803): runs
+    // after the void-refresher so a future contributor can rely on
+    // ordering when the two refreshers' outputs share buckets/labels.
+    if (g_scrape_refresher_ctx) |refresher| refresher(g_scrape_refresher_ctx_ptr);
 
     try metrics_lib.write(&metrics, writer);
 }
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -152,8 +152,15 @@ pub const BeamChain = struct {
     ///
     /// Lock-free monotonic counter; safe to read from any thread.
     states_kept_existing_count: std.atomic.Value(u64) = std.atomic.Value(u64).init(0),
-    // Cached finalized state loaded from database (separate from states map to avoid affecting pruning)
-    cached_finalized_state: ?*types.BeamState = null,
+    // Cached finalized state loaded from database (separate from states
+    // map to avoid affecting pruning).
+    //
+    // Slice c-2b commit 4 of #803: migrated from raw `*types.BeamState`
+    // to `*RcBeamState` (Option B from the PR #828 review thread) so the
+    // cache-hit path in `getFinalizedState` can use the same
+    // tryAcquire-then-drop-lock dance as `statesGet` when chain-worker
+    // is enabled. Mutation is still gated by `events_lock`.
+    cached_finalized_state: ?*RcBeamState = null,
     // Cache for validator public keys to avoid repeated SSZ deserialization during signature verification.
     // Significantly reduces CPU overhead when processing blocks with many attestations.
     public_key_cache: xmss.PublicKeyCache,
@@ -569,10 +576,16 @@ pub const BeamChain = struct {
         }
         self.states.deinit();
 
-        // Clean up cached finalized state if present
-        if (self.cached_finalized_state) |cached_state| {
-            cached_state.deinit();
-            self.allocator.destroy(cached_state);
+        // Clean up cached finalized state if present.
+        // Slice c-2b commit 4: cached_finalized_state is an
+        // RcBeamState now — release drives state.deinit + destroy when
+        // refcount reaches 0. Under chain_worker_enabled the cache may
+        // have outstanding reader acquires; the rc keeps the underlying
+        // allocation alive until the last reader releases. (At deinit
+        // time we are single-threaded by contract, so refcount=1 in
+        // practice.)
+        if (self.cached_finalized_state) |cached_rc| {
+            cached_rc.release();
         }
 
         // Clean up public key cache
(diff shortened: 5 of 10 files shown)


zclawz:
**Commit 3 pushed: `4709e7b` — chain-worker routing + `--chain-worker` CLI flag.**

3 of 6 commits done. PR remains draft until commits 4-6 land.

### What this commit does

Producer-side handlers can now route through the c-1 chain_worker queue under a new `--chain-worker=on|off` CLI flag (default `off`). When on, the chain worker thread is the sole consumer for gossip blocks and gossip attestations; producers fire-and-forget via `BeamChain.submit*` wrappers. When off, `submit*` calls the chain method directly on the producer thread — zero behavioural change vs pre-c-2b.

### Plumbing

- **`Handlers` vtable in `chain_worker.zig`**: 5 function pointers (one per Message variant), each takes `ctx: *anyopaque` plus the variant payload. Avoids a circular import — chain_worker.zig CANNOT import chain.zig because chain.zig already imports chain_worker.zig. chain.zig fills the vtable at `init` with bound handlers casting back to `*BeamChain`; each handler swallows errors via `logger.warn` so producer threads stay fire-and-forget.

- **`ChainOpts.chain_worker_enabled: bool = false`** + `BeamChain.chain_worker: ?*ChainWorker`. When the opt is true, init heap-allocates the worker, wires the vtable, and calls `worker.start()` BEFORE returning so the worker thread is live by the time producer wrappers are reachable. `deinit` does `stop() + deinit() + destroy()` if present.

- **3 new producer wrappers**:
  - `submitBlock(signed_block, prune_fc)` — worker mode: builds `on_block` Message, sendBlock(); queue takes ownership of `signed_block`. Direct mode: calls `onBlock + onBlockFollowup` synchronously.
  - `submitGossipAttestation(gossip)` — worker mode: sendAttestation. Direct mode: `onGossipAttestation`.
  - `submitGossipAggregatedAttestation(agg)` — same pattern.

### Migration scope (intentionally narrow)

Only the **two libp2p gossip dispatch sites** in `chain.zig` migrated:

| Path | Before | After |
|---|---|---|
| block-gossip | `onBlock + onBlockFollowup` direct | `submitBlock(cloned, true)` |
| attestation-gossip | `onGossipAttestation` direct | `submitGossipAttestation` |

Other callers (replay path, req-resp, `processPendingBlocks` tick, finalization followup) stay on the direct path. Migrating them is mechanical follow-up — deferring keeps this commit's blast radius bounded so reviewers can focus on the routing infrastructure.

### CLI wiring

- `pkgs/cli/src/main.zig`: parse `--chain-worker=on|off` (default `off`)
- `pkgs/cli/src/node.zig` + `pkgs/node/src/node.zig`: thread through `NodeOpts` → `ChainOpts.chain_worker_enabled`

### Integration test

`chain.test.chain-worker: end-to-end submitBlock advances state via the worker thread` (~210 LOC including a parallel control chain in off-mode):

1. Boots a BeamChain with `chain_worker_enabled = true`
2. Builds a child block, signs it, calls `submitBlock`
3. Polls (bounded wait + max iters) for the worker to drain: state map gains the post-state root, fork choice head advances
4. Asserts the new state is reachable via `statesGet`
5. Boots a second chain with `chain_worker_enabled = false`, submits the same block via the same wrapper, asserts the synchronous direct path produces an identical result
6. Cleanly shuts down both chains; worker's `stop()` joins its thread before `deinit()` frees the queues

The test is the proof that worker-mode and direct-mode produce behaviourally identical results on the happy path.

### Verified

```
...
41/124 chain.test.chain-worker: end-to-end submitBlock advances state via the worker thread...OK
...
All 124 tests passed.
EXIT:0
```

Test count: 123 → 124 (+1 integration test). All RcBeamState tests still pass; all pre-existing chain.zig integration suites still pass.

### Diff stats vs `9b3b036`

```
pkgs/cli/src/main.zig          |   5 +
pkgs/cli/src/node.zig          |   7 +
pkgs/node/src/chain.zig        | 565 ++++++++++++++++++++++++++++++++++++++++-
pkgs/node/src/chain_worker.zig | 100 ++++++--
pkgs/node/src/node.zig         |  18 ++
5 files changed, 669 insertions(+), 26 deletions(-)
```

### Next

**Commit 4**: drop the `BorrowedState` rwlock when chain-worker is sole writer. Cross-thread readers switch to `acquireConst` + `tryAcquire`. This is where the lock-free reader path actually goes live and the c-2b end-game `zeam_lock_*` histogram drops materialise.

zclawz:
Reviewing now — applying fix shape (a) for the `post_state_rc` leak (move errdefer + `rc_owned` flag into `statesCommitKeepExisting` to mirror `statesPutExclusive`), looking at the lock-hold regression on the kept-existing path, and flagging the `getFinalizedState` cache-path heterogeneity for commit 4. Will report back with diffs.

zclawz:
Pushed `fa7cce1` addressing the leak + lock-hold + cache-heterogeneity points, plus a sketch of the commit-4 BorrowedState shape change for the third comment about `statesGet` not taking an rc acquire today. Walking each item:

## 1. `post_state_rc` leak on `statesCommitKeepExisting` OOM — **fixed**

Picked **fix shape (a)** as you suggested — consistency with `statesPutExclusive` mattered. Diff in `statesCommitKeepExisting`:

```zig
+ var rc_owned = true;
  errdefer {
      self.states_lock.unlock();
      t.released();
+     if (rc_owned) rc.release();
  }
  const gop = try self.states.getOrPut(root);
+ rc_owned = false;  // helper now owns disposal: either release() below or transfer into map
  const effective_rc: *RcBeamState = if (gop.found_existing) blk: {
      rc.release();
      break :blk gop.value_ptr.*;
  } else blk: {
      gop.value_ptr.* = rc;
      break :blk rc;
  };
```

Set `rc_owned = false` BEFORE the disposing branch (release-on-found / hand-off-on-new) so the errdefer can never double-release. Caller in `onBlock` now relies entirely on the helper for rc cleanup on every error path — same shape as `statesPutExclusive`.

**Regression test added** (`chain.statesCommitKeepExisting: getOrPut OOM releases caller rc (no leak)`):
- Wraps `beam_chain.states.allocator` with `std.testing.FailingAllocator{ .fail_index = 0 }` after init.
- Forces `unmanaged.available = 0` to deterministically trigger `growIfNeeded` on the next insert (otherwise the map's spare capacity would absorb the insert without ever calling the allocator).
- Calls `statesCommitKeepExisting` with a fresh root, expects `error.OutOfMemory`.
- `testing.allocator`'s leak detector is the witness: the test passes only when the helper releases the rc on the OOM path. Confirmed by running the test before the fix (FAIL: leaked rc) and after (PASS).

## 2. `rc.release()` lock-hold expansion on `found_existing` — **option (a), documented inline + TODO for option (b)**

Took option (a) — accept the inline release under the exclusive lock — and made the trade-off explicit in a comment block in the helper. Key points captured there:

- Today the only acquires on map-resident states come from the chain worker thread, which is the SAME thread executing the commit, so `release()` on `found_existing` always drops 1→0 inline and synchronously runs `state.deinit() + allocator.destroy(rc)`. That work IS in the critical section now, vs pre-c-2b's caller-side free outside the lock.
- The `LockTimer.start("states", site)` already tags hold-spans per call site, so `zeam_lock_hold_seconds{lock="states", site="onBlock.commit"}` differentiates this from `produceBlock.commit` and others. **No new instrumentation needed** — just a slice (b) → (c-2b) baseline comparison on devnet.
- Added `TODO(slice-c-2c)` referring to option (b) — return the to-be-released rc alongside the borrow, caller releases after dropping the lock — gated on devnet showing a real `onBlock.commit` p99 regression. Tracking-only until then; bisecting an API-surface change with no measured cause is bad engineering.

If you want option (b) as a pre-emptive structural fix instead of a wait-and-see, say the word and I'll do it in the same commit before c-2c lands. My read is the inline cost is low (allocator work, single-thread, no contention) and the TODO is enough until the data says otherwise — but you've been right about subtle perf regressions on this PR before.

## 3. `getFinalizedState` cache path uses raw `*BeamState` — **flagged for commit 4**

Confirmed and intentional for commit 2 — `cached_finalized_state: ?*types.BeamState` is mutated under `events_lock` (tier 5c) only, with no concurrent readers that need rc semantics today. The borrow returned for the cache hit / DB-load paths binds to `events_mutex` (not `states_lock`), so the lock-based lifetime story is consistent for that path even after we drop the states-side rwlock in commit 4.

**Commit 4 plan for the cache path** (will write this into the commit message + PR body when c-4 lands):

- **Option A — keep the cache lock-based:** leave `cached_finalized_state` as `*BeamState`, keep the `events_mutex` binding on the BorrowedState for cache hits. Pro: minimal change, no new rc bookkeeping for a single-mutator field. Con: the borrow heterogeneity becomes permanent — readers hitting the in-memory map get `tryAcquire`-based borrows, readers hitting the cache stay lock-based. HTTP / RPC threads don't care which they got, but it's a footgun if a future caller assumes "all borrows are rc-backed".
- **Option B — migrate the cache to RcBeamState:** make it `?*RcBeamState`, take a `tryAcquire` on the cache hit path before unlocking events_mutex, return a borrow with `.events_mutex` released-on-deinit + an `acquired_rc: ?*RcBeamState` field that `borrow.deinit()` releases. The DB-load path also wraps the loaded state in `RcBeamState.create` before storing.

I'm leaning **Option B** for c-4 because it makes BorrowedState's lifetime story uniform across all paths (rc-acquired pointers, lock no longer load-bearing), which is the whole point of the commit-4 migration. Cost is one rc create on cache miss + one tryAcquire/release on cache hit — both negligible vs the DB load and the full state walk.

**Will sketch in c-4 PR body before requesting review on it.**

## 4. `statesGet` exposes `&rc.state` without acquiring — **commit-4 BorrowedState shape sketch**

You're right that the current `BorrowedState` has no field to carry the acquired rc reference, and commit 4 needs to extend it. Concrete sketch I plan to land on the c-4 commit:

```zig
pub const BorrowedState = struct {
    backing: Backing,
    /// NEW in c-4: when set, deinit calls release() on this rc AFTER
    /// dropping any backing lock. None for cache-path borrows that stay
    /// lock-based (Option A above) or for legacy lock-only paths during
    /// the c-2b → c-4 transition. Set on every statesGet borrow once
    /// the rwlock is dropped.
    acquired_rc: ?*const RcBeamState = null,
    released: bool = false,
    tier5_held: bool = false,
    timer: ?LockTimer = null,

    pub const Backing = union(enum) {
        states_shared_rwlock: *zeam_utils.SyncRwLock,
        states_exclusive_rwlock: *zeam_utils.SyncRwLock,
        events_mutex: *zeam_utils.SyncMutex,
        none,  // NEW in c-4: lock-free borrow, lifetime gated entirely by acquired_rc
    };

    pub fn deinit(self: *BorrowedState) void {
        if (self.released) return;
        self.released = true;
        switch (self.backing) {
            .states_shared_rwlock => |rw| rw.unlockShared(),
            .states_exclusive_rwlock => |rw| rw.unlock(),
            .events_mutex => |m| m.unlock(),
            .none => {},
        }
        if (self.tier5_held) { self.tier5_held = false; leaveTier5(); }
        if (self.timer) |*t| t.released();
        // NEW in c-4: release the rc AFTER unlocking, so the release
        // (which may drop refcount to 0 and run deinit) happens outside
        // any critical section.
        if (self.acquired_rc) |rc| rc.release();
    }

    // cloneAndRelease: unchanged signature, but the implementation
    // releases acquired_rc as part of self.deinit() at the end.
};
```

`statesGet` after c-4 will look like:

```zig
pub fn statesGet(self: *Self, root: types.Root) ?BorrowedState {
    var t = LockTimer.start("states", "statesGet");
    self.states_lock.lockShared();  // still shared for the lookup itself
    t.acquired();
    defer self.states_lock.unlockShared();  // drop AFTER tryAcquire
    defer t.released();
    if (self.states.get(root)) |rc| {
        // tryAcquire instead of acquire: a concurrent pruner may have
        // just dropped the map's reference. If the refcount went to 0
        // we observe the release racing with our get; tryAcquire
        // returns null, we return null too (state was just pruned).
        const acquired = rc.tryAcquire() orelse return null;
        return BorrowedState{
            .state = &acquired.state,
            .backing = .none,             // no lock backing — rc gates lifetime
            .acquired_rc = acquired,
            .timer = null,                // hold-span ends when lock drops above
        };
    }
    return null;
}
```

Splitting into two BorrowedState shapes — one rc-backed, one lock-backed — gets uglier than a single shape with an optional rc field, IMO. Single shape with `Backing.none` + `acquired_rc` keeps the ergonomics ("caller calls `borrow.deinit()`, doesn't care which path produced it") and the API surface stable across the c-2b → c-4 transition. Existing call sites don't need to change; the borrow internals do.

**Commit 4 risks I want to flag now so reviewers can pre-empt them:**

1. **`tryAcquire` returns null mid-pruner-race** — `statesGet` callers today just see "not in map" if pruning won the race; `tryAcquire null` is functionally the same outcome (return null), just observed at the rc level instead of the map level. Existing call sites (HTTP/RPC mostly) handle the null already.
2. **`statesCommitKeepExisting` already returns a `BorrowedState` with the EXCLUSIVE lock held** — that path stays lock-based for c-4 because it needs the cross-call write barrier between commit and the post-commit DB write + `forkChoice.confirmBlock` (PR #820 / #803 rationale). So we keep `Backing.states_exclusive_rwlock` on that path and DON'T set `acquired_rc`. Need an explicit comment in the helper that the borrow type is the same but the lifetime contract is different per call site.
3. **Cache path heterogeneity** — Option A vs B above. Will pick before c-4 lands.

Tests on the new commit:
- `zig build test` — **all 125 pass clean**, including the new OOM regression test (#38) and the existing concurrency tests #39–#42. Commit `fa7cce1`.

Pushed to `feat/agent-slice-c-2b-c-2c-migration-and-stress`. Ready for re-review.

zclawz:
Pushed `c19e5f2` addressing the cosmetic + test-naming items.

## 1. Centralised `produceBlock`/`onBlock` post_state→rc dance

Added `BeamChain.wrapOwnedStateIntoRc(post_state, &gate_consumed)`. Single helper, single ownership story:

```zig
fn wrapOwnedStateIntoRc(
    self: *Self,
    post_state: *types.BeamState,
    gate_consumed: *bool,
) !*RcBeamState {
    const value = post_state.*;
    self.allocator.destroy(post_state);
    gate_consumed.* = true;  // BEFORE the fallible call — see docstring
    return RcBeamState.create(self.allocator, value);
}
```

Both call sites now share the gate-set-BEFORE-create pattern. Switched `produceBlock` from `?*BeamState` nulling to a `bool post_state_consumed` gate to match `onBlock`'s shape exactly. `onBlock`'s caller-supplied (NOT-owned) path stays inline — it sszClones instead of moving, so there is no wrapper to free and no shared dance to factor out.

Diff at the call sites is now near-identical:

```zig
// produceBlock
const post_state_rc = try self.wrapOwnedStateIntoRc(post_state, &post_state_consumed);
try self.statesPutExclusive("produceBlock.commit", block_root, post_state_rc);

// onBlock (owned branch)
post_state_rc = try self.wrapOwnedStateIntoRc(post_state, &post_state_settled);
// ... !post_state_owned branch unchanged: sszClone + create inline
```

Helper docstring spells out the post-call states for success/error/caller-gate semantics so a future reader doesn't have to re-derive the BEFORE-create timing. PR #828's `fa7cce1` regression test (`statesCommitKeepExisting: getOrPut OOM releases caller rc`) still passes — no behaviour change, just unification.

## 2. Renamed misleading test + rewrote docstring

`RcBeamState.tryAcquire: race against release-to-zero — producer + freer` → `RcBeamState.tryAcquire: succeeds while caller holds a pre-bump (no UAF under release-to-zero of separate ref)`.

You're right — every Trier holds its own pre-bump (`rc.acquireWriter()` on the spawn thread), so refcount is ≥ 1 for the entire `Trier.run` body and the `@panic` on null is the reveal that the test ASSUMES success. The new docstring explicitly enumerates:

- **What the test DOES prove**: no UAF (testing.allocator's leak detector + double-free detector), refcount accounting under contention (every successful tryAcquire matched by exactly one release), CAS-loop safety under concurrent release pressure on a SEPARATE reference (the creator's drop racing the Triers').
- **What the test CANNOT prove (and why no test can)**: tryAcquire returning null in a multi-threaded race against release-to-zero is unreachable by safe code. tryAcquire's safety preconditions (lines 318–330) require the caller to hold a reference at the moment of the call; calling tryAcquire on a freed wrapper IS a UAF. The white-box test (`returns null after refcount reaches 0`) covers the null path with manually-set refcount=0 + skipped destroy(), which is the only safe shape.

The safety preconditions on tryAcquire's docstring are the contract surface — that's the right shape, no test needs to "prove" something the API doesn't promise.

## 3. tryAcquire overflow assert (✓)

Acknowledged, no code change. Same pattern as `acquireWriter`'s overflow check; consistent.

## Verification

`zig build test` — **all 125 pass clean** on `c19e5f2`. No panics, no UAFs, no deadlocks. `chain.zig` chain-worker / re-import pressure / finalization-race / submitBlock e2e tests all pass on the new helper. Pushed.

zclawz:
**c-4 landed: drop `BorrowedState` rwlock under `--chain-worker=on`** ([cf12167](https://github.com/blockblaz/zeam/pull/828/commits/cf12167))

Adds `BorrowedState.Backing.none` + `acquired_rc` field. `statesGet` and `getFinalizedState` cache-hit / DB-load paths switch to tryAcquire-then-drop-lock when `self.chain_worker != null`; the legacy lock-held-for-borrow path stays for `--chain-worker=off` (kill-switch). `statesCommitKeepExisting` deliberately remains exclusive-held-across-borrow per the PR #820/#803 cross-call write-barrier rationale.

`cached_finalized_state` migrated from raw `*BeamState` to `*RcBeamState` (Option B from review thread) so the cache hit uses the same shape as the in-memory map.

Tests (6 new): 4 `BorrowedState` cases for `Backing.none` (deinit-releases-rc, drives-final-free, cloneAndRelease OOM + success); 2 chain tests verifying the `Backing.none` shape under `startChainWorker()` and that a long-lived reader borrow does NOT block an exclusive writer (1s watchdog). `zig build test`: **all 131 tests passing**.

zclawz:
**c-5 landed: `lean_chain_state_refcount_distribution` histogram** ([72d0244](https://github.com/blockblaz/zeam/pull/828/commits/72d0244))

Scrape-time observer iterates `BeamChain.states` under shared lock and samples `rc.count()` for each entry. Buckets `[1, 2, 4, 8, 16, 32, +Inf]`.

**Wiring:** added a context-bearing scrape refresher slot (`registerScrapeRefresherCtx(ctx, fn(?*anyopaque) void)`) parallel to the existing `void → void` slot. Trampoline casts back to `*BeamChain` so the observer can iterate the chain's in-memory map without forcing the chain into a global.

**Lock-hold minimization:** stack-buffered snapshot (size 128) for the iteration; observe outside the lock. On overflow falls back to observing under the lock (graceful degradation — observe is just a bucket increment).

**Lifecycle:** registered in `BeamNode.init` after `startChainWorker` (chain at final heap address); cleared in `BeamChain.deinit` BEFORE any chain state is torn down so the metrics endpoint cannot call back into freed memory mid-deinit.

**Audit test:** 3 rounds of N observations (writer-only → +N readers → back to writer-only) → scrape `/metrics` → assert `_count` line grew by exactly `3*N`. Includes a tiny Prometheus-text parser for the count assertion.

All 132 tests pass. Commit 6 (queue-saturation stress mode) up next.

zclawz:
**c-6 landed: queue-saturation stress mode** ([df81046](https://github.com/blockblaz/zeam/pull/828/commits/df81046))

Mode dispatch via `ZEAM_STRESS_MODE=saturation`. New build steps:
- `zig build stress-saturation` — operator-driven full run (~30s default).
- `zig build stress-quick-saturation` — 10s, **wired into `zig build test`** alongside the existing `stress-quick`.

**What it exercises:** 4 block-producer threads + 4 attestation-producer threads slam `submitBlock` / `submitGossipAttestation` against a single chain-worker drainer. Producers correctly handle `error.QueueFull` (block path retains ownership and deinits the clone per the c-2b commit-3 contract; attestation path is value-typed so QueueFull just drops the value).

**Assertions at run end:**
1. `attempts == ok + queue_full + other_err` per queue (no lost producer-side accounting).
2. `queue_full > 0` on both queues (the test is meaningless if we never saturate).
3. `send_ok > 0` on both queues (otherwise the worker is deadlocked).
4. `other_err == 0` (no unexpected submit-error tag).

Plus a saturation-specific watchdog that catches producer-side deadlock (no progress in either send_ok OR queue_full) faster than the post-run assertions could.

**Verified locally** with a 10s `zig build stress-quick-saturation`:
```
=== saturation run summary ===
duration: 10.0s
total attempts: 7746 (773 attempts/sec)
  block: attempts=3016 ok=455 queue_full=2561 other_err=0
  attn : attempts=4730 ok=1037 queue_full=3693 other_err=0
errs g=0 a=0 b=0 queue_full_b=2561 queue_full_a=3693
clean exit — chain-worker queues saturated and drained as expected, no UAF/deadlock observed
```

---

## All 6 commits landed for c-2b + c-2c part 1

| # | SHA | Description |
|---|---|---|
| 1 | `5bf12dd` | `tryAcquire` upgrade-from-weak primitive on `RcBeamState` |
| 2 | `9b3b036` | `BeamChain.states` map value type: `*BeamState` → `*RcBeamState` |
| 3 | `4709e7b` | Producer-side handlers route through chain_worker queues; `--chain-worker` CLI flag |
| 4 | `cf12167` | Drop `BorrowedState` rwlock under `--chain-worker=on` |
| 5 | `72d0244` | `lean_chain_state_refcount_distribution` histogram metric |
| 6 | `df81046` | Queue-saturation stress mode (c-2c part 1) |

Plus three review-fix commits picked up along the way: `fa7cce1` (post_state_rc leak fix + regression test), `c19e5f2` (wrapOwnedStateIntoRc unification + tryAcquire test rename), and `1835008` (zig fmt).

**Test status:** all 132 tests pass clean, plus `stress-quick` (30s) and `stress-quick-saturation` (10s) both wired into `zig build test` and green.

**Open items deferred to follow-ups (per the original plan):**
- **Burn-in (c-2c part 2):** devnet4 ≥24h run with `--chain-worker=on` from `main` post-merge. NOT gated on this PR.
- **Default flip:** separate one-line PR after burn-in clean to flip `--chain-worker` default from `off` to `on`.
- **Lock-hold option (b):** TODO(slice-c-2c) marker in `statesCommitKeepExisting` for moving the kept-existing `rc.release()` outside the exclusive lock if devnet shows a `zeam_lock_hold_seconds{site="onBlock.commit"}` p99 regression vs slice (b) baseline.
- **Cache-path option A vs B (c-4 follow-up):** committed Option B (cache `*RcBeamState`) for uniformity. Reviewable within c-4's diff.

Ready for re-review.

ch4r10t33r (approved):
Looks good.

PR #830, opened 2026-05-06
node: --chain-worker on|off enum + flip default to on (#803 follow-up)

zclawz:
Two-part follow-up to [#828](https://github.com/blockblaz/zeam/pull/828) (slice c-2b + c-2c part 1 of [#803](https://github.com/blockblaz/zeam/issues/803)).

## What

1. **`--chain-worker` becomes a value-taking enum `{on, off}`** instead of a presence-only `bool`. Mirrors the existing `--db-backend rocksdb|lmdb` precedent in `pkgs/cli/src/main.zig` (which uses simargs's enum support).

2. **Compiled-in default flips from `.off` → `.on`.** Per @ch4r10t33r — shipping the chain-worker default-off would mean the burn-in tests the path nobody actually runs in prod. The kill-switch survives via an explicit `--chain-worker off`.

## Why

When I shipped #828 with `@"chain-worker": bool = false`, I documented the lean-quickstart side as `--chain-worker on` / `--chain-worker off`. **But simargs treats `bool` as presence-only** — bare `--chain-worker` flips it to `true`, and `on` / `off` would be parsed as a stray positional, breaking startup. lean-quickstart [#170](https://github.com/blockblaz/lean-quickstart/pull/170) (merged) and [#171](https://github.com/blockblaz/lean-quickstart/pull/171) were already wired for the value-taking shape, so the simpler fix is to make the zeam CLI match what's already deployed in lean-quickstart.

This PR is a strict superset of what would have been a separate "default-flip" PR after burn-in. Combining them: zeam v0.4.14 (the pre-deployment build) is broken on the chain-worker flag anyway — fix the CLI shape and flip the default in the same release.

## Diff

```
 pkgs/cli/src/main.zig   | 15 +++++++++++----
 pkgs/cli/src/node.zig   | 13 +++++++------
 pkgs/node/src/chain.zig | 13 +++++++------
 pkgs/node/src/lib.zig   |  1 +
 pkgs/node/src/node.zig  | 37 ++++++++++++++++++++++++++++++++-----
 5 files changed, 58 insertions(+), 21 deletions(-)
```

New public type: `node_lib.ChainWorkerMode = enum { on, off }` with `pub inline fn isEnabled(self) bool` helper. Replaces the `chain_worker_enabled: bool` field on `BeamNodeOpts` and `NodeOptions`. Doc comments in `chain.zig` + `cli/main.zig` + `cli/node.zig` updated to reflect the new shape.

## Verification

```
$ zig build install
Finished `release` profile [optimized] target(s) in 11.02s
EXIT:0

$ ./zig-out/bin/zeam node --help | grep chain-worker
      --chain-worker STRING        Route gossip block + attestation handlers through the dedicated chain-worker thread (slice c-2b/c-2c, zeam #803). Values: 'on' (default, prod path) or 'off' (kill-switch / legacy synchronous path). (valid: on|off)(default: on)

$ ./zig-out/bin/zeam node --chain-worker hello --custom_genesis /tmp 2>&1 | head -3
# (parser rejects InvalidEnumValue and prints help)

$ ./zig-out/bin/zeam node --chain-worker on --custom_genesis /nonexistent
# (parses cleanly, errors later on missing --node-id as expected)

$ ./zig-out/bin/zeam node --chain-worker off --custom_genesis /nonexistent
# (same — flag accepted)

$ zig build test
All 132 tests passed.
clean exit — no panics, no UAFs, no deadlocks observed
```

## Plan

Once this lands on `main`, cut **v0.4.15** with this fix + the c-2b default-on. That's the build the devnet4 burn-in restarts on (with `ZEAM_CHAIN_WORKER` _unset_ or `=on` — both yield the prod path).

If burn-in regresses, kill-switch is `export ZEAM_CHAIN_WORKER=off` on the spin-node host (already wired through the ansible role on lean-quickstart `main`).

### DIFF
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -76,9 +76,16 @@ pub const NodeCommand = struct {
     @"aggregate-subnet-ids": ?[]const u8 = null,
     @"db-backend": database.Backend = .rocksdb,
     /// Slice c-2b commit 3 of #803: route producer-side gossip
-    /// handlers through the chain-worker queue. Default `false`
-    /// preserves slice-(b) synchronous behavior.
-    @"chain-worker": bool = false,
+    /// handlers through the chain-worker queue. Default `.on`
+    /// post-c-2b/c-2c merge (PR #828 follow-up): the chain-worker
+    /// is the prod path; `.off` is the kill-switch.
+    ///
+    /// Value-taking flag: `--chain-worker on|off`. NOT a
+    /// presence-only bool, NOT `--chain-worker=on` syntax — the
+    /// simargs parser uses enum tag names verbatim, identical to
+    /// the `--db-backend rocksdb|lmdb` shape. See
+    /// `node_lib.ChainWorkerMode`.
+    @"chain-worker": node_lib.ChainWorkerMode = .on,
 
     pub const __shorts__ = .{
         .help = .h,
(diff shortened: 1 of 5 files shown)


zclawz:
Closing this — I was wrong about the CLI shape. I claimed v0.4.14's `bool` `--chain-worker` would parse `on`/`off` as a stray positional and break startup, but that's not what simargs does. The bool branch sets the flag to `true` and moves on; the `on` token after it gets silently ignored as a positional past the active sub-command. Verified against the v0.4.14 binary built from `main` @ `89505fe` — `--chain-worker on` parses cleanly and reaches `Missing required option 'node-id'`, past the flag-parse stage.

So the CLI shape fix in this PR is unnecessary, and I shouldn't have flipped the zeam compiled-in default either — @ch4r10t33r asked for the default flip in lean-quickstart only ([lean-quickstart#171](https://github.com/blockblaz/lean-quickstart/pull/171)), not in zeam. Apologies for the noise.

[lean-quickstart#171](https://github.com/blockblaz/lean-quickstart/pull/171) works as-is against v0.4.14: passing `--chain-worker on` flips the bool to `true`, `on` gets ignored. The deployed v0.4.13 image will still need an image bump to v0.4.14 before #171 is merged (since v0.4.13 doesn't recognise the flag at all and would error).

PR #841, opened 2026-05-07
node: buffer future-slot gossip blocks during clock lag (fixes #788)

### LINKED ISSUE #788: zeam_0: fork choice resets to finalized checkpoint under timing desync (FutureSlot rejection loop)
## Summary

During devnet-4 (2026-04-27 17:24 UTC run), `zeam_0` exhibited periodic **fork choice resets** — its reported head slot dropped back to the last finalized checkpoint (slot 93) while the rest of the network had advanced to slot 200–600+. These resets occurred multiple times throughout the run and caused lag spikes of **Behind=100–160 slots**. `zeam_1` running identically did not exhibit this behaviour and remained at 1–2 slots lag throughout.

## Observed behaviour

From `zeam_0` CHAIN STATUS logs:

```
Current Slot: 200 | Head Slot: 93  | Behind: 107   ← head reset to finalized checkpoint
Current Slot: 220 | Head Slot: 93  | Behind: 127
Current Slot: 249 | Head Slot: 93  | Behind: 156
Current Slot: 260 | Head Slot: 260 | Behind: 0     ← sudden snap-back
```

This pattern repeated at least 4 times during the run. Between resets, `zeam_0` would operate normally (0–3 slots behind), then suddenly drop its head back to the finalized checkpoint slot.

## Associated errors

Concurrent with the resets, `zeam_0` logged:

- `error.FutureSlot` — blocks imported from peers rejected because zeam considers them to be in the future
- `AttestationTooFarInFuture` — attestations rejected for the same reason
- `skipping onInterval for node as chain is already ahead at time=X of the misfired interval time=X`

The `FutureSlot` errors suggest a **clock/timing desync**: after a late `onInterval` fires and processes a burst of slots, zeam's internal clock view jumps forward. Incoming gossip blocks from peers (which are timestamped to the actual wall clock) are then treated as future blocks and rejected, causing zeam to lose its fork-choice head and fall back to the last known-good finalized checkpoint.

## Impact

- `zeam_0` effectively dropped out of attesting during each reset window
- Combined with a `grandine_0` possible restart (finalized=87 vs network's 93 at 18:03 UTC), effective participation fell below 2/3 supermajority
- **Finality stalled permanently at slot 93** from 17:32:35 UTC; chain continued producing blocks to slot ~1004+ with no finalization

## Run details

- Genesis: ~17:24:44 UTC (2026-04-27)
- zeam_0/1 joined at slot 68 (~17:29:36 UTC)
- Finality progression before stall: 72 → 81 → 87 → 93
- zeam_1: healthy throughout (1–2 slots lag, no resets)
- All other clients (ream, ethlambda, nlean, qlean): healthy

## Likely root cause

The fork choice reset appears to be triggered by the timing desync pathway:

1. Mutex contention (ref #786) causes `onInterval` to misfire and catch up in a burst
2. After the burst, zeam's internal slot clock is ahead of the actual wall clock for a moment
3. Incoming gossip blocks are stamped with real wall-clock time and fall into the "future" from zeam's perspective → `FutureSlot` rejection
4. With no new blocks being accepted, fork choice loses its head and falls back to the finalized checkpoint
5. zeam then re-syncs via RPC range requests and recovers — but the attestation window for those slots has already passed

This is distinct from the `UnknownSourceBlock` bug (#752) and the mutex issue (#786), though it is likely exacerbated by #786.

## Steps to reproduce

Run `zeam_0` in a multi-client devnet (devnet-4 config) for >5 minutes. Observe `CHAIN STATUS` logs for periodic `Head Slot` drops back to the finalized checkpoint value while `Current Slot` continues advancing.

## References

- #786 — mutex serialization (likely contributing)
- #752 — UnknownSourceBlock bug (separate but related devnet failure mode)
- Loki logs: `container=zeam_0`, 2026-04-27 17:29–18:32 UTC, host `46.225.10.32:3100`

zclawz:
## Summary

Fixes #788 — `zeam_0` periodically resetting its fork-choice head to the latest finalized checkpoint while the network advanced 100+ slots beyond, dropping out of attestation duties for tens of slots at a time.

## Root cause (re-derived from logs + code, not just the issue body)

Under mutex contention (#786) the libxev `onInterval` tick is delayed long enough for the **forkchoice's local slot clock** (`fcStore.slot_clock.timeSlots`) to lag wall-time by tens of slots. Gossip blocks for those slots arrive at the wall-clock time and are validated against the lagged clock:

```zig
// pkgs/node/src/chain.zig (pre-fix)
const current_slot = self.forkChoice.fcStore.slot_clock.timeSlots.load(.monotonic);
if (block.slot > current_slot + MAX_FUTURE_SLOT_TOLERANCE /*=1*/) {
    return BlockValidationError.FutureSlot;  // dropped
}
```

With no new descendants entering `protoArray`, `forkchoice.computeFCHead` walks from `latest_justified` and finds no children → `head` falls back to the latest finalized checkpoint. That's the symptom #788 traces (`Current Slot: 200 | Head Slot: 93 | Behind: 107`).

The issue body's direction — "zeam's clock jumps **forward**" — turns out to be inverted; logs from the run show zeam's `current_slot` is wall-clock-correct, the **forkchoice clock** is what lags, and gossip blocks at slot `current_slot + N` (N > 1) are the ones being dropped. Same fix either way: stop dropping queueable future-slot blocks.

## Relationship to #803 (mutex contention work)

The slice work that closed out the mutex-contention root cause (#786) is **already merged**:

| Slice | PR | Status |
|---|---|---|
| (a-2) per-resource locks + `BorrowedState` | #805 | ✅ merged |
| (a-3) drop `BeamNode.mutex` + `BlockCache` | #820 | ✅ merged |
| (b) stress harness + metrics audit | #822 | ✅ merged |
| (c-1) chain-worker scaffold | #826 | ✅ merged |
| (c-2a) `RcBeamState` refcounted state | #827 | ✅ merged |
| (c-2b/c-2c) handler routing + state-rwlock drop + queue-saturation stress | #828 | ✅ merged |

So the question is no longer "are mutexes addressed" — they are. The remaining gap on the contention side is operational rather than code:

- **`--chain-worker` defaults to `off`** in main. PR #830 (`enum + flip default to on`) was closed without merging; the chain-worker queue path is in the binary but not active unless `--chain-worker=on` is passed. Until that flips on (or the CLI shape lands as #830 v2), production deployments are still on the synchronous path that originally exhibited the lag, so the symptom this PR addresses can still appear.
- **Slices (d) and (e) of the original 5-slice plan** (parallel net-fetch + missed-root prune; centralised gossip envelope hash-root cache) don't yet have open PRs.

Even once `--chain-worker=on` is the default, this PR is still useful: a queueable-future-slot pathway is correct independent of why the clock lagged (cold-start lag, GC pause, large STF block under any threading model, slow disk on db commit, …). The defensive layer + the contention work close the loop together.

## Fix

Extend the existing `pending_blocks` queue (which already handled the slot+1 per-interval boundary race) to absorb a longer clock-lag window, plus the supporting bookkeeping to keep it from becoming a memory or correctness foot-gun.

### `validateBlock` — three-way future-slot split

| `block.slot` vs `current_slot` | Outcome |
|---|---|
| `≤ current_slot + MAX_FUTURE_SLOT_TOLERANCE` (=1) | falls through; existing per-interval check either queues or processes inline |
| `≤ current_slot + MAX_FUTURE_SLOT_QUEUE_TOLERANCE` (=256, ~17 min at 4s/slot) | new error `FutureSlotQueueable`; gossip path clones + queues |
| `> MAX_FUTURE_SLOT_QUEUE_TOLERANCE` | `FutureSlot` (hard reject) + bumps `lean_blocks_future_slot_dropped_total` |

`MAX_FUTURE_SLOT_QUEUE_TOLERANCE = 256` was sized to comfortably exceed the worst lag observed on devnet-4 (~160 slots) without giving an attacker a meaningful queue to flood.

### `enqueuePendingBlock` — one place, every edge case

New helper that handles all queue-side concerns under `pending_blocks_lock`:

- **Capacity bound** — `MAX_PENDING_BLOCKS = 1024` (~2MB at typical block size). Cap-evicts **oldest by receive order** (FIFO-fair) so a flood of high-slot fake-future blocks cannot starve a legitimate near-future block out of the queue. The evicted block can always be re-fetched via `blocks_by_range` once we catch up.
- **Dedup by root** — peers re-gossip aggressively during a lag window; duplicates are dropped rather than re-run.
- **Pre-finalized eviction (gossip-path sweep)** — drops queued entries whose slot fell below `latest_finalized.slot`; they can never be canonical.
- **OOM-safe append** — allocator failure path frees the clone and bumps the eviction counter; never silently leaks.

### `processPendingBlocks` — drain-path sweep + replay metrics

The drain loop now also evicts pre-finalized entries during its scan so a queue full of dead entries from a previous lag window self-cleans. Updates `lean_pending_blocks_depth` after every drain so gauges don't go stale.

## Metrics

The slice-(b) lesson — *doc-only audits regress silently* — applies; every new path here gets an instrumented counter or gauge so an operator can spot the bug recurring without grepping logs:

- `lean_pending_blocks_depth` (gauge) — instantaneous queue depth.
- `lean_pending_blocks_evicted_total{reason}` — `cap` / `pre_finalized` / `duplicate` / `append_oom`.
- `lean_pending_blocks_replayed_total{result}` — `accepted` / `rejected` / `error`.
- `lean_blocks_future_slot_dropped_total` — **counts the symptom #788 was tracking**; should stay at 0 under healthy conditions, and a non-zero rate is a smoking gun for a recurrence.

## Edge cases covered (with tests)

All seven new tests live in `pkgs/node/src/chain.zig` and run as part of `zig build test`:

1. **`FutureSlotQueueable` for block within queue tolerance** — basic positive case.
2. **`FutureSlot` hard-reject for block beyond queue tolerance** — basic negative case.
3. **Queue-tolerance boundary is inclusive** — locks the on-by-one boundary in code so a future contributor can't accidentally re-introduce the #788 reset by drift-of-one.
4. **Dedup by root** — re-gossip during a lag window doesn't bloat the queue or duplicate work.
5. **Cap eviction drops oldest by receive order** — fills queue to `MAX_PENDING_BLOCKS`, asserts FIFO eviction at the cap-overflow point and the new entry lands at the tail.
6. **Pre-finalized blocks evicted on gossip-path enqueue** — exercises the in-helper sweep.
7. **Pre-finalized blocks evicted on drain-path scan** — exercises the `processPendingBlocks` sweep, ensuring entries that age out between arrival and drain are still cleaned up.

## Verification

- `zig build` — green.
- `zig build test` — all 14 test groups + 139 chain tests pass; new `chain.test.*#788*` tests confirmed OK locally.
- `cargo build / check -p libp2p-glue` — green (untouched).
- `lean_blocks_future_slot_dropped_total` will let us confirm the fix on devnet-4 by inspection rather than waiting for a head-reset to recur.

## Out of scope (deliberate)

- Flipping `--chain-worker` default to `on` (closed-without-merge #830). Independent operational change; orthogonal to whether the queue-on-future-slot path is correct.
- Slices (d) and (e) of the #803 plan (parallel net-fetch + missed-root prune; centralised gossip envelope hash-root cache). No open PRs; tracked under #803.
- `lean_gossip_mesh_running` boolean gauge (#818 review thread).
- Per-network-id labelled future-block metrics (single-instance node only today; see #818 for the matching atomic-array pattern when we go multi-network).

References: #788 (this), #786 (mutex contention root cause; addressed by merged slices of #803), #803 (slice work).

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -150,8 +150,11 @@ const Metrics = struct {
     //     legitimate near-future gossip blocks).
     //   * `lean_pending_blocks_evicted_total{reason}` — cumulative count of
     //     blocks dropped from the queue, by reason: `cap` (capacity hit),
-    //     `pre_finalized` (slot < finalized), `duplicate` (same root
-    //     already queued), `append_oom` (allocator failure).
+    //     `pre_finalized` (slot < finalized), `too_far_future` (slot >
+    //     current_slot + MAX_FUTURE_SLOT_QUEUE_TOLERANCE; drain-side
+    //     eviction added in PR #841 review #8), `duplicate` (same root
+    //     already queued), `append_oom` (allocator failure on capacity
+    //     reservation; the new block is dropped, the queue is unchanged).
     //   * `lean_pending_blocks_replayed_total{result}` — cumulative count
     //     of replays attempted from `processPendingBlocks`, by
     //     terminal result: `accepted` / `rejected` / `error`.
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -142,6 +142,26 @@ const Metrics = struct {
     // measurement floor before deciding whether to bound the queue or add
     // a cursor optimisation.
     lean_pending_blocks_drain_iters: PendingBlocksDrainItersHistogram,
+    // Issue #788: visibility into the future-block queueing path.
+    //   * `lean_pending_blocks_depth` — instantaneous queue depth, set on
+    //     every successful enqueue and every `processPendingBlocks` drain.
+    //     Combined with `_evicted_total{reason="cap"}` it tells operators
+    //     when the queue cap is being hit (which would silently drop
+    //     legitimate near-future gossip blocks).
+    //   * `lean_pending_blocks_evicted_total{reason}` — cumulative count of
+    //     blocks dropped from the queue, by reason: `cap` (capacity hit),
+    //     `pre_finalized` (slot < finalized), `duplicate` (same root
+    //     already queued), `append_oom` (allocator failure).
+    //   * `lean_pending_blocks_replayed_total{result}` — cumulative count
+    //     of replays attempted from `processPendingBlocks`, by
+    //     terminal result: `accepted` / `rejected` / `error`.
+    //   * `lean_blocks_future_slot_dropped_total` — cumulative count of
+    //     gossip blocks hard-rejected as `FutureSlot` (beyond the
+    //     queueable window), the symptom #788 was tracking.
+    lean_pending_blocks_depth: LeanPendingBlocksDepthGauge,
+    lean_pending_blocks_evicted_total: LeanPendingBlocksEvictedCounter,
+    lean_pending_blocks_replayed_total: LeanPendingBlocksReplayedCounter,
+    lean_blocks_future_slot_dropped_total: LeanBlocksFutureSlotDroppedCounter,
     // Chain-worker queue + loop metrics (slice c-1 of #803).
     //   * `_dropped_total{queue="block"|"attestation"}` — producer
     //     `trySend` rejections when the queue was full.
@@ -254,6 +274,11 @@ const Metrics = struct {
     const LockHoldTimeHistogram = metrics_lib.HistogramVec(f32, LockLabel, &NODE_MUTEX_BUCKETS);
     // pending_blocks drain iteration histogram type (slice a-2)
     const PendingBlocksDrainItersHistogram = metrics_lib.Histogram(f32, &[_]f32{ 1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024 });
+    // Issue #788: future-block queue visibility.
+    const LeanPendingBlocksDepthGauge = metrics_lib.Gauge(u64);
+    const LeanPendingBlocksEvictedCounter = metrics_lib.CounterVec(u64, struct { reason: []const u8 });
+    const LeanPendingBlocksReplayedCounter = metrics_lib.CounterVec(u64, struct { result: []const u8 });
+    const LeanBlocksFutureSlotDroppedCounter = metrics_lib.Counter(u64);
     // Validator status gauge types
     const LeanIsAggregatorGauge = metrics_lib.Gauge(u64);
     const LeanAttestationCommitteeSubnetGauge = metrics_lib.Gauge(u64);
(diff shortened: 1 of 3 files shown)


ch4r10t33r (approved):
LGTM

PR #842, opened 2026-05-07
node: parallel net-fetch (slice d) + centralised hash-root cache (slice e); chain-worker default on (#803)

zclawz:
Three orthogonal but related changes that close the operational gap on #803 and shave duplicate work on every gossip block / RPC fetch.

## Slice (d) of #803 — parallel net-fetch + missed-root prune

`BeamNode.fetchBlockByRoots` previously walked `roots` once, calling `forkChoice.hasBlock(root)` per entry — N shared-lock acquisitions on the forkchoice rwlock and N hashmap lookups, in sequence. Under heavy gossip fanout the per-root acquire dominates the lookup cost.

New `ForkChoice.hasBlocksBatch(roots, out)` snapshots presence for every root under a single shared-lock acquisition. `fetchBlockByRoots` now uses it, then dedups against `network.block_cache` and `network.pending_block_roots` in priority order so each root's outcome — `already_in_forkchoice` / `already_in_block_cache` / `already_pending` / `fetched` — is recorded on `lean_block_fetch_dedup_total{outcome}`. Operators can now see (a) how much duplicate fetch work the dedup is saving and (b) whether a future refactor accidentally bypassed any of the three caches.

`hasBlocksBatch` returns `error.LengthMismatch` if `roots.len != out.len` so callers can't accidentally read past the end of `out`.

## Slice (e) of #803 — centralised hash-root cache

A gossip block currently has its hash-tree root computed up to four times: at `BeamNode.onGossip` (pre-lock), at `chain.onGossip` again, inside `chain.onBlock` if no `blockRoot` was passed, and inside `forkchoice.onBlock` if its `opts.blockRoot` is null. The same applies to RPC and replay paths. `hashTreeRoot(BeamBlock, ...)` is a non-trivial SSZ traversal on a wide block body; doing it 3–4× per ingress is pure waste.

This PR threads the producer's already-computed root all the way through:

- **`networks.GossipMessage` left as-is** (the variant is plain `SignedBlock` and changing the shape would ripple through the FFI bridge); instead `chain.onGossip` gains an optional `precomputed_block_root: ?types.Root` parameter that `BeamNode.onGossip` populates from its own pre-lock computation.
- **`chain_worker.Message.on_block`** and the handler vtable gain an optional `block_root: ?types.Root`. `BeamChain.submitBlock` takes the cached root through; the `chainWorkerOnBlockThunk` forwards it into `onBlock` via `CachedProcessedBlockInfo.blockRoot`.
- **The pending-blocks queue** stores a new `PendingBlockEntry { signed_block, block_root }`; `processPendingBlocks` reads the cached root instead of re-hashing on every ready-block pop.
- **`chain.onBlock` and `forkchoice.onBlock`** already accepted an optional precomputed root; they now bump `lean_block_root_compute_skipped_total{site}` when the caller supplied one so an operator can verify the cache is actually being used at every site. A sustained 0 at any site label means the cache plumbing was dropped on the floor in a refactor.

The default `block_root: ?Root = null` on every new field/parameter keeps existing tests + the c-1 chain-worker stub compiling unchanged.

## Chain-worker default flipped to `on`

`pkgs/cli/src/main.zig` / `pkgs/cli/src/node.zig` / `pkgs/node/src/node.zig` flip `chain_worker_enabled` default from `false` to `true`. The `--chain-worker` CLI flag is unchanged otherwise; the legacy synchronous path stays in place as a kill-switch via `--chain-worker false`.

Per the call: the slice (a/b/c) work is all merged, devnet has been on the worker path via lean-quickstart for a while, and the default flip just brings the compiled-in default in line with what production already runs. (Closed-not-merged #830 attempted the same flip + a CLI shape change; the CLI shape change wasn't actually needed, so this PR keeps the flag a `bool` and just flips the default.)

## Edge cases handled

- **`hasBlocksBatch` length mismatch** returns `error.LengthMismatch` rather than UB-reading `out`. Locked in code by a dedicated test.
- **`hasBlocksBatch` empty input** is a no-op (no panic, no deadlock). Locked in code by a dedicated test.
- **`hasBlocksBatch` snapshot freshness** — the shared lock is re-taken on every call so a post-ingest call reflects the new state. Locked in code by a dedicated test.
- **`fetchBlockByRoots` priority order** — forkchoice → block_cache → pending_block_roots, each producing a distinct `outcome` label. Bypassing any cache would zero the corresponding label, which the metric makes immediately visible.
- **`fetchBlockByRoots` on `error.NoPeersAvailable` / other peer errors** — no `fetched` counter bumps; the existing warn-and-return path is preserved.
- **`chain_worker.Message.on_block.block_root = null` default** — every existing call site (tests, stress harness, c-1 stub) compiles unchanged; the worker recomputes via the existing `CachedProcessedBlockInfo` fallback.
- **`PendingBlockEntry` ownership** — `chain.deinit` walks the queue and calls `entry.signed_block.deinit()`; the `block_root` is value-typed so it doesn't need cleanup. The drain (`processPendingBlocks`) defers `queued_entry.signed_block.deinit()` per ready-pop.
- **`chain.onGossip` non-block branches** ignore `precomputed_block_root` (only the block branch reads it). `BeamNode.onGossip` only computes a root when `data.* == .block`, so we pass `null` for attestation/aggregation paths to keep the contract honest.
- **Chain-worker default flip kill-switch** — `--chain-worker false` (or `chain_worker_enabled: false` in `BeamNodeOpts`) still routes everything through the synchronous path, unchanged from before. Tests that explicitly disable the worker (`chain-worker: end-to-end submitBlock advances state via the worker thread` exercises both branches) still pass.

## New metrics

`pkgs/metrics/src/lib.zig`:

- `lean_block_root_compute_skipped_total{site}` — every time a downstream consumer skipped a `hashTreeRoot(BeamBlock)` because the producer threaded a precomputed root through. Sites: `chain.onGossip`, `chain.onBlock`, `chain.processPendingBlocks`, `forkchoice.onBlock`.
- `lean_block_fetch_dedup_total{outcome}` — every `fetchBlockByRoots` per-root outcome. Outcomes: `already_in_forkchoice`, `already_in_block_cache`, `already_pending`, `fetched`.

Both follow the slice (b) lesson from PR #822 — code-side audits on metrics, not just docs, so a regression that drops the cache or the dedup is observable from Grafana.

## Tests

- **`forkchoice.hasBlocksBatch (slice (d) of #803): empty + length-mismatch + presence semantics`** — empty input is a no-op, length-mismatch returns `error.LengthMismatch` rather than reading past `out`, presence semantics match the per-root `hasBlock`, and the snapshot is re-taken on every call (a post-ingest call reflects the new state).
- **`chain_worker.on_block carries optional block_root (slice (e) of #803)`** — the new field defaults to `null` (existing callers compile unchanged), round-trips an explicit value through `Message.deinit`, and the test's SignedBlock heap is freed by `deinit` (testing.allocator leak guard).

Existing tests around `submitBlock` ("chain-worker: end-to-end submitBlock advances state via the worker thread", "chain-worker disabled: submitBlock returns ChainWorkerDisabled") and the stress harness all updated for the new 3-arg signature; both pass.

## Verification

- `zig build` — green.
- `zig build test` — all 12 test groups pass; both new tests confirmed OK locally; clean exit, no panics / UAFs / deadlocks observed.
- Targeted grep + run on `chain.test.*`, `chain_worker.test.*`, `forkchoice.test.*` confirmed locally.

Draft for review — open the comments thread for any naming / scope feedback before un-drafting.

References: #803 (slice work), prior slices: #805 (a-2), #820 (a-3), #822 (b), #826 (c-1), #827 (c-2a), #828 (c-2b/c-2c). Closed-not-merged #830 (CLI shape attempt) is the prior chain-worker default-flip work; this PR supersedes it.

### DIFF
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -77,9 +77,11 @@ pub const NodeCommand = struct {
     @"db-backend": database.Backend = .rocksdb,
     @"chain-spec": ?[]const u8 = null,
     /// Slice c-2b commit 3 of #803: route producer-side gossip
-    /// handlers through the chain-worker queue. Default `false`
-    /// preserves slice-(b) synchronous behavior.
-    @"chain-worker": bool = false,
+    /// handlers through the chain-worker queue. Default `true` post
+    /// devnet-4 burn-in (#788 follow-up + slice (d)/(e) PR): the
+    /// worker path is the supported prod path; the synchronous path
+    /// stays in place as a kill-switch via `--chain-worker false`.
+    @"chain-worker": bool = true,
 
     pub const __shorts__ = .{
         .help = .h,
@@ -104,7 +106,7 @@ pub const NodeCommand = struct {
         .@"aggregate-subnet-ids" = "Comma-separated list of subnet ids to additionally subscribe and aggregate gossip attestations (e.g. '0,1,2'); adds to automatic computation from validator ids",
         .@"db-backend" = "Database backend to use for on-disk state: 'rocksdb' (default) or 'lmdb'",
         .@"chain-spec" = "Path to the chain specification file, if unspecified falls back to the default setting",
-        .@"chain-worker" = "Route gossip block + attestation handlers through the dedicated chain-worker thread. Off by default; the synchronous path stays in place when disabled.",
+        .@"chain-worker" = "Route gossip block + attestation handlers through the dedicated chain-worker thread. On by default; pass `--chain-worker false` to fall back to the legacy synchronous path as a kill-switch.",
         .help = "Show help information for the node command",
     };
 };
--- a/pkgs/cli/src/node.zig
+++ b/pkgs/cli/src/node.zig
@@ -96,10 +96,12 @@ pub const NodeOptions = struct {
     db_backend: database.Backend = .rocksdb,
     chain_spec: ?[]const u8 = null,
     /// Slice c-2b commit 3 of #803: route producer-side gossip
-    /// handlers through the chain-worker queue. Default `false`
-    /// preserves slice-(b) synchronous behavior. Surfaced as
-    /// `--chain-worker` on the `zeam node` CLI.
-    chain_worker_enabled: bool = false,
+    /// handlers through the chain-worker queue. Default `true` post
+    /// devnet-4 burn-in: the worker path is the supported prod path;
+    /// surfaced as `--chain-worker` on the `zeam node` CLI, with
+    /// `--chain-worker false` as the kill-switch for the legacy
+    /// synchronous path.
+    chain_worker_enabled: bool = true,
 
     pub fn deinit(self: *NodeOptions, allocator: std.mem.Allocator) void {
         for (self.bootnodes) |b| allocator.free(b);
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -174,10 +174,19 @@ const Metrics = struct {
     lean_block_root_compute_skipped_total: LeanBlockRootComputeSkippedCounter,
     // Slice (d) of #803: visibility into the parallel net-fetch +
     // missed-root prune dedup. Bumped per `fetchBlockByRoots` call by
-    // outcome bucket: `already_in_forkchoice`, `already_in_block_cache`,
-    // `already_pending`, `fetched`. Lets operators see how much
-    // duplicate fetch work the dedup is saving and — just as
-    // important — catch a regression that bypasses the dedup.
+    // outcome bucket. Every entry of the input `roots` lands in
+    // exactly one bucket, so the outcome counters sum to `roots.len`
+    // per call — useful for asserting `sum(rate(lean_block_fetch_dedup_total))
+    // == sum(rate(… by outcome))` as a Grafana invariant. Buckets:
+    //   * `already_in_forkchoice` — protoArray hit on `hasBlocksBatch`.
+    //   * `already_in_block_cache` — awaiting parent or STF.
+    //   * `already_pending` — RPC in flight; another
+    //     `fetchBlockByRoots` call is already responsible.
+    //   * `fetched` — dispatched to a peer via `blocks_by_root`.
+    //   * `fetch_no_peers` — dispatch attempted but `selectPeer`
+    //     returned `error.NoPeersAvailable` (PR #842 review #1).
+    //   * `fetch_failed` — dispatch attempted but failed for any
+    //     other reason (queue full, encode failure, …).
     lean_block_fetch_dedup_total: LeanBlockFetchDedupCounter,
 
     const ChainHistogram = metrics_lib.Histogram(f32, &[_]f32{ 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10 });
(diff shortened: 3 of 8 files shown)


ch4r10t33r (commented):
## Adversarial review

Solid PR overall — plumbing is consistent, metric semantics match docs, locking discipline is sound. Findings below, severity-tagged. Nothing blocking.

### Minor

**1. `lean_block_fetch_dedup_total{outcome="fetched"}` undercounts on peer-error path** — `pkgs/node/src/node.zig:1452-1462`
The "fetched" bump only runs inside `if (maybe_request) |request_info|`. If `ensureBlocksByRootRequest` returned `error.NoPeersAvailable` (or any other error), missing roots get logged but never counted. The PR description acknowledges this ("no `fetched` counter bumps on peer errors"), but the inline comment claims "outcome buckets sum to `roots.len` for every call" — they don't on the peer-error path. Either reword the comment, or add an `outcome="rpc_dispatch_failed"` (or `no_peers`) bucket so the buckets do sum and a sustained gap is observable from Grafana.

**2. Forkchoice trusts caller-supplied `block_root` — no validation** — `pkgs/node/src/forkchoice.zig:1671-1694`, `pkgs/node/src/chain.zig` (`onBlock`)
`forkChoice.onBlock` stores `opts.blockRoot` straight into `proto_block.blockRoot` (the protoArray index key) and `chain.onBlock` does not re-validate it against `hashTreeRoot(block)`. All current production callers compute the root themselves, so there is no attacker surface today. But this is a latent footgun: a future call site that propagates a peer-supplied root unchecked would index forkchoice under the wrong key, causing `hasBlock(root) -> false` for an ingested block and an infinite-fetch loop in `fetchBlockByRoots`. Suggest a debug-only `std.debug.assert` inside `chain.onBlock` and `forkchoice.onBlock`: when `blockRoot != null`, recompute and assert equality. Free in release builds, catches the regression in CI.

**3. TOCTOU between `hasBlocksBatch` snapshot and the network-cache checks is benign but undocumented** — `pkgs/node/src/node.zig:1402-1425`
The forkchoice shared lock is dropped after `hasBlocksBatch`, and `network.hasFetchedBlock` / `hasPendingBlockRoot` re-acquire their own mutexes. Between the snapshot and the network checks another thread can ingest into forkchoice; we then dispatch an RPC for it. `ensureBlocksByRootRequest` re-runs `shouldRequestBlocksByRoot` internally, so worst case is one redundant RPC. Fine — but the priority-order comment currently reads as if the dedup is atomic. One-line note "snapshot is point-in-time; concurrent ingestion produces at most one extra RPC, which `ensureBlocksByRootRequest` then dedups again" would prevent a future reader from leaning on stronger guarantees than the code provides.

### Nits

**4. Metric loop is a per-root `incr` over a known length** — `pkgs/node/src/node.zig:1457-1460`
`while (i < missing_roots.items.len) : (i += 1)` does N atomic ops just to bump the same `fetched` counter. If the metric type supports `incrBy(N)` it collapses to one atomic. Hot path under heavy fanout.

**5. `chain.onGossip` skip-counter is observability, not enforcement** — `pkgs/node/src/chain.zig` (block branch)
The site label only bumps when `precomputed_block_root != null`. `BeamNode.onGossip` always threads it in production, so the production counter is non-zero. But if a test/mock path adds a direct `chain.onGossip` call in the block branch and forgets to pass the root, the counter silently stays zero and the regression is invisible. PR description frames the metric as a regression-detector, so this is mostly a doc nit — worth noting in the field comment that the guarantee depends on every producer threading the root.

### Test coverage gaps (not blockers)

- Dedup priority order (forkchoice → block_cache → pending) in `fetchBlockByRoots` is not unit-tested — only the forkchoice batch primitive is covered (`forkchoice.zig` test at the end of slice (d)). A fast unit test that primes each cache and asserts the correct outcome label would lock in the behavior the metric is meant to expose.
- The peer-error path that skips the `fetched` bump is not asserted via metric inspection.
- No test verifies that a wrong precomputed root corrupts forkchoice — relevant only if finding #2 is taken.

### Confirmed safe

- `PendingBlockEntry` lifecycle: appended in `chain.onGossip` (block-not-ready branch), freed in `processPendingBlocks` via `defer queued_entry.signed_block.deinit()` and in `BeamChain.deinit`. No double-free, no aliasing `errdefer`.
- `Handlers.on_block` vtable signature change has exactly one implementor (`chainWorkerOnBlockThunk`), updated in this PR. No external break.
- `chain_worker_enabled` default flip is gated only on `BeamNode` construction; `BeamChain` test paths are unaffected.
- `submitBlock` 3-arg signature: all in-tree call sites updated; the `null` default on the message field keeps wire-shape compatible inside the queue.

Ship verdict: findings #1 and #2 are worth a follow-up commit before un-drafting; the rest are optional polish.

ch4r10t33r (dismissed):
LGTM

ch4r10t33r (approved):
LGTM

PR #850, opened 2026-05-07
chain: log duplicate-AttestationData STF rejects with full context

zclawz:
Splits the chain.zig portion of #848 review #6 into its own PR so the wedge-fix for #837 stays focused on `pkgs/node/src/node.zig` + `pkgs/metrics/src/lib.zig`.

## What this changes

When a block is rejected for containing duplicate `AttestationData` entries, log every distinguishing field of the duplicate so the reproduction has enough on-the-wire context to diagnose without pulling block bytes from a peer.

Two `AttestationData` values are equal iff every field (`slot`, `head`, `target`, `source`) matches, so the new log line includes:

- `blockroot`, `slot`, `proposer_index` of the offending block
- the two indices `[i, j]` inside the block whose data hashed to the same value
- `data.slot`
- `data.head.blockroot @ slot`
- `data.target.checkpoint_root @ slot`
- `data.source.checkpoint_root @ slot`

Pre-fix log:

    block contains duplicate AttestationData entries for block root=0x…

Post-fix log:

    duplicate AttestationData rejected: blockroot=0x… slot=N proposer=P duplicate_indices=[i,j] data.slot=S data.head.blockroot=0x…@H data.target.checkpoint_root=0x…@T data.source.checkpoint_root=0x…@U

## Review nits applied (point #9 from #848)

- Format string split across 5 logical chunks (block-level / data / head / target / source) so it stays diff-friendly and operators can grep individual fields.
- `root=` normalised to `blockroot=` / `checkpoint_root=` to match neighbouring logs in `chain.zig` (e.g. line ~1867 gossip block log).
- `block.proposer_index` is `ValidatorIndex` = `u64`, so `{d}` is correct.

## Refs

- Refs #837 ask #5
- Refs #848 review #6 (split-out)

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -2361,27 +2361,11 @@ pub const BeamChain = struct {
             }
 
             // Each unique AttestationData must appear at most once per block.
-            //
-            // Issue #837 ask #5: log the duplicate's distinguishing
-            // fields (slot + checkpoint roots) so the next reproduction
-            // has enough on-the-wire context to diagnose without
-            // pulling block bytes from a peer. Two `AttestationData`
-            // values are equal iff every field (slot, head, target,
-            // source) matches; logging the slot + checkpoint roots
-            // tells operators which validators voted on the same
-            // checkpoint pair.
-            //
-            // The format string is intentionally split across logical
-            // chunks (block-level / data / head / target / source) so
-            // it stays diff-friendly and operators can grep individual
-            // fields. `blockroot=` / `checkpoint_root=` mirror the
-            // naming used elsewhere in this file (e.g. line ~1867's
-            // gossip log).
             {
-                var att_data_set = std.AutoHashMap(types.AttestationData, usize).init(self.allocator);
-                defer att_data_set.deinit();
+                var att_data_map = std.AutoHashMap(types.AttestationData, usize).init(self.allocator);
+                defer att_data_map.deinit();
                 for (aggregated_attestations, 0..) |agg_att, idx| {
-                    const result = try att_data_set.getOrPut(agg_att.data);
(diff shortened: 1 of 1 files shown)


pkgs/node/src/chain.zig
```diff
@@ -2361,18 +2361,53 @@ pub const BeamChain = struct {
...
+            // source) matches; logging the slot + checkpoint roots
+            // tells operators which validators voted on the same
+            // checkpoint pair.
+            //
+            // The format string is intentionally split across logical
+            // chunks (block-level / data / head / target / source) so
+            // it stays diff-friendly and operators can grep individual
+            // fields. `blockroot=` / `checkpoint_root=` mirror the
+            // naming used elsewhere in this file (e.g. line ~1867's
+            // gossip log).
```
anshalshukla: @zclawz remove this comment, its AI slop

pkgs/node/src/chain.zig
```diff
@@ -2361,18 +2361,53 @@ pub const BeamChain = struct {
...
             {
-                var att_data_set = std.AutoHashMap(types.AttestationData, void).init(self.allocator);
+                var att_data_set = std.AutoHashMap(types.AttestationData, usize).init(self.allocator);
```
anshalshukla: @zclawz it is not a set anymore it is a map

pkgs/node/src/chain.zig
zclawz: Done — removed the verbose explanatory comment. Kept only the one-liner.

pkgs/node/src/chain.zig
zclawz: Fixed — renamed `att_data_set` → `att_data_map` throughout.

anshalshukla (approved):

PR #855, opened 2026-05-09
fix(network): stop integer-overflow panic on malformed gossip SSZ (Hive `gossip: ignores malformed ssz`)

zclawz:
## Hive failure

[`gossip: ignores malformed ssz` (test 390)](https://hive.leanroadmap.org/suite.html?suiteid=1778305924-e14785654316449971f512c113089da3.json&suitename=gossip&client=zeam_devnet4#test-390) on `zeam_devnet4`.

The simulator publishes 1024 bytes of `0xef` on a valid block topic and asserts the client stays healthy. The client log shows a clean panic on the network thread:

```
thread 64 panic: integer overflow
std/mem.zig:3394:42 in handleMsgFromRustBridge (zeam)
...
pkgs/network/src/ethlibp2p.zig:1106:27 in createAndRunNetworkThread (zeam)
```

Reproduced locally with a 30-line repro:

```
$ cat /tmp/test_decode.zig
const std = @import("std");
const snappy = @import("snappy.zig");
pub fn main() !void {
    const garbage = [_]u8{0xef} ** 1024;
    _ = snappy.decodeWithMax(std.heap.page_allocator, &garbage, 16 * 1024 * 1024);
}
$ zig run /tmp/test_decode.zig
thread 628 panic: integer overflow
snappy.zig:73:11 in uvarint     s += 7;
snappy.zig:105:27 in decodedLen const varint = uvarint(src);
snappy.zig:265:33 in decodeWithMax
```

## Root cause

`uvarint()` in `blockblaz/zig-snappy` walks every continuation byte (`b >= 0x80`) and increments a `u6` shift counter `s += 7`. After 9 iterations `s == 63`; on the 10th `s += 7` overflows the u6 in safe builds and panics. Every `0xef` is a continuation byte, so 1024 of them blow past the limit before any terminator is seen.

The network thread crash kills the client. Hive's simulator then can't reach `/lean/v0/fork_choice` on the second node, and 26 follow-up tests cascade-fail with `client should connect and send a request: Elapsed(())` and `"Failed to publish message: Duplicate"`.

### 1. Upstream: [`blockblaz/zig-snappy#9`](https://github.com/blockblaz/zig-snappy/pull/9)

`uvarint()` now returns the corrupt sentinel (`bytesRead == 0`) as soon as it sees the 10th continuation byte, before the next `s += 7`. Also tightened `decodedLen()`'s corrupt check from `<= 0` to `== 0` (the old check was a no-op for usize).

This PR bumps the dep to the tagged release [v0.0.5](https://github.com/blockblaz/zig-snappy/releases/tag/v0.0.5) (`050f529`).

### 2. Local: pre-validate snappy header in `handleMsgFromRustBridge`

Even with the upstream fix, we add a `validateSnappyBlockHeader()` gate in front of `snappyz.decodeWithMax`. It uses zeam's own `uvarint` (`multiformats`) — which already has proper bounds checking — to verify:

- the leading varint decodes cleanly,
- the declared uncompressed size fits the per-topic limit (`MAX_GOSSIP_BLOCK_SIZE` for blocks, `MAX_RPC_MESSAGE_SIZE` otherwise),
- the buffer actually carries a body when the declared size is non-zero.

Malformed payloads are dropped with a `writeFailedBytes` debug dump and an error log. If the upstream decoder ever regresses again, gossip stays alive.

## Edge cases covered (new tests in `pkgs/network/src/ethlibp2p.zig`)

| case | expected | covered |
|---|---|---|
| 1024 bytes of `0xef` (the actual Hive payload) | rejected, no panic | ✅ |
| 11 continuation bytes + valid terminator | rejected | ✅ |
| Exactly 10 continuation bytes, no terminator | rejected | ✅ |
| Empty payload | rejected | ✅ |
| Declared size > `MAX_GOSSIP_BLOCK_SIZE` | rejected | ✅ |
| Header-only buffer with non-zero declared size | rejected | ✅ |
| Well-formed header + payload byte | accepted | ✅ |
| Single `0x00` byte (zero-length block) | accepted | ✅ |
| `snappyz.decodeWithMax` on 1024×0xef | `error.Corrupt`, no panic | ✅ |

Upstream `blockblaz/zig-snappy#9` adds matching uvarint-level edge-case tests (10-byte max u64, 10-byte overflow boundary, etc.).

## Verification

```
$ zig build
Finished `release` profile [optimized] target(s) in 0.36s

$ zig build test
...
11/14 ethlibp2p.test.validateGossipSnappyHeader rejects oversized declared size...OK
12/14 ethlibp2p.test.validateSnappyBlockHeader rejects malformed gossip payloads...OK
13/14 ethlibp2p.test.snappyz.decodeWithMax does not panic on 1024 bytes of 0xef...OK
All 14 tests passed.
...
All 145 tests passed.
All 11 tests passed.
```

Full test suite green.

## Files

- `build.zig.zon` — bump `zig_snappy` to [v0.0.5](https://github.com/blockblaz/zig-snappy/releases/tag/v0.0.5) (`050f529`)
- `pkgs/network/src/ethlibp2p.zig` — add `validateSnappyBlockHeader()`, call it in `handleMsgFromRustBridge` before `snappyz.decodeWithMax`, add 3 regression tests

Refs: blockblaz/zig-snappy#9
cc @ch4r10t33r

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -37,8 +37,8 @@
             .hash = "rocksdb-9.7.4-z_CUTnfJAAB2izSeJNWGGdy-Q0aB3wVMa0AohWJDVVBy",
         },
         .zig_snappy = .{
-            .url = "git+https://github.com/blockblaz/zig-snappy#9521bbcb95b59fe21f729abdba4852b88424af0c",
-            .hash = "zig_snappy-0.0.3-bDFzXuRvAABIBsAlnrHwgPDKH78O3Y5konV6Q8mFpcxy",
+            .url = "git+https://github.com/blockblaz/zig-snappy?ref=v0.0.5#050f529bc5fa11242140312ecf550c186f05af1e",
+            .hash = "zig_snappy-0.0.5-bDFzXlh4AADPRkDi2swDMAf8shH39Fc1g5oYRAhFQfTt",
         },
         .snappyframesz = .{
             .url = "git+https://github.com/blockblaz/snappyframesz#df262c69ca24f072dd71b26c70972649458b6659",
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -37,8 +37,8 @@
             .hash = "rocksdb-9.7.4-z_CUTnfJAAB2izSeJNWGGdy-Q0aB3wVMa0AohWJDVVBy",
         },
         .zig_snappy = .{
-            .url = "git+https://github.com/blockblaz/zig-snappy#5058ae298101e409af9d35cf7eeddc2650477164",
-            .hash = "zig_snappy-0.0.3-bDFzXmBjAAAd1yTK5LQvRJ-srdnJbhkHZajl0KcXVJDx",
+            .url = "git+https://github.com/blockblaz/zig-snappy#9521bbcb95b59fe21f729abdba4852b88424af0c",
+            .hash = "zig_snappy-0.0.3-bDFzXuRvAABIBsAlnrHwgPDKH78O3Y5konV6Q8mFpcxy",
         },
         .snappyframesz = .{
             .url = "git+https://github.com/blockblaz/snappyframesz#df262c69ca24f072dd71b26c70972649458b6659",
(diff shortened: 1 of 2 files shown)


ch4r10t33r (approved):
LGTM

PR #857, opened 2026-05-10
fix: serve anchor block via blocks_by_root; produce blocks with no peers

zclawz:
## Summary

Fixes two failing reqresp hive tests for zeam_devnet4:
- 507 reqresp/blocks_by_root/single_known_block -> left:0 right:1
- 508 reqresp/blocks_by_root/multiple_known_blocks -> timeout (180 s)

## Root Cause 1 - test 507

Checkpoint-sync downloads the BeamState but NOT the SignedBlock. Chain.init derives the anchor block root from anchorState.latest_block_header and registers it as fork-choice head, but DbBlocksNamespace has no entry for that root. When blocks_by_root is requested for the head root, db.loadBlock() returns null -> 0 blocks.

Fix: storeSyntheticAnchorBlock() in Chain.init stores a SignedBlock built from ProtoBlock fields (slot, proposer_index, parent_root, state_root) with empty body + zeroed signature. The hive test only checks slot/parent_root/proposer_index.

## Root Cause 2 - test 508

Test isolates zeam from gossip (connect_client_to_lean_spec_mesh=false). With no peers, getSyncStatus() returned .no_peers and ValidatorClient skipped block production entirely. Fork-choice stuck at genesis.

Fix: Remove the .no_peers early-return from maybeDoProposal() and mayBeDoAttestation(). A validator must propose at its slot regardless of peer count; the block self-imports and gossips once peers connect.

### DIFF
--- a/pkgs/cli/src/api_server.zig
+++ b/pkgs/cli/src/api_server.zig
@@ -140,6 +140,11 @@ fn routeConnection(io: std.Io, connection: net.Stream, allocator: std.mem.Alloca
                 ctx.logger.warn("failed to handle finalized checkpoint state request: {}", .{err});
                 _ = request.respond("Internal Server Error\n", .{ .status = .internal_server_error }) catch {};
             };
+        } else if (std.mem.eql(u8, request.head.target, "/lean/v0/blocks/finalized")) {
+            ctx.handleFinalizedCheckpointBlock(&request) catch |err| {
+                ctx.logger.warn("failed to handle finalized checkpoint block request: {}", .{err});
+                _ = request.respond("Internal Server Error\n", .{ .status = .internal_server_error }) catch {};
+            };
         } else if (std.mem.eql(u8, request.head.target, "/lean/v0/checkpoints/justified")) {
             ctx.handleJustifiedCheckpoint(&request) catch |err| {
                 ctx.logger.warn("failed to handle justified checkpoint request: {}", .{err});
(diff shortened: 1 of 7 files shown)


zclawz:
## Follow-up review (issues 6–11 + nits)

---

### Issue 6 — Aggregator inconsistency: `.no_peers` still gates aggregation (chain.zig:3326)

Confirmed. `maybeAggregateOnInterval` has:

```zig
.no_peers => {
    self.logger.warn("skipping aggregation production for slot={d}: no peers connected", .{slot});
    return null;
},
```

While `maybeDoProposal` and `mayBeDoAttestation` now explicitly proceed through `.no_peers`. The PR's rationale — "a validator has a duty to propose/attest regardless of peer connectivity; local fork-choice benefits from attestations" — applies equally to aggregation. If we're attesting locally with no peers, those attestations should also be aggregated locally so the local fork-choice sees full weight.

The asymmetry is indefensible unless there's a specific reason aggregation differs (e.g. aggregation requires a minimum quorum that's impossible solo). If no such reason exists, either:
- Remove the `.no_peers` guard from `maybeAggregateOnInterval` (consistent with proposer/attester)
- Or add a comment explaining why aggregation is intentionally excluded

Also: this function still logs `.warn` for `.no_peers` — inconsistent with the info-level downgrade in the other two paths (see issue 7).

---

### Issue 7 — Log level inconsistency

`maybeDoProposal` and `mayBeDoAttestation` both log `.info` for `.no_peers`. `maybeAggregateOnInterval` logs `.warn`. With the PR's framing that no-peers operation is now normal/expected, all three should be consistent — `.info` at minimum, `.debug` once stable. The current mix will confuse on-call: `warn` implies "something wrong", `info` implies "expected path".

---

### Issue 8 — `produceBlock` failure propagation under no-peers

`validator_client.zig:141`:
```zig
const produced_block = try self.chain.produceBlock(.{ .slot = slot, .proposer_index = slot_proposer_id });
```

Pre-PR this returned an error that terminated the function. The original `.no_peers` early-return meant this was unreachable. Now it's reachable. The question is whether `produceBlock` can fail with a checkpoint-synced state and no parent state in the post-state cache (common right after restart before any block is processed).

If `produceBlock` traverses the post-state cache to find parent state and the anchor block's post-state isn't there, this panics the validator client loop on slot 1. Worth a targeted test or at minimum auditing that `produceBlock` gracefully handles missing parent state with a descriptive error rather than a hard failure.

---

### Issue 9 — Synthetic block is permanent (by design, undocumented)

After restart, `loadBlock` returns the synthetic block → `storeSyntheticAnchorBlock` is skipped → synthetic stays in DB. If the real anchor block ever arrives via gossip or reqresp, `onBlock` → `updateBlockDb` (chain.zig:2756) would overwrite it — good. But that's an unlikely path for a finalized anchor.

This isn't a bug, but the doc-comment should say so explicitly. Something like:

> "The synthetic block persists until the real block is received (unlikely for finalized anchors). It is intentionally replaceable via `updateBlockDb` but will not be actively fetched. See option (b) for a proper fix."

Without this, future readers may assume it gets cleaned up automatically.

---

### Issue 10 — Parallel anchor request race: OK ✓

`chain.init` is single-threaded; reqresp server starts after init completes. No race between `loadBlock` check and `storeSyntheticAnchorBlock` write. Confirmed fine.

---

### Issue 11 — Hive test 507 root verification: confirmed absent

Cross-checked against hive reqresp.rs:874-876 (#1471): hive decodes `LeanSignedBlock` and reads `block.parent_root` — no `hash_tree_root` verification. Test passes. The PR description accurately notes what hive checks, but doesn't flag that this means hive is insufficient to validate protocol correctness for this path. Should be called out explicitly so reviewers don't treat green hive as sufficient confidence for real-world interop.

---

### Nits

1. **`logger.warn` on DB failure in `storeSyntheticAnchorBlock`**: A corrupt DB silently becomes a healthy-looking node serving no anchor. Consider propagating the error up and letting `Chain.init` decide whether to fatal — warn-and-continue is the wrong posture for init-time DB failures.

2. **Info log "stored synthetic anchor block"**: Should say explicitly `(body may not match real anchor block; hash_tree_root will differ if anchor slot > 0 and body was non-empty)` so on-call has a clue when interop fails in prod.

3. **Doc-comment "checkpoint-sync placeholder"** understates the wire-protocol implication. At minimum: "this block's hash_tree_root will NOT match the stored key for non-genesis checkpoint sync; it is intentionally a placeholder and must not be relied upon for protocol-correct BlocksByRoot responses."

---

### Summary (all issues)

| # | Type | Verdict |
|---|------|---------|
| 1 | CRITICAL — hash_tree_root mismatch | Must fix before merge |
| 2 | Medium — loadBlock leak | Must fix before merge |
| 3 | Low — memory contract fragile | Nice-to-have |
| 6 | Medium — aggregator inconsistency | Fix or document |
| 7 | Low — log level inconsistency | Fix for consistency |
| 8 | Medium — produceBlock failure path | Audit/test needed |
| 9 | Low — synthetic block permanence | Document |
| 10 | OK — no race | ✓ |
| 11 | Hive test insufficient | Improve test coverage |

ch4r10t33r (approved):
LGTM

PR #859, opened 2026-05-11
feat: fork-choice test driver API for hive lean-spec-tests

zclawz:
## Summary

Fixes the failing `lean-spec-tests-fork-choice` hive tests by adding the lean HTTP test-driver surface and hardening the driver against review findings.

Closes #858

## Root cause

Hive lean-spec-tests drives zeam through lean test-driver HTTP endpoints. Before this PR, fork-choice init/step returned 404, so 83/84 tests failed. The lone passing test expected init failure, so 404 accidentally satisfied it.

## HTTP surface added

- `POST /lean/v0/test_driver/fork_choice/init`
- `POST /lean/v0/test_driver/fork_choice/step`
- `GET /lean/v0/test_driver/fork_choice/snapshot`
- `POST /lean/v0/test_driver/state_transition/run`
- `POST /lean/v0/test_driver/verify_signatures/run`

Note: `verify_signatures/run` now returns HTTP 501 / `NotImplemented` until XMSS signature verification is wired, rather than silently accepting signatures.

## Key changes

- Add persistent isolated `ForkChoiceDriverState` reset per init request.
- Parse `anchorState` + `anchorBlock`, validate anchor state root, and initialize `ForkChoice`.
- Dispatch block / tick / attestation / aggregated-attestation / checks steps and return lean-compatible snapshots.
- Add nested checkpoint + safe-target checks in the spectest runner.
- Harden the HTTP driver:
  - clone anchor state instead of bitwise move + dummy reset leak
  - free driver-owned genesis pubkey slices and label-map keys
  - duplicate JSON-backed labels before storing in the persistent map
  - reject negative tick integers without trapping
  - escape JSON error strings
  - return `accepted=false` for unknown step types
  - split malformed/protocol step errors into HTTP 400 instead of spec rejection JSON
  - validate gossip attestation data like the runner
  - document single-driver serialized mutex behavior

## Testing

- `zig build` passed
- `zig build test` passed, including stress harnesses


ch4r10t33r (approved):
LGTM

PR #869, opened 2026-05-13
refactor mock network flow

anshalshukla:
Removes onloop scheduling in case of mock network instead created a `pending_publishes` array that serves as a network layer for the mock network

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,18 @@ jobs:
       with:
         toolchain: nightly
 
+    - name: Prefer rustup shims on PATH (macOS)
+      if: runner.os == 'macOS'
+      run: |
+        echo "$HOME/.cargo/bin" >> "$GITHUB_PATH"
+        command -v cargo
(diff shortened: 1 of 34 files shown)


ch4r10t33r (commented):
Few comments
- Real gossip delivers messages right away. Old mock delivered them ~1 ms later. New mock waits up to 800 ms. For tests with multi-step flows ("node A sends block → node B sees it → produces attestation → aggregator picks it up → puts it in next block"), every step now waits a tick. A 3-step chain that used to finish in one libxev pass now spreads across 3 ticks. Is this expected?
- Drain loop can run forever: When draining, if a handler reacts by publishing a new message, it gets appended to the queue and the loop keeps draining. 
- Can we add `errdefer allocator.destroy(drain_cb)` here
```
const drain_cb = try allocator.create(...);
drain_cb.* = .{ ... };
try clock.subscribeOnSlot(drain_cb);
```
If subscribeOnSlot fails (out of memory), drain_cb is leaked.

ch4r10t33r (approved):
LGTM

PR #870, opened 2026-05-13
Track gossip chain queue depth through processing

zclawz:
## Summary
- make `lean_chain_queue_depth{queue="block|attestation|aggregated_attestation"}` track outstanding chain-worker work until the worker finishes processing it or explicitly discards it during shutdown
- route gossip aggregated attestations through a dedicated `aggregated_attestation` worker queue so backlog/drop metrics stay labelable by work class
- update the worker dispatch loop to give block, raw-attestation, and aggregated-attestation queues one dispatch opportunity per loop, preventing hot raw-attestation traffic from indefinitely starving aggregated attestations
- document the aggregated queue capacity rationale, queue ordering semantics, shutdown discard accounting, and the approximate atomic-snapshot behavior of outstanding-depth gauges
- clarify metric help/comment text so dashboards know about the new `aggregated_attestation` queue label

## Validation
- `zig fmt pkgs/node/src/chain_worker.zig pkgs/node/src/chain.zig pkgs/metrics/src/lib.zig`
- `zig build` (EXIT:0)
- `zig build test --summary all` (EXIT:0)

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -171,8 +171,9 @@ const Metrics = struct {
     // Chain-worker queue + loop metrics (slice c-1 of #803).
     //   * `_dropped_total{queue="block"|"attestation"}` — producer
     //     `trySend` rejections when the queue was full.
-    //   * `_depth{queue="..."}` — instantaneous queue depth, set on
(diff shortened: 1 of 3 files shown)


ch4r10t33r (approved):
LGTM

PR #874, opened 2026-05-13
feat: move aggregate FFI off libxev thread using Io.Threaded worker (closes #873)

### LINKED ISSUE #873: Off aggregate from libxev thread to avoid slot advance blocking
From observation on https://github.com/blockblaz/zeam/issues/863 and bench result by https://github.com/blockblaz/zeam/pull/861, the aggregate is the most cost on the libxev thread. 

please try to use the attachment way to implement it and make a PR @zclawz 

[2026-05-13-aggregate-off-libxev-design.md](https://github.com/user-attachments/files/27708221/2026-05-13-aggregate-off-libxev-design.md)

zclawz:
Closes #873

## Summary

Move the XMSS recursive-aggregation FFI call off the libxev main thread to a dedicated `std.Io.Threaded` worker. The `xmss_aggregate` FFI takes 1–5 s; running it synchronously on the libxev event loop blocked slot ticks and network completions, causing the spikes observed in issue #867.

### `pkgs/node/src/chain.zig`
- Add `aggregate_io: *std.Io.Threaded` and `aggregate_group: std.Io.Group` fields (dedicated worker, `concurrent_limit = .limited(1)`)
- Init `aggregate_io` in `BeamChain.init` with `async_limit = .nothing, concurrent_limit = .limited(1)`
- Deinit order in `BeamChain.deinit`: `aggregate_group.cancel` → `chain_worker.stop` → `aggregate_io.deinit`
- Replace `maybeAggregateOnInterval` + caller-side publish with `submitAggregateOnInterval` (returns in microseconds)
- New `aggregateImpl` worker: runs on the dedicated thread, owns the snapshot, calls `forkChoice.aggregate`, then `node.publishProducedAggregations`
- `ConcurrencyUnavailable` (previous task in-flight): skip with metric, free snapshot immediately

### `pkgs/node/src/node.zig`
- At `interval_in_slot == 2`: replace old pattern with `self.chain.submitAggregateOnInterval(self, interval)`
- Promote `publishProducedAggregations` to `pub`

- New `zeam_aggregate_skip_total{reason}` counter (reasons: `in_flight`, `not_aggregator`, `not_synced`, `missing_state`)
- New `zeam_aggregate_worker_duration_seconds` histogram

## Testing
- `zig build` ✅
- `zig build test` ✅

### DIFF
--- a/pkgs/cli/src/node.zig
+++ b/pkgs/cli/src/node.zig
@@ -398,7 +398,8 @@ pub const Node = struct {
         const cpu_count = std.Thread.getCpuCount() catch 2;
         const reserved_system_threads: usize = 4; // main, p2p, api server, metrics server
         const desired_workers = @max(@as(usize, 1), cpu_count -| reserved_system_threads);
-        const worker_count = @min(desired_workers, @as(usize, ThreadPool.max_thread_count));
+        const zig_worker_budget = @max(@as(usize, 1), (desired_workers + 1) / 2);
+        const worker_count = @min(zig_worker_budget, @as(usize, ThreadPool.max_thread_count));
         self.thread_pool = try ThreadPool.init(.{
             .allocator = allocator,
             .io = std.Io.Threaded.global_single_threaded.io(),
@@ -406,13 +407,16 @@ pub const Node = struct {
         });
         errdefer self.thread_pool.deinit();
 
-        // Limit the rayon thread pool used by the XMSS aggregate prover so that
-        // it doesn’t consume all CPU cores. Reserve 3 threads for libxev,
-        // the chain worker, and the rust-libp2p network thread (issue #873).
+        // Coordinate the Zig worker pool and the rayon pool used by the XMSS
+        // aggregate prover from the same post-system-thread budget so they do
+        // not independently claim every remaining CPU. Prefer the Zig pool for
+        // the extra worker on odd counts since aggregate verification enters
+        // rayon from Zig workers. Both pools still keep a minimum of one worker
+        // so tiny/cgroup-limited systems remain functional.
         // Must be called before setupProver/setupVerifier since rayon’s global
         // pool is initialized lazily on first use.
-        const rayon_threads = cpu_count -| 3; // saturating sub — 0 means rayon default
-        xmss.setRayonThreads(if (rayon_threads > 0) rayon_threads else 1);
+        const rayon_threads = @max(@as(usize, 1), desired_workers -| worker_count);
+        xmss.setRayonThreads(rayon_threads);
 
         // Pre-warm the XMSS verifier on the main thread before any worker can
         // call `verifyAggregatedPayload`. The Rust-side verifier setup is
--- a/pkgs/cli/src/node.zig
+++ b/pkgs/cli/src/node.zig
@@ -406,6 +406,14 @@ pub const Node = struct {
         });
         errdefer self.thread_pool.deinit();
 
+        // Limit the rayon thread pool used by the XMSS aggregate prover so that
+        // it doesn’t consume all CPU cores. Reserve 3 threads for libxev,
+        // the chain worker, and the rust-libp2p network thread (issue #873).
+        // Must be called before setupProver/setupVerifier since rayon’s global
+        // pool is initialized lazily on first use.
+        const rayon_threads = cpu_count -| 3; // saturating sub — 0 means rayon default
+        xmss.setRayonThreads(if (rayon_threads > 0) rayon_threads else 1);
+
         // Pre-warm the XMSS verifier on the main thread before any worker can
         // call `verifyAggregatedPayload`. The Rust-side verifier setup is
         // documented as idempotent but is not hardened against first-time-init
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -128,6 +128,11 @@ const Metrics = struct {
     /// Wall time for the per-slot aggregation tick (interval 2): `maybeAggregateOnInterval`
     /// plus `publishProducedAggregations` when the aggregator produces gossip aggregates.
     zeam_node_aggregation_interval_tick_seconds: AggregationIntervalTickHistogram,
+    /// Counter for skipped aggregate submissions, labeled by reason.
+    /// Reasons: "in_flight", "not_aggregator", "not_synced", "missing_state".
+    zeam_aggregate_skip_total: AggregateSkipCounter,
+    /// Histogram for the wall-clock duration of the aggregate FFI worker.
+    zeam_aggregate_worker_duration_seconds: AggregateWorkerDurationHistogram,
     // BeamNode mutex contention metrics (issue #786)
     // Wait time = how long a callsite blocked before acquiring BeamNode.mutex.
     // Hold time = how long the callsite kept the mutex locked.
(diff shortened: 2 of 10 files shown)


GrapeBaBa:
@zclawz I think there is another problem, since aggregate FFI underlying using ray thread pool seems use all the cpu cores, is there an approach to set the thread number to cpu count -3, because we need 3 three threads for libxev, chainworker and rust libp2p network

pkgs/cli/src/node.zig
```diff
@@ -406,6 +406,14 @@ pub const Node = struct {
         });
         errdefer self.thread_pool.deinit();
 
+        // Limit the rayon thread pool used by the XMSS aggregate prover so that
+        // it doesn’t consume all CPU cores. Reserve 3 threads for libxev,
+        // the chain worker, and the rust-libp2p network thread (issue #873).
+        // Must be called before setupProver/setupVerifier since rayon’s global
+        // pool is initialized lazily on first use.
+        const rayon_threads = cpu_count -| 3; // saturating sub — 0 means rayon default
```
GrapeBaBa: @zclawz you can use `desired_workers`, but I also found another problem, rayon_threads and zig thread pool both use all remain threads which still conflict. can we shift the XMSS parallel verify into rust side which also use the rayon thread pool?

zclawz: Good catch @GrapeBaBa — fixed in 8580daba1ad81e378a73f930ee47b1a74eda72af. Now both pools share the `desired_workers` budget instead of independently claiming the remaining CPUs.

zclawz: Addressed this too in `3fa289f`: aggregate payload verification is now batched through Rust FFI and parallelized on the configured rayon pool, so it no longer fans out through the Zig ThreadPool. Zig still does the serial pre-pass for structural checks/pubkey-cache warmup, then Rust/rayon owns the CPU-heavy XMSS verification. `zig build` passes.

GrapeBaBa: @zclawz after this change, what other tasks will run on the zig thread pool? if we can remove it or reduce its thread num and assign more thread to rayon thread pool?

zclawz: After this PR, the Zig `ThreadPool` still has two remaining consumers:

1. **Block signature verification** (`verifySignaturesParallel` in `chain.zig`) — parallel XMSS pubkey-cache warmup + signature checks on gossip blocks. This is Zig-native code (pubkey cache lookups + hash-sig FFI per-validator), so it can't trivially move to rayon without new FFI surface.

2. **Attestation compaction** (`compactAttestations` in `forkchoice/types`) — parallel per-`AttestationData` grouping during block production. Also Zig-native.

Given that aggregate verification is now on rayon, the Zig pool workload is lighter. We could reduce `zig_worker_budget` (e.g. cap at 2) and hand the savings to rayon — that seems reasonable for devnet validator counts. Removing the pool entirely would need Rust wrappers for those two callsites first.

Want me to open a follow-up issue to track the pool-sizing question?

GrapeBaBa: @zclawz is it good if we shift all this parallel to rust side? or Is there a way we can pin thread on specific core which make zig and rust side threads not compete cores

zclawz: Good question — both are worth thinking about but have different cost/benefit profiles right now.

## 1. Shifting the remaining Zig parallel work to Rust/rayon

The two remaining `ThreadPool` consumers are:

- **`verifySignaturesParallel`** — pubkey cache lookups (Zig `HashMap`) + per-validator XMSS hash-sig verification (already FFI into Rust)
- **`compactAttestations`** — grouping attestations by `AttestationData` during block production (pure Zig, no FFI)

Moving these to Rust would mean:

- **New FFI surface**: You'd need to pass entire blocks (or at least validator indices + signatures + the pubkey cache) across the boundary for sig verification. The pubkey cache is a Zig-owned `HashMap` — either you duplicate it on the Rust side (memory cost + sync overhead) or you pass callback pointers back into Zig (complex, fragile).
- **Attestation compaction** is pure Zig data structure work with no Rust involvement today. Moving it to Rust means serializing `AttestationData` structs across FFI just to do a group-by — that's more overhead than the parallelism saves.
- **Ownership gets messy**: rayon tasks would need to borrow Zig-owned data, which means either copying everything or unsafe pointer sharing with lifetime guarantees you can't express across FFI.

**Verdict**: Not worth it now. The FFI marshalling cost would likely eat the gains, especially for compaction. The current split (Zig pool for Zig-native work, rayon for Rust-native work) is the natural boundary. If sig verification becomes a bottleneck, we could look at moving *just* the hash-sig batch to a rayon `par_iter` since that part is already FFI — but that's a targeted optimization, not a wholesale move.

## 2. CPU core pinning

Feasible in principle — the mechanisms are:

- **Linux**: `pthread_setaffinity_np` / `sched_setaffinity` per-thread, or `cpuset` cgroups for process-level isolation
- **Zig side**: `std.Thread` exposes the underlying `pthread_t`, so you can call `pthread_setaffinity_np` via `@cImport`
- **Rayon side**: rayon doesn't support affinity natively, but you can use `rayon::ThreadPoolBuilder::spawn_handler` to set affinity on each spawned thread

The complexity though:

- You need to decide the split at startup (e.g., cores 0–3 for Zig, 4–7 for rayon), which means knowing the topology and workload characteristics in advance
- It's brittle across different machines (CI vs production vs dev laptops with different core counts)
- NUMA-awareness adds another layer if you want it to actually help on multi-socket systems
- The current budget-sharing approach (`desired_workers / 2` split) is simpler and adapts to any core count automatically

**Verdict**: Core pinning is a meaningful optimization for production validators on known hardware, but it's premature right now. The budget-sharing approach in this PR eliminates the *contention* problem (both pools fighting for the same cores) without the rigidity of pinning. Pinning would be a good follow-up once we have profiling data showing cache-line contention or NUMA effects.

## Recommendation

**For now**: merge the budget-sharing approach as-is — it's simple, portable, and solves the immediate problem.

**Follow-up issues worth filing**:
1. **Profile the Zig↔Rust thread interaction** under load to see if there's actual cache-line bouncing that pinning would fix
2. **Consider batching sig verification FFI calls** into rayon (since the hash-sig part is already Rust) — this is a lighter-weight version of "shift to Rust" that doesn't require moving Zig data structures

Neither is urgent for v0.4.x though.

zclawz: Good question! The actual rayon configuration *is* already on the Rust side — `xmss_set_rayon_threads` is a Rust FFI function that calls `rayon::ThreadPoolBuilder::new().num_threads(n).build_global()`. What stays in Zig is just the thread count *calculation*, and I think that's the right split:

**Why the budget calculation should stay in Zig:**
The Zig side is the system orchestrator — it knows the total CPU count, how many threads it's handing to the Zig worker pool, how many are reserved for libxev/p2p/etc. Rust can't know the Zig pool size without being told. If we moved the calculation to Rust, we'd need to pass all that context as parameters anyway, which is equivalent to what we already do (pass the computed `rayon_threads` value).

**On thread pinning / CPU affinity:**
Thread pinning is an interesting idea — assigning cores 0–N to Zig workers and N+1–M to rayon would eliminate all contention. In practice though:
- It's Linux-only (`sched_setaffinity`); macOS has `thread_policy_set` but it's advisory not binding
- On containerised/cgroup-limited deployments, we may not have the cores we expect
- The budget-sharing approach in this PR (splitting the remaining budget between the two pools) achieves the same goal more portably

The budget split (Zig workers + rayon each get a share of what's left after reserving system threads) should be sufficient. Worth revisiting thread pinning if we see scheduler contention in profiling, but I'd rather avoid platform-specific complexity for now.

Happy to add a Rust-side helper that does both the CPU query and the calculation if you feel strongly the logic should live in Rust — just say the word!

GrapeBaBa (approved):

PR #875, opened 2026-05-13
fix(types): clone participants bitlist when building aggregation bits

shariqnaiyer:
When running a devnet with ream zeam ethlambda in 2 subnets we get the following error and ethlambda rejects zeams blocks at certain slots which lead to no finality in ethlambda during these devnets. 

```
  ethlambda_s0_p0.log:
  WARN ethlambda_blockchain: Failed to process block slot=7  proposer=1 block_root=6526aa5a parent_root=fa625e7b err=Aggregated proof
  participants don't match attestation aggregation bits
  WARN ethlambda_blockchain: Failed to process block slot=16 proposer=4 block_root=a5631b2f parent_root=5884dc6c err=Aggregated proof
  participants don't match attestation aggregation bits
```

Some AI context:
```
  When zeam builds an aggregated attestation, it has to produce two parallel pieces of data:

  - attestation.aggregation_bits — lives in the block body, marks which validators are covered.
  - proof.participants — lives in the signature group on the signed block, marks the same set.

  Both are SSZ Bitlists. They are supposed to represent the same set of validators, and ethlambda (and the spec) require them to be
  byte-for-byte equal when verifying a block:

  // ethlambda/crates/blockchain/src/store.rs:1187
  if attestation.aggregation_bits != aggregated_proof.participants {
      return Err(StoreError::ParticipantsMismatch);
  }

  zeam built aggregation_bits from proof.participants like this (in block.zig:compactSingleProof and compactMultiProofWithPrep):

  var att_bits = try attestation.AggregationBits.init(allocator);  // length 0
  for (0..cloned_proof.participants.len()) |i| {
      if (cloned_proof.participants.get(i) catch false) {
          try attestation.aggregationBitsSet(&att_bits, i, true);  // grows to i+1
      }
  }

  aggregationBitsSet only extends the bitlist up to index + 1. So the resulting att_bits.len() equals (highest set index) + 1, while
  proof.participants.len() is whatever the FFI aggregate routine returned — which can include trailing zero bits.

  Why the lengths can differ

  proof.participants is the merged bitfield from a recursive aggregation. The FFI aggregate() call merges multiple child proofs and
  gossip signatures, and the merged bitlist's length is determined by the highest index across all inputs — including children that
  contributed to coverage but whose own highest bit was higher than the highest bit in this proof.

  Concretely: if a parent proof merges children where one child covers validator 5 (length 6) and the final selected subset only ends
  up with validators {0, 2} set, participants.len() can stay at 6 while the bit at index 5 is 0. Rebuilding att_bits by re-setting TRUE
   bits produces a bitlist of length 3 (highest set is 2 → length 2+1).

  In SSZ encoding, those two bitlists are different:
  - length 3, bits 101 → encoded with the length delimiter at bit position 3
  - length 6, bits 101000 → encoded with the length delimiter at bit position 6
  
  Same set of set-bits, different bytes on the wire, different tree_hash_root, different BitList == BitList result.

  Why it was invisible inside zeam

  zeam's own block-verification path (chain.zig:2396) compares cardinality only:

  if (validator_indices.items.len != participant_indices.items.len) {
      ...
  }

  So zeam happily produced mismatched-length pairs and happily re-imported them. ream was also lenient enough not to fire. Only
  ethlambda's strict equality check caught the bug — but ethlambda then rejected every zeam block carrying an aggregated attestation,
  kept retrying it from the orphan queue, and cross-subnet attestations never reached ethlambda's chain. Justification stayed pinned to
   slot 0.

  Why slot 1/4 worked but slot 7/16 failed

  The path that diverges only fires when there's >0 attestations and the aggregation step actually runs. zeam blocks at slot 1 had
  attestation_count=0 (nothing to mismatch) and slot 4 had attestation_count=1 from a single-validator path that happened to produce a
  length-matched bitlist. From slot 7 onward, real aggregation kicked in and the lengths diverged.
  ```

I have tested this in the devnets and it fixes the issue. However feel free to request changes or close this PR.

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -2448,7 +2448,7 @@ pub const BeamChain = struct {
                 var participant_indices: std.ArrayList(usize) = try types.aggregationBitsToValidatorIndices(&signature_proof.participants, self.allocator);
                 defer participant_indices.deinit(self.allocator);
 
-                if (validator_indices.items.len != participant_indices.items.len) {
+                if (validator_indices.items.len != participant_indices.items.len or !std.mem.eql(usize, validator_indices.items, participant_indices.items)) {
                     zeam_metrics.metrics.lean_attestations_invalid_total.incr(.{ .source = "block" }) catch {};
                     self.logger.err(
                         "attestation signature mismatch index={d} validators={d} participants={d}",
--- a/pkgs/node/src/forkchoice.zig
+++ b/pkgs/node/src/forkchoice.zig
@@ -1071,12 +1071,12 @@ pub const ForkChoice = struct {
                     try types.sszClone(self.allocator, types.AggregatedSignatureProof, best_proof.?.*, &cloned_proof);
                     errdefer cloned_proof.deinit();
 
-                    var att_bits = try types.AggregationBits.init(self.allocator);
+                    var att_bits: types.AggregationBits = undefined;
+                    try types.sszClone(self.allocator, types.AggregationBits, cloned_proof.participants, &att_bits);
                     errdefer att_bits.deinit();
 
                     for (0..cloned_proof.participants.len()) |i| {
                         if (cloned_proof.participants.get(i) catch false) {
-                            try types.aggregationBitsSet(&att_bits, i, true);
                             if (i >= covered.capacity()) {
                                 try covered.resize(i + 1, false);
                             }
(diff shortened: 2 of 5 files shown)


anshalshukla (dismissed):

ch4r10t33r (approved):
LGTM

PR #876, opened 2026-05-13
log attestation subnet coverage in chain status

zclawz:
## Summary
- track aggregated attestation payload coverage from the latest block slot separately from fork-choice's known/new payload pools
- add a chain-status info log for the previous slot showing per-subnet coverage for new payloads, block payloads, their combined union, and full-network coverage
- de-duplicate validators across payloads before calculating percentages

Example shape:

```text
attestation aggregate coverage: new=subnet0=12/16(75.00%) | block=subnet0=10/16(62.50%) | combined=subnet0=14/16(87.50%) | network=14/64(21.88%)
```

## Validation
- `zig fmt pkgs/node/src/forkchoice.zig pkgs/node/src/chain.zig`
- `zig build test -Dprover=dummy --summary all` is running locally in `/tmp/subnet-coverage-test.log`; will update once complete.

### DIFF
--- a/pkgs/node/src/forkchoice.zig
+++ b/pkgs/node/src/forkchoice.zig
@@ -3197,22 +3197,6 @@ fn collectCoverageFromPayloads(
     }
 }
 
-fn collectCoverageFromPayloadsForData(
-    map: *AggregatedPayloadsMap,
-    att_data: types.AttestationData,
-    committee_count: types.SubnetId,
-    seen: []bool,
-    has_subnet: []bool,
-) void {
-    const payloads = map.get(att_data) orelse return;
-    for (payloads.items) |*stored| {
-        if (stored.slot != att_data.slot) continue;
-        for (0..stored.proof.participants.len()) |validator_index| {
-            markCoverageIfParticipant(stored.proof.participants, validator_index, committee_count, seen, has_subnet);
-        }
-    }
-}
-
 fn collectSourceCoverageFromPayloadsForData(
     map: *AggregatedPayloadsMap,
     att_data: types.AttestationData,
@@ -3226,18 +3210,18 @@ fn collectSourceCoverageFromPayloadsForData(
 
         if (stored.source_payload_participants) |source_payload| {
             for (0..source_payload.len()) |validator_index| {
-                markCoverageIfParticipantNoSubnet(source_payload, validator_index, committee_count, payload_seen);
+                markCoverageValidatorOnly(source_payload, validator_index, committee_count, payload_seen);
             }
         } else {
             // Unknown/legacy source: treat the child payload itself as payload-originated.
             for (0..stored.proof.participants.len()) |validator_index| {
-                markCoverageIfParticipantNoSubnet(stored.proof.participants, validator_index, committee_count, payload_seen);
+                markCoverageValidatorOnly(stored.proof.participants, validator_index, committee_count, payload_seen);
             }
         }
 
         if (stored.source_gossip_participants) |source_gossip| {
             for (0..source_gossip.len()) |validator_index| {
-                markCoverageIfParticipantNoSubnet(source_gossip, validator_index, committee_count, gossip_seen);
+                markCoverageValidatorOnly(source_gossip, validator_index, committee_count, gossip_seen);
             }
         }
     }
@@ -3261,7 +3245,10 @@ fn markCoverageIfParticipant(
     has_subnet[subnet_index] = true;
 }
 
-fn markCoverageIfParticipantNoSubnet(
+/// Mark validator-level coverage without tracking per-subnet bits.
+/// Validates subnet membership to keep the same inclusion criteria as
+/// markCoverageIfParticipant but does not set has_subnet.
+fn markCoverageValidatorOnly(
     participants: types.AggregationBits,
     validator_index: usize,
     committee_count: types.SubnetId,
@@ -3353,14 +3340,6 @@ fn countSeen(seen: []const bool) usize {
     return count;
 }
 
-fn countTrue(values: []const bool) usize {
-    var count: usize = 0;
-    for (values) |value| {
-        if (value) count += 1;
-    }
-    return count;
-}
-
 fn recordAggregateCoverageMetrics(section: []const u8, seen: []const bool, has_subnet: []const bool) void {
     zeam_metrics.metrics.zeam_attestation_aggregate_coverage_validators.set(
         .{ .section = section },
@@ -3368,7 +3347,7 @@ fn recordAggregateCoverageMetrics(section: []const u8, seen: []const bool, has_s
     ) catch {};
     zeam_metrics.metrics.zeam_attestation_aggregate_coverage_subnets.set(
         .{ .section = section },
-        @intCast(countTrue(has_subnet)),
+        @intCast(countSeen(has_subnet)),
     ) catch {};
 }
 
--- a/pkgs/node/src/forkchoice.zig
+++ b/pkgs/node/src/forkchoice.zig
@@ -1424,7 +1424,8 @@ pub const ForkChoice = struct {
     ) !void {
         var cloned_proof: types.AggregatedSignatureProof = undefined;
         try types.sszClone(self.allocator, types.AggregatedSignatureProof, proof, &cloned_proof);
-        errdefer cloned_proof.deinit();
+        var cloned_proof_owned = true;
+        errdefer if (cloned_proof_owned) cloned_proof.deinit();
 
         {
             self.signatures_mutex.lock();
@@ -1444,6 +1445,7 @@ pub const ForkChoice = struct {
                 .slot = attestation_data.slot,
                 .proof = cloned_proof,
             });
+            cloned_proof_owned = false;
 
             if (is_from_block) {
                 if (self.latest_block_aggregated_payloads_slot == null or self.latest_block_aggregated_payloads_slot.? != attestation_data.slot) {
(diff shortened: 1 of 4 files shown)


zclawz:
Review note on #876:

I found one blocker in `storeAggregatedPayload` after the new block-coverage bookkeeping was added.

After the first `gop.value_ptr.append(... cloned_proof ...)` succeeds, the function now does more fallible work for `is_from_block` (`sszClone`, `getOrPut`, second append). The original `errdefer cloned_proof.deinit()` is still active, so if any of those later operations fail, the function returns an error and deinitializes `cloned_proof` even though ownership has already been transferred into `latest_known_aggregated_payloads`. That leaves the known-payload map holding freed proof internals, which can become a UAF/double-free later during aggregation/pruning/deinit.

Suggested fix: make the ownership transfer explicit before any subsequent `try` can run. For example, append through a local `stored_proof = cloned_proof`, clear/disarm the errdefer after the successful append, or reorder/preallocate all fallible block-map work before inserting into the known/new map.

I also ran `zig fmt --check pkgs/node/src/forkchoice.zig pkgs/node/src/chain.zig` on the PR head; formatting is clean. I did not run the full build/test suite inline.

g11tech:
@zclawz also generate appropriate metrics for this based on the current process in which metrics are create and used in the repo.

pkgs/node/src/forkchoice.zig
```diff
@@ -2712,15 +3149,251 @@ fn stageAggregatedAttestation(
...
+    return count;
+}
+
+fn countTrue(values: []const bool) usize {
+    var count: usize = 0;
+    for (values) |value| {
+        if (value) count += 1;
+    }
+    return count;
+}
```
anshalshukla: identical functions

pkgs/node/src/forkchoice.zig
```diff
@@ -1148,11 +1205,121 @@ pub const ForkChoice = struct {
             break;
         }
 
+        self.logBlockProposalPayloadCoverage(slot, &agg_attestations);
```
anshalshukla, line 1326: We have a mutex locked here, this is a heavy function and should be definitely avoided with a lock in place. We do multiple count_seen which will be huge once the validator count scales up

pkgs/node/src/forkchoice.zig
```diff
@@ -2712,15 +3149,251 @@ fn stageAggregatedAttestation(
...
     var it = map.iterator();
     while (it.next()) |entry| {
+        deinitAggregatedPayloadsList(allocator, entry.value_ptr);
+    }
+    map.deinit();
+}
+
+fn collectCoverageFromPayloads(
+    map: *AggregatedPayloadsMap,
+    slot: types.Slot,
```
anshalshukla, line 3623: Too many changes in forkchoice, can we separate it out. Let's have minimal code in this file and avoid coverage/ metrics things as much as possible. It's easier to read imo

pkgs/node/src/forkchoice.zig
```diff
@@ -2712,15 +3149,251 @@ fn stageAggregatedAttestation(
...
+    payload_seen: []bool,
+    gossip_seen: []bool,
+) void {
+    const payloads = map.get(att_data) orelse return;
+    for (payloads.items) |*stored| {
+        if (stored.slot != att_data.slot) continue;
+
+        if (stored.source_payload_participants) |source_payload| {
+            for (0..source_payload.len()) |validator_index| {
+                markCoverageIfParticipantNoSubnet(source_payload, validator_index, committee_count, payload_seen);
```
anshalshukla: why not use `markCoverageIfParticipant`, also `noSubnet` suffix looks very misleading to me

pkgs/node/src/forkchoice.zig
```diff
@@ -2712,15 +3149,251 @@ fn stageAggregatedAttestation(
...
+    const payloads = map.get(att_data) orelse return;
+    for (payloads.items) |*stored| {
+        if (stored.slot != att_data.slot) continue;
+        for (0..stored.proof.participants.len()) |validator_index| {
+            markCoverageIfParticipant(stored.proof.participants, validator_index, committee_count, seen, has_subnet);
+        }
+    }
+}
+
+fn collectSourceCoverageFromPayloadsForData(
```
anshalshukla, line 3652: why not modify collectCoverageFromPayloadsForData itself, we already have a fallback in place. If not, we should still remove `collectCoverageFromPayloadsForData` if we are not calling it anywhere

pkgs/node/src/forkchoice.zig
```diff
@@ -318,6 +318,13 @@ pub const ForkChoice = struct {
     // Aggregated signature proofs that are known and contribute to fork choice weights.
     // Used for recursive signature aggregation when building blocks.
     latest_known_aggregated_payloads: AggregatedPayloadsMap,
+    // Aggregated signature proofs observed in the most recently processed block slot.
+    // This is observability-only; fork choice uses latest_known_aggregated_payloads.
+    latest_block_aggregated_payloads: AggregatedPayloadsMap,
+    latest_block_aggregated_payloads_slot: ?types.Slot,
+    /// Coverage of latest_new_aggregated_payloads captured just before the last
+    /// acceptNewAttestationsUnlocked merge.  Used to report "prev_new" in printSlot.
+    saved_pre_merge_new_coverage: ?SavedPreMergeCoverage = null,
```
anshalshukla, line 327: let's abstract this out completely in a separate Coverage struct maybe having everything here is confusing and prone to errors

pkgs/node/src/forkchoice.zig
```diff
@@ -898,6 +909,52 @@ pub const ForkChoice = struct {
             defer self.signatures_mutex.unlock();
 
             if (self.latest_new_aggregated_payloads.count() > 0) {
+                // Capture coverage of new payloads BEFORE merging them into known.
+                // Saved as `saved_pre_merge_new_coverage` and used by
+                // formatAggregationCoverageReport as the "prev_new" section.
```
anshalshukla, line 914: This is a locked path we should avoid so much coverage related load here

anshalshukla (changes requested):

pkgs/node/src/forkchoice.zig
zclawz: Fixed in [997d52b](https://github.com/blockblaz/zeam/commit/997d52b): removed `countTrue` entirely and updated `recordAggregateCoverageMetrics` to use `countSeen` (same logic, `bool` slice = `countSeen` works for both).

pkgs/node/src/forkchoice.zig
zclawz: Acknowledged — `getProposalAttestationsUnlocked` is called while the caller holds the FC lock, so `logBlockProposalPayloadCoverage` (which allocates + iterates the full validator set) runs under that lock. The cleanest fix is to return the coverage data as part of `ProposalAttestationsResult` and let the caller log after releasing the lock. That is a slightly larger refactor to do safely. Happy to do it as a follow-up commit or as a separate PR if you prefer to keep this focused.

pkgs/node/src/forkchoice.zig
zclawz: Agreed. Extracting `collectCoverageFromPayloads*`, `appendCoverageSection`, `logBlockProposalPayloadCoverage` etc. into a `pkgs/node/src/coverage.zig` with a `Coverage` struct is the right direction. I will do this in a follow-up commit to keep the diff reviewable — the current PR already has 714 lines changed in forkchoice.zig and adding another large move in the same patch would make it harder to review.

pkgs/node/src/forkchoice.zig
zclawz: Fixed in [997d52b](https://github.com/blockblaz/zeam/commit/997d52b): renamed to `markCoverageValidatorOnly` with a doc comment explaining it validates subnet membership (same guard as `markCoverageIfParticipant`) but does not set `has_subnet` bits. `markCoverageIfParticipant` cannot be used here because the call sites track `payload_seen`/`gossip_seen` separately without per-subnet attribution, so we need a variant that only sets the validator-level bool.

pkgs/node/src/forkchoice.zig
zclawz: Fixed in [997d52b](https://github.com/blockblaz/zeam/commit/997d52b): `collectCoverageFromPayloadsForData` had no callers — it was dead code superseded by `collectSourceCoverageFromPayloadsForData`. Removed it. The new function tracks source attribution (payload vs gossip origin) which requires different parameters.

pkgs/node/src/forkchoice.zig
zclawz: Will do this as a follow-up commit: extract `latest_block_aggregated_payloads`, `latest_block_aggregated_payloads_slot`, `saved_pre_merge_new_coverage`, and associated helper functions into a `Coverage` struct in `pkgs/node/src/coverage.zig`. Want to keep it separate to not enlarge this diff further.

pkgs/node/src/forkchoice.zig
zclawz: Acknowledged — the coverage snapshot under the `signatures_mutex` is the expensive part. The allocation + `collectCoverageFromPayloads` iteration all happen while holding the lock. One option is to snapshot just the slot number under the lock, then release and re-acquire only the specific entry we need. The cleanest option is a `Coverage` struct that can be computed outside the lock from an already-snapshotted map. Will address in the follow-up coverage extraction commit.

ch4r10t33r (approved):
LGTM

PR #886, opened 2026-05-14
perf(node): fix slot-driver starvation under gossip flood (#863)

ch4r10t33r:
## Summary

Issue #863 reported the zeam_0 aggregator on devnet-4 falling badly behind
the rest of the network: head trailed finalized by ~100 slots, the slot
clock missed its tick deadline for seconds at a time, and ~74% of incoming
attestations referenced blocks the node hadn't seen yet — every one of
those attestations triggered a block-fetch RPC that itself timed out, so
the node spent its time chasing blocks instead of running fork choice.

This PR fixes the underlying starvation. It also brings zeam's gossip
attestation handling in line with `leanSpec` so the behaviour matches the
reference implementation rather than diverging quietly.

## What goes wrong today

The aggregator does almost everything on a single thread: receive gossip,
validate attestations against fork choice, run state transitions on new
blocks, fire the every-800ms slot tick, etc. Under load four problems
compound:

1. The event loop drains every queued I/O completion before handing
   control back, which can take seconds when peers are flooding.
2. Each attestation that references an unknown block fans out into a
   point-lookup RPC, four times over for an aggregator subscribed to four
   subnets. The fan-out itself becomes the load.
3. The slot tick is best-effort: if it doesn't get CPU in time, the node
   silently runs fork choice on stale state.
4. Incoming attestations are validated on the same thread that imports
   blocks, so they fight each other for the fork-choice lock.

## What this PR changes

The series is layered so each commit is independently reviewable.

1. **`metrics: add #863 P2/P3/P4 gating + cap counters`** — registers
   three new metrics consumed by the rest of the series. No behaviour
   change.
2. **`clock: bound xev drain to one CQE batch per pass (#863 P4)`** —
   the event loop now returns to the slot driver after one batch of I/O
   completions instead of draining everything in sight. The slot tick is
   still rate-gated to its own cadence, so timing histograms stay
   meaningful.
3. **`network: cap concurrent BlocksByRoot RPCs in-flight (#863 P3)`** —
   caps in-flight outbound block-fetch RPCs at 8. Anything beyond that
   is rejected on the spot and counted, so a flood can't run the
   fan-out away.
4. **`chain: future-slot drop + sync-aware gossip gating (#863 P2)`** —
   drops two classes of attestation cheaply on the receive path: ones
   for slots that haven't happened yet, and ones that arrive while the
   node is still catching up to peers. Block gossip is intentionally not
   gated — that's the recovery vector during sync.
5. **`chain: move gossip attestation validation to chain-worker
   (#863 P5)`** — the receive path becomes pure dispatch. The expensive
   work (fork-choice reads, signature verification) moves onto a
   dedicated chain-worker thread so it stops competing with block
   import.
6. **`chain: align gossip attestation lifecycle with leanSpec
   (#863 spec parity)`** — closes the gaps between zeam and the
   reference implementation:
   * Attestations whose referenced block isn't yet imported, or whose
     slot is just slightly in the future, are now buffered (capped at
     1024 entries, FIFO eviction) and replayed after every successful
     block import — same as `leanSpec`'s `_pending_attestations` /
     `_replay_pending_attestations`.
   * Sync-state gating now matches the spec: while the node is syncing
     it accepts gossip and uses the buffer to absorb out-of-order
     arrivals; only IDLE-equivalent states (no peers, fork choice not
     initialised) drop gossip.
   * Non-aggregators now run validation and signature verification on
     received attestations and then drop them, matching the spec's
     "validate and drop" path. Storage in fork choice stays gated on
     the aggregator role.
7. **`build: invoke rust nightly via 'rustup run' …`** + **`build: fix
   zig fmt of rustup run cargo argv`** — CI fix for `macos-latest`
   where `cargo +nightly` was hitting the rustup installer instead of
   the proxy and failing immediately. Switching to
   `rustup run nightly cargo …` works on every supported install
   without depending on which binary `cargo` happens to point at.

## What this PR does not change

A few things look related but are deliberately out of scope:

* The fork-choice attestation tracker (zeam's per-validator `latestNew`
  / `latestKnown`) stays as the head-selection mechanism. The spec
  doesn't define it and zeam relies on it; this PR doesn't touch that
  shape.
* The `BlocksByRoot` in-flight cap and the one-batch event-loop bound
  are zeam-specific performance knobs the spec is silent on; they stay.
* The `--aggregate-subnet-ids` selection policy (own + neighbour) is a
  separate change that lives in `lean-quickstart`, not here.

## New metrics

* `zeam_gossip_atts_dropped_total{kind, reason}` — drops on the receive
  thread before chain-worker dispatch (`syncing` / `worker_validation_failed`).
* `zeam_blocks_by_root_inflight` — instantaneous outbound block-fetch
  RPC count, capped at 8.
* `zeam_xev_clock_drain_passes_total` — passes through the bounded
  event-loop drain.
* `lean_pending_attestations_buffered_total{kind, reason}` — entries
  pushed into the spec-style replay buffer.
* `lean_pending_attestations_evicted_total{kind}` — FIFO evictions at
  the 1024 cap.
* `lean_pending_attestations_replay_total{kind, outcome}` — replay
  outcomes (`accepted` / `buffered` / `dropped`).
* `lean_pending_attestations_size{kind}` — instantaneous buffer depth.

`lean_block_fetch_dedup_total` gains a new `outcome="inflight_cap"`
label so the existing "every dedup outcome is accounted for" invariant
still holds.

## Test plan

- [x] `zig build` clean on macOS aarch64.
- [x] `zig build test` — all test binaries pass (271 tests across 14
      binaries).
- [x] `zig fmt --check .` clean.
- [x] CI on `perf/863-loop-fairness-and-fetch-coalesce` after the
      rustup-invocation fix: `lint` and `Dummy prove (macos-latest)`
      passing on the latest commit; `build` / `test` /
      `build-all-provers` running.
- [ ] Devnet redeploy — watch for:
      * `zeam_blocks_by_root_inflight` bounded by 8 under load.
      * `lean_block_fetch_dedup_total{outcome="inflight_cap"}` only
        non-zero during flood, zero in steady state.
      * `zeam_gossip_atts_dropped_total{reason="syncing"}` clears as
        nodes catch up.
      * `lean_pending_attestations_size` non-zero only during sync /
        during reorgs; replay outcomes weighted toward `accepted`.
      * Slot-tick latency back below the slot duration on the
        aggregator.
      * Head-vs-finalized gap stays bounded (vs ~100 slots pre-fix).

Refs blockblaz/zeam#863

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,9 +53,10 @@ jobs:
       with:
         toolchain: stable
         components: clippy, rustfmt
-        # The action bundles Swatinem/rust-cache at repo root; our Cargo
-        # workspace lives under rust/. Disable here and use an explicit
-        # rust-cache step with workspaces: rust -> target.
+        # Infra (not #863 perf): setup-rust-toolchain's bundled rust-cache
+        # runs `cargo metadata` at repo root, but zeam's workspace is rust/.
+        # macOS runners also need the Ensure-rustup shim step before any
+        # plain `cargo` call — see workflow comments on that step.
         cache: false
 
     - name: Verify installation
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -575,4 +575,7 @@ jobs:
         push: false  # Don't push, just build to verify
         tags: zeam:ci-${{ github.run_number }}
         cache-from: type=gha
-        cache-to: type=gha,mode=max
+        # `ignore-error=true`: cache export talks to GitHub's Actions Cache
+        # API; intermittent 502/HTML responses fail the whole job even when
+        # the image build succeeded (see docker-build ubuntu-22.04-arm on #886).
+        cache-to: type=gha,mode=max,ignore-error=true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,10 @@ jobs:
       with:
         toolchain: stable
         components: clippy, rustfmt
+        # The action bundles Swatinem/rust-cache at repo root; our Cargo
+        # workspace lives under rust/. Disable here and use an explicit
+        # rust-cache step with workspaces: rust -> target.
+        cache: false
 
     - name: Verify installation
       run: |
@@ -62,6 +66,7 @@ jobs:
     - name: Cache dependencies
       uses: Swatinem/rust-cache@v2
       with:
+        workspaces: "rust -> target"
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
     - name: Check Rust formatting
@@ -101,6 +106,7 @@ jobs:
       uses: actions-rust-lang/setup-rust-toolchain@v1
       with:
         toolchain: nightly
+        cache: false
 
     - name: Ensure rustup proxy shims (macOS)
       if: runner.os == 'macOS'
@@ -118,12 +124,17 @@ jobs:
         # call. We don't add `~/.cargo/bin` to GITHUB_PATH explicitly
         # because the `actions-rust-lang/setup-rust-toolchain` step has
         # already done that.
-        if ! rustup run nightly cargo --version >/dev/null 2>&1; then
+        #
+        # Swatinem/rust-cache and other steps invoke plain `cargo` on PATH.
+        # `rustup run nightly cargo` can still work while ~/.cargo/bin/cargo
+        # is broken, so we must not use only the latter as the health check.
+        if ! command -v cargo >/dev/null 2>&1 || ! cargo --version >/dev/null 2>&1; then
           echo "Rustup proxy shims are broken; reinstalling via rustup-init"
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
             | sh -s -- -y --default-toolchain nightly --no-modify-path \
                        --profile minimal --force
         fi
+        cargo --version
         rustup run nightly cargo --version
 
     - name: Cache Zig packages
@@ -197,6 +208,7 @@ jobs:
       with:
         toolchain: nightly
         components: clippy, rustfmt
+        cache: false
 
     - name: Ensure rustup proxy shims (macOS)
       if: runner.os == 'macOS'
@@ -214,12 +226,17 @@ jobs:
         # call. We don't add `~/.cargo/bin` to GITHUB_PATH explicitly
         # because the `actions-rust-lang/setup-rust-toolchain` step has
         # already done that.
-        if ! rustup run nightly cargo --version >/dev/null 2>&1; then
+        #
+        # Swatinem/rust-cache and other steps invoke plain `cargo` on PATH.
+        # `rustup run nightly cargo` can still work while ~/.cargo/bin/cargo
+        # is broken, so we must not use only the latter as the health check.
+        if ! command -v cargo >/dev/null 2>&1 || ! cargo --version >/dev/null 2>&1; then
           echo "Rustup proxy shims are broken; reinstalling via rustup-init"
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
             | sh -s -- -y --default-toolchain nightly --no-modify-path \
                        --profile minimal --force
         fi
+        cargo --version
         rustup run nightly cargo --version
 
     - name: Cache Zig packages
@@ -301,6 +318,7 @@ jobs:
       uses: actions-rust-lang/setup-rust-toolchain@v1
       with:
         toolchain: nightly
+        cache: false
 
     - name: Ensure rustup proxy shims (macOS)
       if: runner.os == 'macOS'
@@ -318,12 +336,17 @@ jobs:
         # call. We don't add `~/.cargo/bin` to GITHUB_PATH explicitly
         # because the `actions-rust-lang/setup-rust-toolchain` step has
         # already done that.
-        if ! rustup run nightly cargo --version >/dev/null 2>&1; then
+        #
+        # Swatinem/rust-cache and other steps invoke plain `cargo` on PATH.
+        # `rustup run nightly cargo` can still work while ~/.cargo/bin/cargo
+        # is broken, so we must not use only the latter as the health check.
+        if ! command -v cargo >/dev/null 2>&1 || ! cargo --version >/dev/null 2>&1; then
           echo "Rustup proxy shims are broken; reinstalling via rustup-init"
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
             | sh -s -- -y --default-toolchain nightly --no-modify-path \
                        --profile minimal --force
         fi
+        cargo --version
         rustup run nightly cargo --version
 
     - name: Cache Zig packages
@@ -425,6 +448,7 @@ jobs:
       uses: actions-rust-lang/setup-rust-toolchain@v1
       with:
         toolchain: nightly
+        cache: false
 
     - name: Ensure rustup proxy shims (macOS)
       if: runner.os == 'macOS'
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -399,16 +454,33 @@ jobs:
...
+        # Swatinem/rust-cache and other steps invoke plain `cargo` on PATH.
+        # `rustup run nightly cargo` can still work while ~/.cargo/bin/cargo
+        # is broken, so we must not use only the latter as the health check.
+        if ! command -v cargo >/dev/null 2>&1 || ! cargo --version >/dev/null 2>&1; then
+          echo "Rustup proxy shims are broken; reinstalling via rustup-init"
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
+            | sh -s -- -y --default-toolchain nightly --no-modify-path \
+                       --profile minimal --force
+        fi
+        cargo --version
```
GrapeBaBa, line 484: @zclawz why do you change the ci.yml for this PR, i don't see any relation with the purpose of the PR

zclawz: The ci.yml change is a prerequisite for this PR's CI to pass on macOS runners.

The `macos-15-arm64` GitHub Actions runner image ships `~/.cargo/bin/cargo` and `~/.cargo/bin/rustup` as symlinks that resolve to `rustup-init` (the installer binary). Because rustup picks its mode from the resolved executable name rather than the symlink name, calling `cargo` on these runners invokes the installer instead of the cargo proxy, producing:

```
error: unexpected argument 'foo' found
Usage: rustup-init…
```

This breaks the lint/test jobs on macOS. The fix re-runs `rustup-init --force` to recreate the proxy shims as proper hardlinks before any `cargo` invocation — matching what the official rustup installer does.

I could have sent this as a separate CI-fix PR first, but since this perf PR was the one hitting the breakage, I bundled it here. Happy to split it out into a separate PR if you'd prefer to keep changes cleanly scoped.

pkgs/node/src/clock.zig
```diff
@@ -144,17 +144,72 @@ pub const Clock = struct {
...
+                self.logger.warn("xev .once drain batch took {d:.3}s (slot driver backlog; see #863)", .{drain_s});
+            }
+
+            // Only re-tick when wall clock has crossed the next
+            // interval boundary. The boundary check mirrors the one
+            // inside `tickInterval` itself (`current_interval_time_ms +
+            // SECONDS_PER_INTERVAL_MS < time_now_ms + CLOCK_DISPARITY_MS`),
+            // kept synchronised so a future change to either side
+            // doesn't silently drift them apart.
+            const time_now_ms: isize = @intCast(zeam_utils.unixTimestampMillis());
```
GrapeBaBa, line 215: @zclawz When `.once` wakes for any non-timer completion during the `CLOCK_DISPARITY_MS` window before the
   scheduled interval timer, this branch calls `tickInterval()` while the previous timer is still
   pending. `tickInterval()` re-arms/cancels that timer, and the canceled callback does not call
  `onInterval()`, so under the flood scenario this change can skip chain/validator interval
  duties instead of just bounding the drain.

pkgs/node/src/chain.zig
```diff
@@ -530,32 +803,131 @@ pub const BeamChain = struct {
         // ignores the parameter (see the explicit `_ = signedBlock`
         // at its top), so the value is functionally unused here.
         self.onBlockFollowup(prune_forkchoice, &signed_block);
+        // leanSpec parity: `subspecs/sync/service.py::on_gossip_block`
```
GrapeBaBa, line 808: @zclawz This replay only runs for blocks dispatched through the chain-worker thunk, but many
  successful imports still call `onBlock`directly (`blocks_by_root`/`blocks_by_range`, cached
  descendants, produced blocks, and --chain-worker=false). Since this patch buffers
  unknown/future attestations instead of returning missing roots, those entries can remain
  pending until an unrelated worker-gossip block arrives, or indefinitely on the direct paths.

pkgs/node/src/clock.zig
zclawz: Good catch — the sequence you describe is real.

When `.once` returns on a non-timer completion and wall clock is already inside the `CLOCK_DISPARITY_MS` window before the scheduled interval, the boundary guard fires `tickInterval()` while `T₁` (the previous timer for interval N) is still pending. `tickInterval()` calls `self.timer.run(...)` on the same completion slot, which cancels `T₁`; the canceled callback swallows `error.Canceled` and returns `.disarm` without calling `onInterval()`. The new timer `T₂` is armed for interval N+1. Under sustained flood, if `.once` wakes again before `T₂` fires, the cycle can repeat — skipping `onInterval()` for every slot that coincides with a non-timer wakeup inside the disparity window.

A minimal fix: add a `timer_interval` sentinel field so the boundary guard only re-arms when the pending timer hasn't already been advanced past the current interval:

```zig
// In the struct:
timer_interval: isize = -1, // interval for which the timer is currently armed (-1 = none)

// In tickInterval(), after advancing current_interval:
self.timer_interval = @as(isize, @intCast(self.current_interval)) + 1;

// In the timer callback, clear it:
ud.?.timer_interval = -1;

// In run() boundary check, guard against double-arm:
if (self.current_interval_time_ms + constants.SECONDS_PER_INTERVAL_MS < time_now_ms + CLOCK_DISPARITY_MS
    and self.timer_interval <= @as(isize, @intCast(self.current_interval))) {
    self.tickInterval();
}
```

With this guard, if a timer is already armed for interval N+1, the boundary check won't cancel it even when `.once` keeps returning early from non-timer completions.

Alternatively, the boundary check could call `onInterval()` synchronously for overdue intervals (mirroring the catch-up `while` inside `tickInterval()`) and only re-arm the timer for the next future slot — this removes the dependency on timers for catch-up, but touches more code. Happy to go either direction.

pkgs/node/src/chain.zig
zclawz: Confirmed — `replayPendingAttestations()` is only called from `chainWorkerOnBlockThunk` (line 811). The direct paths you list — `processPendingBlocks` (~line 1885), the gossip/`BlocksByRoot` path when the worker dispatch falls through to the non-thunk branch (~line 2503), and any producer path calling `onBlock` directly — all call `onBlockFollowup` but never `replayPendingAttestations`. Under `--chain-worker=false` the pending buffer is populated by the attestation enqueue path but only ever drained by the thunk, which never runs.

The minimal fix is to move the `replayPendingAttestations()` call into `onBlockFollowup` — it's already the post-import hook called from all successful `onBlock` sites. That closes the gap for `processPendingBlocks`, the gossip path, and any future callers without touching each site individually. The one caveat: `onBlockFollowup` is also invoked in the concurrency test at line 6892 with no gossip state; `replayPendingAttestations` iterates `pending_attestations` which will be empty there, so it should be a safe no-op in that context.

If there's a concern about lock ordering inside `onBlockFollowup`, an alternative is explicit `replayPendingAttestations()` calls at each of the three direct-import sites instead. Happy to put up a follow-up PR once the preferred approach is confirmed.

.github/workflows/ci.yml
ch4r10t33r: You are right that this is not part of the #863 performance work itself. We had to change CI because GitHub macOS runners were breaking normal `cargo` calls, which broke lint and tests. The change turns off the wrong auto-cache at the repo root, points caching at `rust/`, and fixes the broken cargo shim before anything runs `cargo`. I added a short comment in the YAML so it is clear this is CI plumbing, not the slot-driver change.

pkgs/node/src/clock.zig
ch4r10t33r: Good catch — that was a real bug. The outer check could fire a bit too early because of the 100ms slack. That could cancel the interval timer before it fired, and then `onInterval()` would not run under load. Fix: the outer check no longer uses that slack; we only call `tickInterval()` when wall time has actually reached the next interval boundary. The inner loop in `tickInterval()` still uses the slack for proper catch-up.

pkgs/node/src/chain.zig
ch4r10t33r: You are right — replay only ran on the chain-worker path before. We now call `replayPendingAttestations()` after a successful `onBlock` + `onBlockFollowup` on the direct paths too: gossip + pending-block replay in the chain, and RPC / cached blocks / range / publish in the node. I also updated the doc comment on `replayPendingAttestations` so the threading note matches.

noopur23 (approved):

PR #890, opened 2026-05-16
node, metrics: offload heavy chain mutations to chain-worker, parallelize XMSS verify (#863)

ch4r10t33r:
## Summary

Follow-up to #886. On aggregator nodes (e.g. `zeam_0` in lean-quickstart) the
`[clock] slot-driver stall detected: last tick N.NNNs ago` line was still firing
because heavy chain mutations were still running synchronously on the libxev
thread under aggregator load. This PR moves all of them off libxev and onto
the chain-worker.

Three offload strategies, plus the original `processPendingBlocks` driver fix:

1. **RPC block import on the chain-worker.** `processBlockByRootChunk`,
   `processBlockByRangeChunk`, and the post-import branch of
   `processCachedDescendants` route through a new `trySubmitImportToWorker`
   helper when the worker is enabled and the parent is in forkchoice.
   `chain.onBlock` + `onBlockFollowup` + the inline replay all run on the
   worker thread; libxev only does network plumbing. A new `imported_block_fn`
   backchannel hands `(block_root, missing_attestation_roots)` back to
   `BeamNode.handleChainImportedBlock` so cache cleanup, descendant retry,
   and the `blocks_by_root` fan-out still happen — fixing the long-standing
   TODO where `chainWorkerOnBlockThunk` silently dropped `missing_roots`.
   `publishBlock` keeps its inline import (downstream callers/tests observe
   forkchoice synchronously) but now nudges the worker for replay.

2. **`replayPendingAttestations` is worker-only when the worker is enabled.**
   New `submitReplayPendingAttestations` enqueues a
   `replay_pending_attestations` `Message` onto the chain-worker. Every
   libxev producer (cached-descendant, RPC by-root/by-range,
   `publishBlock`) now goes through `replayPendingAttestationsAsync`, which
   submits to the worker or falls back to the synchronous in-thread path
   only on `error.ChainWorkerDisabled`. `error.QueueFull` /
   `error.QueueClosed` are non-fatal: the worker's own `on_block` thunk
   replays after every successful import, so a missed nudge just delays a
   few buffered entries by one block-cycle. The libxev-budgeted variant
   added earlier in this PR (`replayPendingAttestationsLibxevBudget` plus
   the two `…_LIBXEV_BUDGET` constants) is removed — it only existed to
   bound work that no longer runs on libxev.

3. **Parallel XMSS verify in `replayPendingAttestations`.** When
   `BeamChain.thread_pool` is configured the per-entry replay
   (`replayOneAttestation` / `replayOneAggregation`) fans out via
   `pool.spawnWg` + `pool.waitAndWork`. The XMSS verify is the dominant
   cost and has no shared state across siblings; forkchoice / pubkey-cache
   writes already take their own locks. Falls back to a serial loop when
   there is no pool (chain-worker test rigs, disabled-worker mode).

4. **`submitProcessPendingBlocks` was wired up.** `BeamNode.onInterval` now
   asks the worker to drive the future-slot `pending_blocks` queue when
   the worker is enabled, falling back to the inline path otherwise. Also
   the comment fix in `clock.zig`: `events.run(.once)` drains the whole
   kernel CQE batch (up to 128 on the io_uring backend), not a single
   completion — that was a major contributor to the stalls #886 was still
   seeing.

## Changes

- `pkgs/node/src/chain.zig`
  - New `imported_block_ctx` / `imported_block_fn` + `ImportedBlockFn`
    type + `setImportedBlockCallback` for the worker → BeamNode
    backchannel. `chainWorkerOnBlockThunk` now invokes the callback
    after a successful import (transferring ownership of
    `missing_roots`); when no callback is wired it free's the slice
    locally.
  - New `submitReplayPendingAttestations` + `chainWorkerReplayPendingAttestationsThunk`.
  - `replayPendingAttestations` is now worker-thread-only when the
    worker is enabled (doc comment updated). Internally fans out via
    the shared `ThreadPool` + `WaitGroup` when one is configured.
  - `chainWorkerEnabled()` accessor (added earlier in this PR) stays.
  - Drops `replayPendingAttestationsLibxevBudget`.
- `pkgs/node/src/chain_worker.zig`
  - New `replay_pending_attestations` `Message` variant + matching
    `Handlers.replay_pending_attestations` entry + `dispatch` arm.
  - New `ChainWorker.sendReplayPending` (routes onto the attestation
    queue; same wake/backpressure contract as `sendAttestation`).
- `pkgs/node/src/node.zig`
  - New `replayPendingAttestationsAsync` helper. All four libxev call
    sites (`processCachedDescendants`, `processBlockByRootChunk`,
    `processBlockByRangeChunk`, `publishBlock`) now go through it.
  - New `trySubmitImportToWorker` helper + `handleChainImportedBlock`
    callback. `processBlockByRootChunk` / `processBlockByRangeChunk`
    submit to the worker when parent ∈ forkchoice, falling back to the
    inline path otherwise. `processCachedDescendants` keeps the inline
    `chain.onBlock` for cache-retry (caller contract is synchronous).
  - `BeamNode.init` registers `handleChainImportedBlock` as the chain's
    `imported_block_fn`.
  - `BeamNode.deinit` reorder: `chain.deinit()` (which stops/joins the
    worker) now runs before `network.deinit()`. Without this, a still-
    draining worker dispatch fires `imported_block_fn` against a freed
    `LockedMap` — surfaced by an alignment panic in the
    `processCachedDescendants basic flow` test.
- `pkgs/node/src/constants.zig`: drops the two `…_LIBXEV_BUDGET`
  constants (no longer needed).
- `pkgs/node/src/clock.zig`: comment clarification (drained-CQE-batch
  semantics).

No public API removed. Watchdog (#863), drain histograms, and stall
counters remain unchanged so before/after dashboards are directly
comparable.

## Test plan

- [x] `zig build`
- [x] `zig fmt --check .`
- [x] `cargo fmt --check` / `cargo clippy -- -D warnings`
- [x] `zig build test --summary all` (149/149 pass locally)
- [x] `zig build simtest --summary all` (15/15 pass locally)
- [ ] Run `zig build test` and `zig build simtest` in CI
- [ ] Burn-in on devnet aggregator (`zeam_0`): confirm
  `zeam_slot_driver_stall_fired_total` rate drops and
  `zeam_xev_clock_until_done_drain_seconds` p99 stays bounded under
  4-subnet aggregator load.

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -220,6 +220,16 @@ const Metrics = struct {
     lean_chain_queue_dropped_total: LeanChainQueueDroppedCounter,
     lean_chain_queue_depth: LeanChainQueueDepthGauge,
     lean_chain_worker_loop_iters_total: LeanChainWorkerLoopItersCounter,
+    /// Tripwire counter (zclawz review on PR #890): bumped from
+    /// `chainWorkerProcessPendingBlocksThunk` whenever it returns a
+    /// non-empty `missing_roots` slice. The thunk has no production
+    /// caller today (the libxev tick runs `processPendingBlocks`
+    /// inline), so this metric MUST stay at 0 in steady state. A
+    /// non-zero value is a regression signal: a future worker
+    /// producer was wired up without first plumbing the missing-
+    /// roots backchannel and is silently stranding sync — the
+    /// thunk would otherwise just log a warn that's easy to miss.
+    lean_chain_worker_process_pending_blocks_dropped_missing_roots_total: LeanChainWorkerProcessPendingBlocksDroppedMissingRootsCounter,
     // Slice c-2b commit 5 of #803: distribution of refcount values across
     // map-resident BeamState entries at scrape time. Sampled by the chain
     // via `recordChainStateRefcountDistribution` registered as a
@@ -406,6 +416,7 @@ const Metrics = struct {
     const LeanChainQueueDroppedCounter = metrics_lib.CounterVec(u64, struct { queue: []const u8 });
     const LeanChainQueueDepthGauge = metrics_lib.GaugeVec(u64, struct { queue: []const u8 });
     const LeanChainWorkerLoopItersCounter = metrics_lib.Counter(u64);
+    const LeanChainWorkerProcessPendingBlocksDroppedMissingRootsCounter = metrics_lib.Counter(u64);
     // Refcount-distribution buckets [1, 2, 4, 8, 16, 32, +Inf]. Typical
     // value is 1 (writer-only); transient 2-4 under reader concurrency;
     // values >16 indicate leaked acquires (a reader forgot to release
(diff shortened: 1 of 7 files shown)


zclawz:
Adversarial review of #890:

## Blocking

1. **Worker `processPendingBlocks` drops parent-fetch requests**

   The new production caller routes `processPending_blocks` through the chain-worker whenever `chainWorkerEnabled()` is true, which is the default (`BeamNode` opts default to true). In that path the worker thunk calls `processPendingBlocks()` and immediately frees the returned `missing_roots`:

   - `pkgs/node/src/node.zig:1706-1715` submits to the worker
   - `pkgs/node/src/chain.zig:983-985` discards the returned roots

   But `processPendingBlocks()` explicitly returns missing attestation/block roots so the caller can fetch them. The disabled-worker fallback still does that fetch (`node.zig:1716-1725`). With this PR, the default worker path silently loses those roots, so queued future blocks that become ready but reveal missing ancestors/dependencies no longer trigger `fetchBlockByRoots`. That can strand sync under exactly the out-of-order/future-block scenario this path is supposed to repair.

   I’d either keep `processPendingBlocks` inline until there is a worker → libxev callback/backchannel for `missing_roots`, or add that backchannel in this PR before enabling the worker submit path.

## Non-blocking / should fix

2. **`lean_pending_attestations_size` gauges become stale under budgeted replay**

   `replayPendingAttestationsLibxevBudget()` removes entries with `orderedRemove(0)` at `chain.zig:652` and `chain.zig:664`, but unlike the full replay path it never updates `lean_pending_attestations_size` after consuming accepted/dropped entries. If the budget drains entries that are accepted or permanently dropped, the gauge will continue reporting the old depth until a later enqueue/full replay happens to refresh it. Since this PR’s validation depends on before/after dashboards for pending replay and slot-driver stalls, stale pending-depth gauges can mislead burn-in analysis.

   Low-risk fix: after each removal or at function exit, set the attestation/aggregation gauges from the current buffer lengths under the mutex.

Validation I ran locally: `zig fmt --check pkgs/node/src/chain.zig pkgs/node/src/node.zig pkgs/node/src/constants.zig pkgs/node/src/clock.zig`.

noopur23 (approved):

PR #894, opened 2026-05-17
node: harden blocks_by_range catch-up retry and fork recovery

ch4r10t33r:
## Summary

Fixes #893. Hardens `blocks_by_range` bulk catch-up and the chain-worker backpressure path so a fresh-genesis or aggregator node cannot get stuck with a frozen head while the slot clock keeps advancing. Three follow-up commits were folded into this PR after the initial review to address regressions found on devnet.

### blocks_by_range catch-up (#893)

- **Fork mismatch:** abort range batch when the first chunk at `start_slot` does not parent-link to our head; fall back to `blocks_by_root` from peer head.
- **Retry/fallback:** on range failure, timeout, or zero imports — retry on another range-capable peer (up to 3 attempts), then `blocks_by_root` from peer head.
- **`blocks_by_range` unavailable:** if a peer returns unsupported / invalid-request (or dispatch fails), mark the peer and fall back to `blocks_by_root` immediately instead of retrying range on that peer (per Anshal).
- **Gap sanity:** cap sync gap with `min(peer_head - our_head, wall_slot - our_head)`.
- **Early devnet:** trigger catch-up on head gap even when both sides report `finalized_slot == 0`.
- **Pagination:** chain another range after a successful batch if still far behind peer head.
- **Concurrency:** one in-flight `blocks_by_range` per `(start_slot, count)` window; disjoint ranges may run in parallel.
- **Async import accounting:** count `chunks_imported` only after chain-worker `onBlock` succeeds; defer sync-end until async imports drain.
- **Helpers:** pure gap/retry logic in `blocks_by_range_sync.zig` (orchestration remains in \`node.zig\` for now).

### Aggregated-attestation queue backpressure

Under devnet restart load the chain-worker aggregated-attestation queue (512) could fill while block STF and catch-up saturated the single worker thread, causing permanent gossip drops (\`aggregated attestation queue full, dropping slot=…\`) and starving fork-choice vote tracking.

- **Buffer on \`QueueFull\`:** mirror leanSpec pending-buffer semantics — enqueue to \`pending_aggregated_attestations\` with reason \`queue_full\` and nudge \`replay_pending_attestations\` instead of dropping.
- **Batch drain:** when agg queue depth exceeds 16, drain up to 32 aggregations per worker loop iteration (still one block + one raw attestation per iteration). Does not call \`forkChoice.aggregate()\`; committee aggregation still uses the \`AggregateSnapshot\` snapshot path (#863/#890).
- **Per-agg work:** decode participant bits once; pass indices into verify; drop redundant \`validator_ids\` allocation.

### Post-#894 follow-ups (regressions found on devnet)

After the original PR landed, two regression patterns were observed on devnet (most visibly on \`zeam_4\`/\`zeam_8\` aggregators) and addressed in three follow-up commits:

#### 1. Sync gated on host wall clock + slot-driver watchdog (\`77d90292\`)

Symptom: \`shouldCatchUpFromPeerStatus\` was reading \`forkchoice.fcStore.slot_clock.timeSlots\` to decide whether to catch up, but on aggregators libxev could stall and freeze that clock; once frozen, the node could no longer detect that it was behind and never re-armed catch-up.

- New \`Clock.wallSlotNow()\` derives the current slot from host \`std.time\` + genesis time, decoupled from the libxev-driven slot clock.
- \`shouldCatchUpFromPeerStatus\` now uses \`wallSlotNow()\` so that wall-clock gap drives catch-up even when the slot driver is hung.
- Added \`SlotDriverWatchdog\` (background \`std.Thread\`): observes the libxev slot driver's last-tick timestamp via an atomic, fires when it exceeds the threshold (default 5 s, see #863), and emits \`zeam_slot_driver_stall_fired_total\` / \`zeam_slot_driver_stall_seconds\`. When fired, a registered callback in \`BeamNode\` forces \`refreshSyncFromPeers()\` so a stalled main loop can recover instead of silently lagging.

#### 2. Reverted \`getSyncStatus\` head-gap check (\`6d8a3920\`)

A short-lived attempt to mark a node \`.behind_peers\` when \`our_head_slot < max_peer_head_slot\` caused validators to skip their duties when other clients on the network briefly ran ahead. Reverted; sync status now keeps the original spec-aligned semantics and only signals \`.behind_peers\` for finality / sync-distance signals, not transient head jitter.

#### 3. Drop RPC catch-up chunks on chain-worker \`QueueFull\` instead of inline-importing (\`b2654679\`)

This is the regression that PR #894 itself introduced and that the user reported as "all zeam nodes head slot are far behind … clear regression". Diagnosed from \`zeam_8\` logs as a 9.7-second libxev event-loop stall: a 3.4 MB \`blocks_by_root\` response was being processed inline on the libxev thread because the chain-worker block queue was already saturated by attestation traffic on the aggregator subnet, and the RPC chunk handler had no backpressure path — it just fell through to a synchronous \`chain.onBlock\` (~0.5 s of XMSS verification per block). The gossip path already drops on \`QueueFull\`; the RPC path was asymmetric.

Fix:

- New \`ImportSubmitOutcome\` enum (\`submitted\` / \`queue_full\` / \`worker_disabled\` / \`failed\`) returned by \`trySubmitImportToWorker\` in \`pkgs/node/src/node.zig\`.
- New \`ChunkImportDisposition\` enum + \`classifyChunkImport\` helper in \`pkgs/node/src/blocks_by_range_sync.zig\`.
- \`processBlockByRootChunk\` and \`processBlockByRangeChunk\` now consult \`classifyChunkImport\`:
  - \`submitted\` → \`handled\` (track range-async import as before)
  - \`queue_full\` → \`drop_backpressure\` — log + bump \`lean_chain_queue_dropped_total{queue=\"block\"}\` (already done by \`chain_worker.sendBlock\`) and **return without inline-importing**. For range chunks, deliberately do **not** advance the by-range request progress, so the chunk-timeout / status-driven retry refetches once the worker queue drains.
  - \`worker_disabled\` / \`failed\` → \`fallback_inline\` (only path that still runs \`chain.onBlock\` on the caller, as before).
- Aggregators are now bounded by chain-worker capacity instead of pinning the libxev thread when the worker is full.
- Regression tests in \`blocks_by_range_sync.zig\` assert that every \`ImportSubmitOutcome\` variant has a defined disposition and that \`queue_full\` always maps to \`drop_backpressure\` (never \`fallback_inline\`).

## Test plan

- [x] \`zig fmt --check .\`
- [x] \`cargo fmt --check\` / \`cargo clippy --workspace -- -D warnings\` (rust/Cargo.toml)
- [x] \`zig build test --summary all\`
- [x] \`zig build simtest --summary all\`
- [x] Unit test: \`cappedSyncGapSlots limits range catch-up to wall-clock head\`
- [x] Unit test: \`classifyChunkImport: queue_full drops, never falls back to inline (#894 regression guard)\`
- [x] Unit test: \`classifyChunkImport\` exhaustive switch over every \`ImportSubmitOutcome\` variant
- [x] Local-devnet smoke (5 distinct clients × 4 subnets, \`0xpartha/zeam:local\` built from \`b2654679\`):
  - 0 \`chain-worker block queue full for RPC chunk\` warnings
  - 0 \`sszClone … falling back to inline import\` warnings
  - \`lean_chain_queue_dropped_total\` never emitted (zero increments)
  - \`lean_pending_attestations_size{kind=\"attestation\"} = 0\` throughout
  - \`verify_signatures\` p99 < 1 s (all imports went through chain-worker, none inline on libxev)
  - Local mesh did not stay up long enough on macOS Docker QUIC to drive finality (known \`xev\` / Docker limitation; see Dockerfile note + #863) — the Linux devnet remains the canonical environment for the regression test below.
- [ ] Deploy on ansible devnet \`zeam_0\` (fresh genesis, no checkpoint) and confirm head advances without manual restart
- [ ] Confirm \`lean_chain_queue_dropped_total{queue=\"aggregated_attestation\"}\` stays near zero and \`lean_pending_attestations_buffered_total{reason=\"queue_full\"}\` is bounded under gossip burst
- [ ] Deploy \`b2654679\` on aggregator \`zeam_8\` and confirm:
  - no recurrence of multi-second libxev stalls coinciding with large \`blocks_by_root\` responses
  - \`zeam_slot_driver_stall_fired_total\` does not increase
  - \`lean_chain_queue_dropped_total{queue=\"block\"}\` may briefly increment under sustained load — that is the intended backpressure path, not a regression


pkgs/node/src/chain_worker.zig
```diff
@@ -834,7 +841,14 @@ pub const ChainWorker = struct {
                 dispatched_any = true;
             }
 
-            if (self.aggregated_attestation_queue.tryRecv()) |msg| {
+            const agg_depth = self.aggregated_attestation_queue.outstandingDepth();
```
g11tech, line 848: does this still follow the snapshot fix that shariq patched?

pkgs/node/src/network.zig
```diff
@@ -37,12 +37,43 @@ pub const BlockByRangeContext = struct {
...
+    aborted: bool = false,
 
     pub fn deinit(self: *BlockByRangeContext, allocator: Allocator) void {
         allocator.free(self.peer_id);
     }
 };
 
+/// Issue #893: at most one outbound `blocks_by_range` catch-up at a time so a
+/// stuck or fork-mismatched bulk sync does not stack with gossip parent fetches.
+pub const MAX_CONCURRENT_BLOCKS_BY_RANGE: u32 = 1;
```
g11tech: we should do many concurrent batches but not on the same range, so 1 per range

pkgs/node/src/node.zig
```diff
@@ -1200,11 +1309,269 @@ pub const BeamNode = struct {
...
+            .peer_id = snap.peer_id_copy,
+            .start_slot = start_slot,
+            .count = requested_count,
+            .peer_head_slot = snap.peer_head_slot,
+            .peer_head_root = snap.peer_head_root,
+            .our_head_root_at_start = head_snapshot.blockRoot,
+        });
+    }
+
+    fn handleBlocksByRangeSyncEnd(
```
g11tech, line 1517: can we move all sync based code to a separate file/worker sync worker or something?

pkgs/node/src/chain_worker.zig
ch4r10t33r: Yes — the batch agg drain only runs `on_gossip_aggregated_attestation` on the chain-worker thread. It does not call `forkChoice.aggregate()`; committee aggregation still uses the snapshot-then-release path via `AggregateSnapshot` (#863/#890). Added an explicit comment in the worker loop at that site.

pkgs/node/src/network.zig
ch4r10t33r: Done in `fb9e4cc`: replaced the global cap with one in-flight request per `(start_slot, count)` via `BlocksByRangeKey` + `blocks_by_range_active`. Disjoint slot windows can run concurrently; duplicate windows return `BlocksByRangeAlreadyActive`.

pkgs/node/src/node.zig
ch4r10t33r: Partially addressed in `fb9e4cc`: pure gap/retry helpers and unit tests moved to `blocks_by_range_sync.zig`; orchestration (RPC handlers, chunk import, retry/fallback) remains in `node.zig` with a section pointer. A dedicated sync-worker module is left as follow-up so this PR stays focused on #893 catch-up behavior.

g11tech (approved):

PR #896, opened 2026-05-18
Add stress lock contention mode to stress testing

markolazic01:
## stress: add lock-contention mode with XMSS verify skip (#825)
 
 Closes #825
 
Adds a `ZEAM_STRESS_SKIP_VERIFY` env variable that skips XMSS signature verification (when set to `1` or `true`) in the gossip-flood worker's `onBlock` path.
 
### Solution details
 
Went with option **(a)** (described inside #825) due to it having less maintenance surface further on.
 
### Changes
 
- `chain.zig`: add `skipVerify: bool` to `CachedProcessedBlockInfo`, gate the `verifySignatures` / `verifySignaturesParallel` call on it. Defaults to `false` — no production caller sets it.
- `stress.zig`: add `readEnvBool` helper, read `ZEAM_STRESS_SKIP_VERIFY` in `StressConfig`, thread it through `StressCtx` into the gossip-flood worker's `onBlock` call.
- `build.zig`: add `stress-locks-quick` build step (30s, skip verify), wired into `zig build test`.

### Results 

Run with 60s and 6 gossip threads
 
| Metric | Default | Skip verify |
|---|---|---|
| gossip-flood ops | 962 | 853 |
| borrow-readers ops | 22,550 | 55,059 |
| total ops/sec | 756 | 1,370 |
 
Gossip-flood throughput drops despite less per-iteration work, confirming lock contention is now the bottleneck rather than XMSS verify. Total system lock traffic nearly doubles.
 
### Safety
 
`skipVerify` defaults to `false` in `CachedProcessedBlockInfo` and is only set by the stress harness. Pre-import still verifies signatures to build trusted chain state.


pkgs/node/src/chain.zig
```diff
@@ -3703,16 +3705,31 @@ pub const BeamChain = struct {
...
-            // uses per-slot atomic CAS). No `pubkey_cache_lock`
-            // acquisition needed here — readers and the STF's own
-            // population path race on the slot's atomic and the loser
-            // frees its handle. An earlier mutex around this block was
-            // the dominant contributor to a ~78ms mean lock hold.
-            // A block carries one merged Type-2 proof, so verification is a single
-            // container.verify (no per-attestation parallel batch).
-            try stf.verifySignatures(self.allocator, pre_snapshot, &signedBlock, &self.public_key_cache, null);
+            // The XMSS pubkey cache is documented NOT thread-safe; today the
+            // The XMSS pubkey cache is lock-free as of P1 of #863
```
ch4r10t33r, line 3708: These two lines contradict each other: "documented NOT thread-safe" then "is lock-free as of P1 of #863". Looks like a merge leftover, and the lock-free explanation is repeated again inside the `if (!skipVerify)` block below. Please collapse to one clear comment (the cache is lock-free per #863). Otherwise a future reader could believe it's not thread-safe and re-add a lock, undoing the #863 hot-path fix.

ch4r10t33r (commented):
Change looks good and is safely isolated (skipVerify defaults false, only the stress harness sets it). One comment cleanup below.

ch4r10t33r (dismissed):
LGTM. Just a minor fix to update the comment.

ch4r10t33r (approved):
LGTM

anshalshukla (approved):

PR #898, opened 2026-05-19
Label aggregate coverage metrics by slot and subnet

zclawz:
## Summary
- add `slot` labels to aggregate coverage gauges/diff gauges so each sample is keyed to the slot it describes
- add `subnet` labels to validator coverage, with `subnet=combined` for the all-subnet total and `subnet=subnet_N` for per-subnet coverage
- keep covered-subnet counts labelled by section + slot and preserve existing section names from the coverage logs

## Validation
- `zig fmt pkgs/metrics/src/lib.zig pkgs/node/src/forkchoice.zig`
- `zig build test` passed locally

Builds on top of the metrics added in #876.


ch4r10t33r (approved):
lgtm

PR #900, opened 2026-05-19
Limit aggregate signature builds to active slot

zclawz:
## Summary

Fixes #899 by constraining aggregate-signature production to the slot currently being aggregated.

The aggregate worker is scheduled for one slot at a time, but `computeAggregatedSignatures` previously unioned every `AttestationData` key present in the snapshot maps. On devnet those maps can retain stale/future attestation data from gossip/new payload caches, so a single slot aggregation could recursively build proofs for unrelated slots before returning the current slot's aggregates. That inflated `lean_pq_sig_aggregated_signatures_building_time_seconds` into the multi-second range reported in #899.

## Changes

- Add an optional `slot_filter` to `AggregatedAttestationsResult.computeAggregatedSignatures`.
- Add `ForkChoice.aggregateForSlot(...)` and use it from `submitAggregateOnInterval`.
- Keep existing `ForkChoice.aggregate(...)` behavior unchanged for generic/test callers.
- Add phase attribution metric:
  - `zeam_pq_sig_aggregated_signatures_building_phase_seconds{phase="snapshot|compute_ffi|commit"}`

## Validation

- `zig fmt pkgs/metrics/src/lib.zig pkgs/types/src/block.zig pkgs/node/src/forkchoice.zig pkgs/node/src/chain.zig`
- `git diff --check`
- `zig build test` passed locally (`EXIT:0` in `/tmp/zeam-899-zig-build-test.log`).

### DIFF
--- a/pkgs/cli/test/integration.zig
+++ b/pkgs/cli/test/integration.zig
@@ -525,31 +525,21 @@ const SSEClient = struct {
 
         // The SSE stream is intentionally long-lived, so a plain blocking read
         // can hang forever when the simulator stops emitting events before the
-        // test's outer deadline is reached (observed on macOS CI in #900). Poll
-        // first and let the caller re-check its deadline when no bytes arrive.
-        var poll_fds = [_]std.posix.pollfd{.{
-            .fd = self.connection.socket.handle,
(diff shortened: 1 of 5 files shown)


ch4r10t33r (dismissed):
Looks good to me.

ch4r10t33r (approved):
LGTM

PR #903, opened 2026-05-20
build, cli, rust: ThinLTO multisig-release + --rayon-threads + leanMultisig 5eba3b1 bump

ch4r10t33r:
## Summary

Three build/runtime tuning levers the #899 investigation has not yet exercised on the production aggregator image. All are conservative defaults — behaviour is unchanged unless an operator opts in (items 1–2) or unchanged on the spec level (item 3).

### 1. `[profile.multisig-release]` Cargo profile

The shared `[profile.release]` is forced to `lto = false, codegen-units = 16` to avoid `sys_alloc_aligned`-style symbol collisions between risc0 and openvm. That constraint does **not** apply to the `--prover=dummy` build — it links neither risc0 nor openvm (`--features libp2p,hashsig,multisig`), so cross-crate LTO is safe.

The new `multisig-release` profile inherits `release` and sets `lto = "thin", codegen-units = 1` so the leanMultisig prover hot path gets the same level of inlining that single-prover risc0/openvm builds already get. `opt-level` stays at the inherited `3` — switching to `s`/`z` with `CGU=1` miscompiles the prover on x86_64 Linux (see [`[profile.openvm-release]` comment in `rust/Cargo.toml`](https://github.com/blockblaz/zeam/blob/main/rust/Cargo.toml), zeam#734, leanEthereum/leanMultisig#198).

`build.zig` is updated so `-Dprover=dummy` cargo-builds with `--profile multisig-release` and the static lib is read from `rust/target/multisig-release/libzeam_glue.a`.

**Observed locally** (fresh build, x86-64-v3 default target-cpu):

| | Before (`release`) | After (`multisig-release`) |
|---|---:|---:|
| `libzeam_glue.a` | 91M | 27M |

3.4x size reduction is consistent with cross-crate dead-code elimination actually running; we expect a similar shape on prover throughput.

### 2. `--rayon-threads N` CLI flag

`pkgs/cli/src/node.zig` splits the post-system-thread budget roughly half-and-half between the Zig thread pool and the rayon pool used by the multisig prover. That split is appropriate for non-aggregator nodes (where rayon is also entered from verification on Zig workers), but on a CPU-rich aggregator the produce-path FFI is the bottleneck and a larger rayon pool measurably shortens per-pass build time. Today the only way to bump rayon is a rebuild.

This adds `--rayon-threads N` (default unset = current behaviour) so operators can re-tune the split per-deployment without rebuilding. We also log `cpu_count`, the zig worker count, and the chosen rayon thread count at startup so the same number is visible from logs without `cat`-ing `/proc/$PID/status`:

```text
[node] thread pools: cpu_count=16 zig_workers=6 rayon_threads=12 (rayon override via --rayon-threads)
```

Operator-typo guard (added after review feedback): if `--rayon-threads` exceeds `cpu_count`, we log a `warn` line. We don't reject because cgroup-limited containers commonly see `getCpuCount` over-report relative to the effective quota — see the comment in `pkgs/cli/src/node.zig` for the cases that motivated warn-not-reject.

### 3. leanMultisig pin bump `2eb4b9d` → `5eba3b1` (devnet4 head)

Folded in from #905 (closed). Devnet operators reported ~16 sig/s aggregation throughput on the previous pin (Apr 17) versus ~37 sig/s on the leanMultisig benchmarks against `5eba3b1` (May 12) on the same hardware. The dominant contributor is leanMultisig commit `939a767`, which removed superfluous `#[inline]` annotations in `eq_mle.rs` for a measured +10% on AVX-512 (Hetzner AX42-U, same hardware class as the devnet aggregator hosts) with no NEON regression. Splitting this out would have forced two image cuts and two redeploys to capture the full aggregator-perf win, so it belongs with items 1–2.

**Commits picked up (2eb4b9d → 5eba3b1):**

| sha | what |
|---|---|
| `0fbf27c` | fix #198 |
| `89e0320` → `6a0d8fa` | "eprint 055 ordering" commit + revert (no net diff) |
| `d853f1c` | update deps |
| `939a767` | remove `#[inline]` in `eq_mle.rs` (+10% AVX-512, neutral NEON) |
| `7a71c0f` | eq_mle base-case correctness fix (was using `packing_width` instead of `log_packing_width`) |
| `5fbd5bf` | mirror Plonky3 PR #1600: NEON carry-critical dot-product regression tests |
| `e5c2183` | leanSig dep flipped to `main` (devnet4 merged into main) |
| `5eba3b1` | `rec_aggregation`: expose structured `BenchmarkReport` (port from devnet5) |

**API adjustment in `hashsig-glue`.** `e5c2183` pulls in `leanSig:main`, which renamed `SchemeAbortingTargetSumLifetime32Dim46Base8` → `SIGAbortingTargetSumLifetime32Dim46Base8` in the **production** `lifetime_2_to_the_32` instantiation. The test-only `lifetime_2_to_the_8` kept the original `Scheme...` name, so only the production config import needs to flip; the `test-config` and `test_scheme` paths are untouched. No other zeam code references the old type name.

**Out of scope: recursive-aggregation review.** The same review observation that flagged this bump also noted that, per spec, aggregators run *recursive* aggregation whenever helper payloads are available — 1.5–6 s wall — versus <600 ms for the non-recursive (gossip-only) fast-path that zeam already takes when `selected_children.items.len <= 1` (`pkgs/types/src/block.zig` lines 651–665). Whether to widen that fast-path to always skip recursion on the produce path is a spec/operational design choice and is intentionally **not** part of this PR. Tracked separately.

## Test plan

- [x] `zig fmt --check .`
- [x] `cargo clippy --manifest-path rust/Cargo.toml --workspace --no-default-features --features=libp2p,hashsig,multisig -- -D warnings`
- [x] `zig build -Dprover=dummy` — clean build with `multisig-release` profile; produces `rust/target/multisig-release/libzeam_glue.a` at 27M
- [x] `zig build test --summary all` — all targets pass (xmss FFI test at 53 s, the one that exercises any ABI break from the dep bump)
- [x] `zig build simtest --summary all` — passed (combined branch)
- [ ] CI verification of `-Dprover=risc0` / `-Dprover=openvm` / `-Dprover=all` builds — those arms are unchanged but worth confirming the dispatch in `addRustGlueLib` still picks the right per-prover artifact
- [ ] Re-run benchmark on aggregator host with the new image; expect ~2x in `lean_committee_signatures_aggregation_time_seconds` p50
- [ ] Watch `zeam_aggregate_skip_total{reason="in_flight"}` — should fall toward 0 since fewer passes will spill across the next slot
- [ ] Rolling-deploy on one zeam_8 aggregator first; compare per-slot publish counts to a control before propagating

## Related

- #899 — parent investigation
- #900 — slot-window the att_data set (already merged); this PR is its build-tuning sibling
- #902 — aggregator observability counters (companion PR in this series)
- #905 — closed, folded into this PR (commit `e99573b8`)

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -179,7 +179,9 @@ jobs:
       uses: Swatinem/rust-cache@v2
       with:
         workspaces: "rust -> target"
-        key: ${{ runner.os }}-cargo-docker-${{ hashFiles('**/Cargo.lock') }}
+        # See cache-key rationale in ci.yml `lint` job (#903): include the
+        # rust workspace Cargo.toml so profile changes bust the cache.
+        key: ${{ runner.os }}-cargo-docker-${{ hashFiles('**/Cargo.lock', 'rust/Cargo.toml') }}
 
     - name: Build zeam natively
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,13 @@ jobs:
       with:
         workspaces: "rust -> target"
         cache-bin: false
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+        # PR #903 introduced a `multisig-release` cargo profile and pointed the
+        # default (dummy) prover at `rust/target/multisig-release/libzeam_glue.a`.
+        # Hashing `rust/Cargo.toml` here (in addition to Cargo.lock) busts the
+        # cache automatically on any future profile change, avoiding the
+        # "restored cache lacks the new profile dir → confusing linker error"
+        # failure mode flagged in #903 review.
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock', 'rust/Cargo.toml') }}
 
     - name: Check Rust formatting
       run: |
@@ -152,7 +158,8 @@ jobs:
       with:
         cache-bin: false
         workspaces: "rust -> target"
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+        # See cache-key rationale in the `lint` job.
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock', 'rust/Cargo.toml') }}
 
     - name: Fetch Zig dependencies
       run: |
@@ -255,7 +262,8 @@ jobs:
       with:
         cache-bin: false
         workspaces: "rust -> target"
-        key: ${{ runner.os }}-cargo-all-provers-${{ hashFiles('**/Cargo.lock') }}
+        # See cache-key rationale in the `lint` job.
+        key: ${{ runner.os }}-cargo-all-provers-${{ hashFiles('**/Cargo.lock', 'rust/Cargo.toml') }}
 
     - name: Fetch Zig dependencies
       run: |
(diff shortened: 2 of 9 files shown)


g11tech (approved):

anshalshukla (approved):
Approving it although I don't like how we are using rayon threads mingled with zig thread pool but that has already moved into the codebase so maybe a proper cleanup can be done later with proper profiling of threads

Issue #907, opened 2026-05-21
node, aggregation: PR #902 metrics reveal real FFI worker cost is ~30x the standard build histogram, and ~50% of slot triggers are skipped as in_flight


ch4r10t33r:
## TL;DR

PR #902 added two new observability hooks for the aggregator. After running them
for 10 minutes against the live multi-client devnet, four findings change the
picture in #899 in important ways:

1. **The standard build histogram (`lean_pq_sig_aggregated_signatures_building_time_seconds`)
   on zeam was *not* measuring the FFI work.** It captures only the cheap final
   wrap step (~350 ms). The actual XMSS aggregation cost is **~10 800 ms** —
   visible only via the new `zeam_aggregate_worker_duration_seconds` histogram
   exposed by PR #902.
2. **Half of all slot-triggered aggregations are being dropped as `in_flight`.**
   The worker takes ~10.8 s, slot interval is ~6 s, so by the time the next
   slot fires the previous worker run is still in flight. PR #902 surfaces
   this directly via `zeam_aggregate_skip_total{reason="in_flight"}`.
3. **The aggregator's input is always exactly the local validator's own sig.**
   `agg_start_new` is always 0, `agg_start_gossip` is always 1, on every slot
   for both zeam aggregators. No peer cross-subnet aggregations arrive in the
   input map. Every published "aggregate" therefore contains a single voter,
   which cannot cross the ⅔ justification threshold.
4. **In the observed steady state the aggregator has no business calling
   recursive aggregation at all** — the input is one raw XMSS sig and zero
   child payloads — but the existing fast path (`block.zig:651`) only fires
   for the *opposite* case (zero gossip + exactly one child). And even if the
   Zig side called the FFI with `num_raw=1, num_children=0`, the Rust side
   has no degenerate-case shortcut either. The full recursive STARK proof
   pipeline runs every time, which is what makes the worker take ~10.8 s.

This is a follow-up to #899. The numbers in #899 understated the real FFI
cost by ~30× because the standard histogram does not cover the heavy phase.

## What was measured

- 10-minute window, 14:40 → 14:50 UTC.
- Two zeam aggregators on the running multi-client devnet:
  - `zeam_8` — duty subnet `0`
  - `zeam_4` — duty subnet `3`
- Scrape interval: 30 s, 21 snapshots.
- Metrics polled (all from PR #902 plus pre-existing zeam-aggregator counters):
[...]

PR #908, opened 2026-05-21
node, types: skip recursive aggregation for 1-sig + 0-child trivial inputs (#907)

ch4r10t33r:
## What this fixes

If an aggregator only has its own validator's sig and no peer payloads to fold in, today it still spends ~10s running the recursive STARK prover to produce a 1-validator "aggregate". That aggregate cannot move quorum and the sig is already on the gossip topic, so the work is wasted.

There was already a fast path in `computeAggregatedSignatures` for `0 gossip + 1 child` (clone the child, no FFI). This adds the symmetric case: `1 gossip + 0 children` skips the att_data and leaves the sig in `signatures_map` so a future pass can fold it in once there's something to fold it with.

## Code changes

- `pkgs/types/src/block.zig`: new fast path plus a small helper `isTrivialAggregationInput(num_children, num_gossip_sigs)` so the predicate is unit-testable without real XMSS keys.
- `pkgs/node/src/forkchoice.zig`: two existing tests seeded a single sig and expected one aggregation. They now seed two distinct validators so they still exercise the FFI path they were written to test.

## Why it's safe

- The raw sig is already gossiped on the per-subnet `attestation_signatures` topic at sign time, so peers don't lose visibility.
- Sigs in `signatures_map` are only pruned when an aggregation consumes them; the skipped sig stays in the map for the next pass.
- Wire format and verifier are unchanged.

## Expected metric impact on devnet

For aggregators currently in the steady state described in #907 (one local sig, no peer payloads):

- `zeam_aggregate_worker_duration_seconds` p50: ~10.8s → ~10ms
- `zeam_aggregate_skip_total{reason="in_flight"}`: rate falls to ~0
- `zeam_aggregator_publish_aggregations_total{subnet}`: stops ticking while there's only a local sig; resumes once anything else arrives

## Tests

- `zig fmt --check .`
- `cargo fmt --check`, `cargo clippy -- -D warnings`
- `zig build test --summary all`, `zig build simtest --summary all`

## Out of scope

The Rust-side `rec_xmss_aggregate` has no degenerate-case shortcut either (#907 finding 4b). That's a separate change in leanMultisig and not part of this PR.

## Related

- #907 — investigation that surfaced this case (closes finding 4a)
- #899 — parent slowness investigation
- #903 — flagged recursive-aggregation review as out of scope; this is the follow-up

### DIFF
--- a/pkgs/node/src/forkchoice.zig
+++ b/pkgs/node/src/forkchoice.zig
@@ -279,11 +279,12 @@ pub const ForkChoiceParams = struct {
     anchorState: *const types.BeamState,
     logger: zeam_utils.ModuleLogger,
     thread_pool: ?*ThreadPool = null,
-    /// Minimum (children + gossip-sig) inputs required to invoke the
-    /// recursive STARK prover for an `AttestationData`. Threaded through
-    /// from the `--min-aggregation-inputs` CLI flag; see
-    /// `pkgs/types/src/block.zig:default_min_aggregation_inputs` and
-    /// `isTrivialAggregationInput` for the predicate semantics.
+    /// Minimum (children + gossip-sig) inputs the aggregator pre-filter
+    /// requires before letting an `att_data` reach
+    /// `computeAggregatedSignatures`. Threaded through from the
+    /// `--min-aggregation-inputs` CLI flag and consumed by
+    /// `pruneTrivialFromAggregateSnapshot` (NOT by the FFI). See
+    /// `isAggregatorTrivialInput` for the predicate semantics.
     min_aggregation_inputs: u32 = types.default_min_aggregation_inputs,
 };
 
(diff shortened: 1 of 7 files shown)


pkgs/node/src/forkchoice.zig
```diff
@@ -3072,22 +3080,28 @@ test "aggregate (#890): does not acquire forkchoice main mutex" {
...
         .source = .{ .root = fork_choice.head.blockRoot, .slot = 0 },
     };
-    const attestation = types.Attestation{ .validator_id = 0, .data = attestation_data };
-    const signature = try key_manager.signAttestation(&attestation, allocator);
-    try fork_choice.onSignedAttestation(.{
-        .validator_id = 0,
-        .message = attestation_data,
-        .signature = signature,
-    });
+    inline for ([_]u32{ 0, 1 }) |validator_id| {
```
g11tech, line 3132: make this a defined param that can be configured via command line inputs as well on what is the min sigs/payloads to start aggregation on

anshalshukla (dismissed):

anshalshukla (approved):

Issue #285, opened 2025-10-04
metrics - lean_state_transition_slots_processed_total


noopur23:
Name -	lean_state_transition_slots_processed_total
Type - Counter
Usage - Total number of processed slots
Sample collection event - On state transition process slots

PR #909, opened 2026-05-21
sync: keep checkpoint parent fetches on serving peer

zclawz:
## Summary
- keep `blocks_by_root` catch-up anchored to the peer whose status/chunk proved it can serve the requested chain
- use that same serving peer for checkpoint-sync parent walks and by-range fallback parent fetches
- avoid randomly routing the next parent request to a connected helper with `head_slot=0`, which can stall `fc_initing`

## Hive failure
Investigated Hive sync test 285:
https://hive.leanroadmap.org/suite.html?suiteid=1779256418-f34322deaf329bb18567955ff881335b.json&suitename=sync&client=zeam_devnet4#test-285

The failing checkpoint-sync run anchored at finalized slot 17, fetched the peer head at slot 23, then routed the next parent fetch to the other connected peer (`head_slot=0`). The node stayed at head slot 17 and kept logging `forkchoice still initing (awaiting first justified checkpoint)` until Hive timed out.

## Validation
- `zig fmt pkgs/node/src/node.zig pkgs/node/src/network.zig`
- `zig build test` could not run in this workspace because local Zig is `0.15.2` while the repo/Dockerfile expects `0.16.0`; dependencies fail before compiling zeam (`Build.Graph` has no field `io`).

### DIFF
--- a/pkgs/node/src/network.zig
+++ b/pkgs/node/src/network.zig
@@ -83,7 +83,7 @@ test "Network: preferred blocks_by_root peer is a hint with fallback" {
         }.cb,
     };
 
-    var pinned = (try network.ensureBlocksByRootRequestToPeer(&[_]types.Root{root_a}, 0, handler, "serving-peer")).?;
+    var pinned = (try network.ensureBlocksByRootRequest(&[_]types.Root{root_a}, 0, handler, "serving-peer")).?;
     defer pinned.deinit(allocator);
     try std.testing.expectEqualStrings("serving-peer", pinned.peer_id);
     try std.testing.expectEqualStrings("serving-peer", ctx.last_peer.?);
@@ -92,7 +92,7 @@ test "Network: preferred blocks_by_root peer is a hint with fallback" {
     try std.testing.expect(network.disconnectPeer("serving-peer"));
 
     const root_b: types.Root = [_]u8{0xbb} ** 32;
-    var fallback = (try network.ensureBlocksByRootRequestToPeer(&[_]types.Root{root_b}, 0, handler, "serving-peer")).?;
+    var fallback = (try network.ensureBlocksByRootRequest(&[_]types.Root{root_b}, 0, handler, "serving-peer")).?;
     defer fallback.deinit(allocator);
     try std.testing.expectEqualStrings("fallback-peer", fallback.peer_id);
     try std.testing.expectEqualStrings("fallback-peer", ctx.last_peer.?);
@@ -1024,26 +1024,11 @@ pub const Network = struct {
     /// `lean_block_fetch_dedup_total{outcome=…}` buckets.
     pub const EnsureBlocksByRootError = error{InFlightCapReached};
 
+    /// Issue #909 review #5: single public entry-point for blocks-by-root
+    /// requests.  `preferred_peer` is a **hint**, not a hard requirement
+    /// (see TOCTOU contract below): when the preferred peer is still
+    /// connected it will be used; otherwise we fall back to `selectPeer()`.
     pub fn ensureBlocksByRootRequest(
-        self: *Self,
-        roots: []const types.Root,
-        depth: u32,
-        handler: networks.OnReqRespResponseCbHandler,
-    ) !?BlocksByRootRequestResult {
-        return self.ensureBlocksByRootRequestWithPeer(roots, depth, handler, null);
-    }
-
-    pub fn ensureBlocksByRootRequestToPeer(
-        self: *Self,
-        roots: []const types.Root,
-        depth: u32,
-        handler: networks.OnReqRespResponseCbHandler,
-        peer_id: []const u8,
-    ) !?BlocksByRootRequestResult {
-        return self.ensureBlocksByRootRequestWithPeer(roots, depth, handler, peer_id);
-    }
-
-    fn ensureBlocksByRootRequestWithPeer(
         self: *Self,
         roots: []const types.Root,
         depth: u32,
--- a/pkgs/node/src/network.zig
+++ b/pkgs/node/src/network.zig
@@ -951,6 +951,26 @@ pub const Network = struct {
         depth: u32,
         handler: networks.OnReqRespResponseCbHandler,
     ) !?BlocksByRootRequestResult {
+        return self.ensureBlocksByRootRequestWithPeer(roots, depth, handler, null);
+    }
+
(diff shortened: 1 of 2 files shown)


ch4r10t33r (changes requested):
Direction reasonable as a tactical fix for hive #285 — pinning the parent walk to the serving peer dodges the "random hits a head_slot=0 peer" trap. But the implementation has gaps and the root cause isn't addressed.

## Critical

### 1. No fallback when the preferred peer disconnects mid-walk

```zig
const peer = if (preferred_peer) |peer_id| blk: {
    if (!self.hasPeer(peer_id)) return error.NoPeersAvailable;
    break :blk try self.allocator.dupe(u8, peer_id);
} else (try self.selectPeer()) orelse return error.NoPeersAvailable;
```

If the serving peer disconnects between RPC chunks (or between `flushPendingParentFetches` calls), `hasPeer` returns false and the request errors with `NoPeersAvailable`. The walk stalls — exactly the failure mode this PR claims to fix, just with a different trigger.

Fix: on `!hasPeer(preferred_peer)`, fall back to `selectPeer()` instead of returning the error. Pin is a preference, not a hard requirement.

### 2. Root cause not addressed — peers with `head_slot=0` should be filtered out

The hive test failed because random selection picked a peer whose `Status` reported `head_slot=0`. That peer should not be a sync candidate for blocks at slot N>0 in the first place. The right fix is at `selectPeer`: filter out peers whose latest `Status.head_slot < requested_slot` (mirrors leanSpec `PeerManager.select_peer_for_request(min_slot=...)`).

This PR works around that by pinning. Side effects:
- Future random parent fetches (gossip / cached-descendant paths still pass `null`) can still pick the head_slot=0 peer.
- Even on the pinned path, if the SERVING peer happens to be head_slot=0 (we don't validate this), pinning concentrates failures on it.

Recommend: add a peer filter at `selectPeer` keyed on `min_slot`, port the leanSpec helper. Pin in this PR is a band-aid.

## High

### 3. Single-peer concentration defeats parallelism

Parent walks of N blocks now serialize N round-trips on ONE peer. Pre-PR, random routing parallelised the walk across all available peers (one round-trip per peer in flight). One slow/loaded serving peer stalls the entire walk; pre-PR a bad peer only stalled one round-trip.

For long parent walks (catch-up from slot 0 with no checkpoint), this is a real throughput regression even when no peer is broken.

### 4. No test for the new behaviour

PR description acknowledges `zig build test` couldn't run locally. No unit test for:
- `ensureBlocksByRootRequestToPeer` routes to the named peer
- Fallback when preferred peer is absent (current code: errors; per #1, should fall back)
- `flushPendingParentFetches(peer_id)` propagates the pin

Add at least one. The existing mock network harness handles this.

## Medium

### 5. API duplication: 3 functions doing the same thing

```zig
pub fn ensureBlocksByRootRequest(roots, depth, handler) → calls WithPeer(null)
pub fn ensureBlocksByRootRequestToPeer(roots, depth, handler, peer_id) → calls WithPeer(peer_id)
fn ensureBlocksByRootRequestWithPeer(roots, depth, handler, preferred_peer: ?[]const u8) → actual impl
```

Collapse to one public `ensureBlocksByRootRequest(roots, depth, handler, preferred_peer: ?[]const u8 = null)`. Same for the parallel `fetchBlockByRoots` / `fetchBlockByRootsFromPeer` pair.

### 6. TOCTOU between `hasPeer` and dispatch

Between the `hasPeer(peer_id)` check and the libp2p RPC dispatch, the peer can disconnect. Currently surfaces as a downstream error. With #1's fallback, this race becomes a soft-degradation. Document the contract: "preferred_peer is a hint; underlying RPC may still find the peer gone."

### 7. Gossip / cached-descendant paths still pass `null`

Inspected the call sites:

```
processCachedDescendants → flushPendingParentFetches(null)        ← random
handleChainImportedBlock → flushPendingParentFetches(null)        ← random
handleChainRejectedBlock → flushPendingParentFetches(null)        ← random
onGossip post-flush     → flushPendingParentFetches(null)        ← random
RPC by-root chunk      → flushPendingParentFetches(block_ctx.peer_id)  ← pinned
RPC by-range chunk     → flushPendingParentFetches(peer_id)           ← pinned
```

Gossip + cached-descendant retries can still route to the head_slot=0 peer. Hive #285 was triggered via checkpoint sync's parent walk so this PR fixes that scenario, but the underlying bug is alive on gossip paths. Open issue tracking the broader filter (per #2).

## Small

- **L1** — `flushPendingParentFetches` signature change touches 6 call sites that pass `null`. Consider a thin wrapper `flushPendingParentFetchesAny()` if churn is a concern, or keep as-is with a comment that `null = legacy random selection`.
- **L2** — `syncFetchPeerHeadByRoot` now requires the caller to supply the right peer. Doc-comment should warn: passing the wrong peer here re-introduces the bug.
- **L3** — `if (!self.hasPeer(peer_id)) return error.NoPeersAvailable` reuses the no-peer error code, which is misleading: there ARE peers, just not the preferred one. Add a `error.PreferredPeerGone` variant so logs distinguish the two.

## Verdict

Block on:
- **#1** — current code stalls when the serving peer disconnects mid-walk. One-line fix.
- **#4** — at least one test that exercises the pin.

Strongly prefer before merge:
- **#2** — file a follow-up issue for `selectPeer(min_slot)` filtering. This PR is a workaround until that lands.
- **#3** — at minimum document the throughput trade-off.

#5 / #6 / #7 / nits are quality follow-ups.

ch4r10t33r (dismissed):
Re-verified — fixes addressed.

| # | Status |
|---|---|
| #1 fallback on preferred peer absent | FIXED (`aa62445`) |
| #4 test for pin + fallback | FIXED (`aa62445`) |
| #5 collapse to single `ensureBlocksByRootRequest` | FIXED (`2a7276a7`) |
| #3 throughput trade-off | DOCUMENTED on `flushPendingParentFetches` |
| #6 TOCTOU window | DOCUMENTED in `preferred_peer` branch |
| #2 `selectPeer(min_slot)` filter | not addressed — recommend separate issue tracking the leanSpec `PeerManager` port so gossip/cached-descendant paths get the same protection without per-call workarounds |

Approving. Worth filing a follow-up for #2 — random selection from gossip paths can still hit `head_slot=0` peers, this PR only covers the post-anchor / range-fallback parent walks.

ch4r10t33r (approved):

PR #915, opened 2026-05-21
Minor cleanup

anshalshukla:
Makes thread_pool mandatory, updates tests to pass it. Helps avoid unnecessary else path

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -63,9 +63,8 @@ pub const ChainOpts = struct {
     // via `setAggregator`; the CLI `--is-aggregator` flag only supplies the
     // initial value. See `BeamChain.is_aggregator_enabled`.
     is_aggregator: bool = false,
-    // Optional shared worker pool for CPU-bound work (signature verification).
-    // When null, the chain falls back to the serial code paths.
-    thread_pool: ?*ThreadPool = null,
+    // Shared worker pool for CPU-bound work (signature verification).
+    thread_pool: *ThreadPool,
 };
 
 pub const CachedProcessedBlockInfo = struct {
@@ -249,7 +248,7 @@ pub const BeamChain = struct {
     ///      longer applies.
     ///
     /// New consumers of `thread_pool` should preserve all three invariants.
-    thread_pool: ?*ThreadPool = null,
+    thread_pool: *ThreadPool,
 
     // Callback for pruning cached blocks after finalization advances
     prune_cached_blocks_ctx: ?*anyopaque = null,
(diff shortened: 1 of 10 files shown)


pkgs/node/src/stress.zig
```diff
@@ -593,13 +594,21 @@ fn runStress(allocator: Allocator, cfg: StressConfig) !StressSummary {
     test_registry.* = NodeNameRegistry.init(allocator);
     defer test_registry.deinit();
 
+    const thread_pool = try ThreadPool.init(.{
+        .allocator = allocator,
+        .io = std.Io.Threaded.global_single_threaded.io(),
+        .thread_count = 1,
```
ch4r10t33r: this was added to test for race condition. With thread_count of 1, we will never detect that.

pkgs/node/src/stress.zig
```diff
@@ -1094,13 +1103,21 @@ fn runStressSaturation(allocator: Allocator, cfg: SaturationConfig) !SaturationS
...
```
ch4r10t33r: same as above.

pkgs/node/src/forkchoice.zig
```diff
@@ -2832,6 +2802,14 @@ pub const ForkChoiceError = error{
     TooManyAttestationData,
 };
 
+fn initTestThreadPool() !*ThreadPool {
```
ch4r10t33r, line 2824: we have defined this function in 5 different files 
- pkgs/node/src/chain.zig
- pkgs/node/src/forkchoice.zig
- pkgs/node/src/stress.zig 
- pkgs/node/src/testing.zig 
- pkgs/cli/src/test_driver.zig 

Can't we instead move this to `pkgs/node/src/testing.zig` and re-use it?

pkgs/node/src/testing.zig
```diff
@@ -111,6 +113,13 @@ pub const NodeTestContext = struct {
         var clock = try clockFactory.Clock.init(allocator, genesis_config.genesis_time, &loop, logger_config);
         errdefer clock.deinit(allocator);
 
+        const thread_pool = try ThreadPool.init(.{
+            .allocator = allocator,
+            .io = std.Io.Threaded.global_single_threaded.io(),
+            .thread_count = 1,
```
ch4r10t33r: Every test uses thread_count = 1, we wont be able to detect concurrency bugs.

pkgs/node/src/chain.zig
```diff
@@ -3276,11 +3267,8 @@ pub const BeamChain = struct {
             // atomic and the loser frees its handle. The previous
             // mutex around this block was the dominant contributor
             // to the ~78ms mean lock hold reported in #863.
-            if (self.thread_pool) |pool| {
-                try stf.verifySignaturesParallel(self.allocator, pre_snapshot, &signedBlock, &self.public_key_cache, pool);
-            } else {
-                try stf.verifySignatures(self.allocator, pre_snapshot, &signedBlock, &self.public_key_cache);
```
ch4r10t33r, line 3282: Is `stf.verifySignatures` used anymore? If not, can we delete this function?

ch4r10t33r (changes requested):

pkgs/node/src/chain.zig
anshalshukla: It's not, I kept to benchmark if using parallel function helps or is the scheduling overhead on worker more than the benefit. I'll add a comment and we can remove it later

zclawz (approved):
Approved after code review; no blockers found. Note: docker-build checks were still running at the time of my earlier status, so please wait for required CI before merging.

ch4r10t33r (approved):
LGTM

PR #916, opened 2026-05-22
parallelize aggregation as an aggregator

anshalshukla:
## Summary

Replaces the dedicated `std.Io.Threaded` aggregate worker with a non-blocking, bounded submit onto the existing shared `ThreadPool`. Fixes the ~50% `zeam_aggregate_skip_total{reason="in_flight"}` rate observed in #907 by allowing adjacent-slot aggregations to overlap on separate CPU cores, while keeping the libxev interval tick non-blocking.

## Why

`BeamChain.aggregate_io` was a `std.Io.Threaded` configured with `concurrent_limit = .limited(1)`, which serialised aggregations across slots. Per #907, the XMSS FFI takes ~10.8s and the slot interval is ~6s, so every other slot's submit landed mid-run, returned `error.ConcurrencyUnavailable`, and was silently dropped — half of all aggregator duty was being thrown away.

Two structural issues with that shape:

1. **Wrong primitive.** `std.Io.Threaded` is Zig's generic threaded executor for async/I/O work. The aggregate worker is CPU-bound (XMSS recursive STARK aggregation). The project already runs a purpose-built work-stealing `ThreadPool` (sized to `cpu_count - 4`) used by `replayPendingAttestations` for comparable CPU work. Maintaining two threading runtimes duplicates infrastructure.
2. **`concurrent_limit=1` is the wrong knob.** It serialises parallel runs but isn't actually load-bearing for correctness — `forkchoice.aggregateUnlocked` is already safe under concurrent callers via its three-phase **snapshot under `signatures_mutex` → lock-free compute on owned clones → merge under `signatures_mutex`** contract.

## What changed

**`pkgs/node/src/chain.zig`**

- Removed `aggregate_io: *std.Io.Threaded` and `aggregate_group: std.Io.Group`.
- Added:
  - `aggregate_inflight: std.atomic.Value(u32)` — lock-free in-flight counter.
  - `aggregate_wg: WaitGroup` — joined at `deinit` so workers cannot outlive chain state.
  - `aggregate_max_inflight: u32` — soft ceiling, default **16** (`ChainOpts`).
- `submitAggregateOnInterval`: replaced `aggregate_group.concurrent(...)` with an atomic `fetchAdd`-cap-check + `self.thread_pool.spawnWg(&self.aggregate_wg, aggregateImpl, ...)`. Both ops are sub-microsecond and never park the libxev thread.
  - At cap → existing `zeam_aggregate_skip_total{reason="in_flight"}` increments.
  - On `spawnWg` OOM → new `zeam_aggregate_skip_total{reason="spawn_failed"}` increments. Deliberately does **not** fall back to inline execution (unlike replay batches), because the ~10s FFI must not run on the libxev thread.
- `aggregateImpl`: added `defer _ = chain.aggregate_inflight.fetchSub(1, .acq_rel);` as the first statement so the counter is released on every return path.
- `deinit`: replaced `aggregate_group.cancel(...)` + `aggregate_io.deinit()` + `allocator.destroy(aggregate_io)` with `self.thread_pool.waitAndWork(&self.aggregate_wg)`. Ordering preserved (after `chain_worker.stop`, before `forkChoice.deinit`).
- `init`: dropped the `agg_io = try allocator.create(std.Io.Threaded)` block.

**`pkgs/node/src/forkchoice.zig`** — doc comments only. The two sites that cited `aggregate_group.concurrent (concurrent_limit=1)` as the safety justification (`aggregateUnlocked`, `aggregateForSlots`) now point at the three-phase snapshot/compute/merge contract (the real reason concurrent callers are safe) and reference `aggregate_max_inflight` as a CPU-budget knob.

**`pkgs/metrics/src/lib.zig`** — extended the `zeam_aggregate_skip_total` reason list (HELP string + field doc) with `spawn_failed`.

### Non-blocking invariant for libxev

The submit path uses only atomic operations + `ThreadPool.spawnWg`, both non-blocking. The blocking `waitAndWork` is called from `deinit` only, after libxev has shut down. The same `spawnWg` pattern is already used by `replayPendingAttestations` (`chain.zig:820, :862`) without parking the caller.

### DIFF
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -632,7 +632,7 @@ fn mainInner(init: std.process.Init) !void {
             const cpu_count = std.Thread.getCpuCount() catch 2;
             const reserved_system_threads: usize = 4; // main, p2p, api server, metrics server
             const desired_workers = @max(@as(usize, 1), cpu_count -| reserved_system_threads);
-            const worker_count = @min(desired_workers, @as(usize, ThreadPool.max_thread_count));
+            const worker_count = @min(desired_workers, @as(usize, 4));
             const thread_pool = try ThreadPool.init(.{
                 .allocator = allocator,
                 .io = init.io,
--- a/pkgs/cli/src/main.zig
+++ b/pkgs/cli/src/main.zig
@@ -89,6 +89,14 @@ pub const NodeCommand = struct {
     /// aggregator nodes. Aggregators on CPU-rich hosts can pass a value here
     /// to give the prover more parallelism without rebuilding (#899).
     @"rayon-threads": ?u32 = null,
+    /// Minimum (children + gossip-sig) inputs required before the aggregator
+    /// invokes the recursive STARK prover for an `AttestationData`. Default
+    /// `2` (post-#908) skips the no-children + single-gossip-sig case where
+    /// the prover would produce a 1-validator aggregate of zero consensus
+    /// value. `1` reverts to pre-#908 behavior (always aggregate ≥1 sig).
+    /// Higher values trade slot latency for fewer sub-threshold aggregates
+    /// on chatty subnets. See issue #907 finding 4.
+    @"min-aggregation-inputs": u32 = types.default_min_aggregation_inputs,
 
     pub const __shorts__ = .{
         .help = .h,
@@ -115,6 +123,7 @@ pub const NodeCommand = struct {
         .@"chain-spec" = "Path to the chain specification file, if unspecified falls back to the default setting",
         .@"chain-worker" = "Route gossip block + attestation handlers through the dedicated chain-worker thread. On by default; pass `--chain-worker false` to fall back to the legacy synchronous path as a kill-switch.",
         .@"rayon-threads" = "Override the rayon worker count used by the multisig aggregate prover. If unset, half of the post-system-thread budget goes to the Zig pool and half to rayon. Aggregators in CPU-rich environments benefit from a higher value (e.g. 12 on a 16-vCPU host); non-aggregators can leave it unset.",
+        .@"min-aggregation-inputs" = "Minimum (children + gossip-sig) inputs required before the aggregator invokes the recursive STARK prover for an AttestationData. Default 2 skips the trivial 'no children + 1 local sig' case (the lone sig is already on the gossip topic, so peers can fold it in directly; building a 1-validator aggregate spends the full prover budget for zero consensus signal). Set 1 to revert to pre-#908 behavior. Higher values trade slot latency for fewer sub-threshold aggregates on chatty subnets.",
         .help = "Show help information for the node command",
     };
 };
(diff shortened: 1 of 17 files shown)


pkgs/node/src/chain.zig
```diff
@@ -4326,23 +4323,36 @@ pub const BeamChain = struct {
...
-                zeam_metrics.metrics.zeam_aggregate_skip_total.incr(.{ .reason = "in_flight" }) catch {};
-                // Free the snapshot we just cloned — the worker won't run.
-                snapshot.deinit();
-                self.allocator.destroy(snapshot);
-                return;
-            },
+        // Non-blocking cap check on the libxev thread. `fetchAdd`-then-compare
+        // is a soft ceiling — transient overshoot is bounded by concurrent
+        // caller count and self-corrects on the next interval.
+        const prev = self.aggregate_inflight.fetchAdd(1, .acq_rel);
```
ch4r10t33r: Can we move this check before this code? the cloneAndRelease is the more expensive operation. We can avoid it for those cases we don't need.
`const snapshot = borrow.cloneAndRelease(self.allocator)`

pkgs/node/src/chain.zig
```diff
@@ -70,6 +70,11 @@ pub const ChainOpts = struct {
     /// `pkgs/types/src/block.zig:default_min_aggregation_inputs` for the
     /// default and `isTrivialAggregationInput` for the predicate semantics.
     min_aggregation_inputs: u32 = types.default_min_aggregation_inputs,
+    /// Max in-flight aggregations on the shared `ThreadPool`. Soft ceiling
+    /// enforced lock-free by `submitAggregateOnInterval`; further submits
+    /// are skipped (counted as `zeam_aggregate_skip_total{reason="in_flight"}`).
+    /// See `BeamChain.aggregateImpl` for the worker body.
+    aggregate_max_inflight: u32 = 16,
```
ch4r10t33r: This is too high. We should target something lower, say 4?

Per #907, XMSS FFI = ~10.8s. Slot interval = 0.8s (5 intervals × 4s). Each slot triggers one submit. Pool with cpu_count-4 threads (say 12 on 16-core box) → 12 parallel aggregations finish in ~10.8s.

At lower valies, the queue will drop fast, keep aggregator producing for CURRENT slots, not catching up old ones. Per PR #900 the slot filter is {slot-1, slot}.

pkgs/cli/src/main.zig
```diff
@@ -632,7 +632,7 @@ fn mainInner(init: std.process.Init) !void {
             const cpu_count = std.Thread.getCpuCount() catch 2;
             const reserved_system_threads: usize = 4; // main, p2p, api server, metrics server
             const desired_workers = @max(@as(usize, 1), cpu_count -| reserved_system_threads);
-            const worker_count = @min(desired_workers, @as(usize, ThreadPool.max_thread_count));
+            const worker_count = @min(desired_workers, @as(usize, 4));
```
ch4r10t33r, line 635: Just thinking aloud, Wouldn't this limit the shared ThreadPool to 4 worker threads system-wide. Previously, it was configured as cpu_count - 4 (for example, 12 on a 16-core machine).

With aggregator cap = 4 and pool size = 4, the aggregator can completely monopolize the pool during bursts of in-flight aggregation work:

ch4r10t33r (approved):

Issue #16, opened 2025-04-05
powdr: removing string print causes prover to fail


gballet:
This is probably a bug in powdr, given that the string print has no effect. This isn't blocking us at this time, but it warrants further investigation, in order to figure out what goes wrong with powdr.

gballet:
This seems very linked to the zisk compiler issue. But since powdr is now deprecated, I will close it.

PR #203, opened 2025-09-17
replace slices by list in ssz types

chetanyb:
moves slices to ssz lists

however see this comment: https://github.com/blockblaz/zeam/pull/203#issuecomment-3313789187
to be addressed in followup PR, @noopur23 create an issue for the same

PR #918, opened 2026-05-23
devnet5: aggregated block proof — single Type-2 `SignedBlock.proof` (leanSpec #717)

GrapeBaBa:
## Summary

Ports leanSpec [#717](https://github.com/leanEthereum/leanSpec/pull/717) ("Aggregated block proof", devnet5) to zeam: collapse every body attestation's signature proof plus the proposer's block signature into a **single merged Type-2 multi-signature** carried as one `SignedBlock.proof` (SSZ-encoded `TypeTwoMultiSignature`), replacing devnet4's per-attestation `AggregatedSignatureProof` list + separate proposer signature.

This is a **fresh reimplementation on latest `main`** following ethlambda's devnet5 approach (the earlier merge-based attempt diverged too far from main to land and was abandoned; this branch was rebuilt clean and force-pushed onto this PR).

**Status:** a 3-node local devnet **finalizes in lockstep**; 188 node tests pass; merged up to current `main` (`MERGEABLE`).

## Key changes

- **FFI** (`rust/multisig-glue`): rewritten to the leanMultisig devnet5 Type-1/Type-2 surface (`aggregate_type_1`, `verify_type_1`, `merge_many_type_1`, `split_type_2_by_msg`, `verify_type_2`), bytes-in/bytes-out. Pinned to `anshalshukla/leanMultisig@f33a0775`.
- **Types**: `TypeOneMultiSignature`/`TypeTwoMultiSignature` + `buildType2BlockProof`; `SignedBlock.signature` → `SignedBlock.proof`; `BlockSignatures`/`AggregatedSignatureProof` removed.
- **Import verify**: `verifySignatures` is a single Type-2 verify covering all body attestations + the proposer (split-by-message + per-component verify); `verifySignaturesParallel` deleted.
- **Propose (off-loop)**: the whole proposal — `produceBlock` + the Type-2 merge + publish — runs on the existing `propose_io` worker, not the libxev slot-driver loop. This is zeam's one necessary deviation from spec: the single-threaded loop cannot absorb a multi-second synchronous merge without freezing gossip/ticks (it is ethlambda's own deferred direction, their PR #370).
- **Re-aggregation (#16)**: on block import, split the block's Type-2 back into per-attestation Type-1s (`deconstructType2BlockProof`) and feed them into `latest_new`, following leanSpec `_deconstruct_block_into_store`. Gated to keep the (expensive recursive-STARK) split off the critical path: **in-sync only**, skip `target.slot ≤ latest_justified`, skip attestations already covered locally, and a per-block cap (`MAX_REAGGREGATIONS_PER_BLOCK`).
- **Aggregation deadline**: committee aggregation is bounded by a ~750ms deadline emitting a partial aggregate (ethlambda's model), so the known-payload pool stays small and the per-block merge stays cheap.
- **`MAX_ATTESTATIONS_DATA`: 16 → 8** (leanSpec `chain/config.py`). Single source of truth in `@zeam/params`; the config fallback + the forkchoice block-validation guard now derive from it, and a clamp ensures the builder can only *lower* the cap, never exceed the `verifySignatures` hard limit.
- **spectest**: leanSpec submodule bumped to `3fb619d` (current main — includes #717 and #781); the runner consumes the new fixtures.

## Finalization fixes (what actually made it finalize)

Empirically root-caused on a 3-node local devnet (stuck at `justified=0`; ethlambda finalizes on the same box) via the unbuffered `consensus.log` + coverage metrics, then verified against leanSpec:

1. **leanSpec [PR #781](https://github.com/leanEthereum/leanSpec/pull/781) — `latest_justified` monotonicity.** `process_attestations` advanced the justified pointer unconditionally, so non-monotonic targets within one block could drag it backward and freeze finality. Fixed to advance only when `target.slot > latest_justified.slot` (the per-slot `justified_slots` bitfield still records every justified target). Exact match to #781.
2. **A builder-side time-forward guard in `getProposalAttestations`** (`target.slot ≤ source.slot → skip`). zeam applies block attestations to the LMD-GHOST tracker **eagerly on import** (unlike leanSpec, which defers block-vote head-weight). Without the guard, genesis-target self-votes — one distinct `AttestationData` per early slot — accumulate past the cap, fill the block budget, and (because they are applied to the tracker) pull the head back to genesis → stall + divergence. The guard drops exactly those provably-non-justifying votes. This is a deliberate, experiment-confirmed zeam deviation (leanSpec keeps the genesis self-votes because it defers their head-weight).

## Deviations from spec (intentional)

- Off-loop propose (single-threaded loop constraint).
- Aggregation deadline + partial aggregate (ethlambda; not in spec).
- Re-aggregation gated in-sync + capped (ethlambda) vs spec's run-for-every-block.
- The builder time-forward guard above (compensates for zeam's eager block-vote tracker application).

## Verification

- 3-node local devnet (ReleaseSafe) **finalizes in lockstep** at `MAX_ATTESTATIONS_DATA=8` (and at `=1`); head tracks the wall clock, no divergence.
- Component count is convergence-driven, not cap-driven: steady-state blocks merge **1 component** even at cap 8 (matches ethlambda).
- `zig build test`: **188 node tests pass**, plus the saturation/ingestion stress harness.


pkgs/node/src/chain.zig
```diff
@@ -397,6 +399,14 @@ pub const BeamChain = struct {
...
     aggregate_max_inflight: u32,
 
+    /// Dedicated Io.Threaded for the block-proposal worker.
+    /// `concurrent_limit = .limited(1)` enforces at most one in-flight
+    /// proposal task; a second submit returns `error.ConcurrencyUnavailable`.
+    propose_io: *std.Io.Threaded,
+
+    /// Long-lived group hosting submitted proposal tasks.
+    propose_group: std.Io.Group = .init,
+
```
anshalshukla: Let's remove this and move back to aggregation via thread_pool

pkgs/node/src/chain.zig
```diff
@@ -542,6 +552,18 @@ pub const BeamChain = struct {
...
+        // off the slot-driver loop. concurrent_limit = .limited(1) caps in-flight to one.
+        // (Aggregate work uses main's #907 counter + WaitGroup model, below.)
+        const prop_io = try allocator.create(std.Io.Threaded);
+        errdefer allocator.destroy(prop_io);
+        prop_io.* = std.Io.Threaded.init(allocator, .{
+            .async_limit = .nothing,
+            .concurrent_limit = .limited(1),
+        });
+        errdefer prop_io.deinit();
+
```
anshalshukla: rm

pkgs/node/src/node.zig
```diff
@@ -2632,44 +2633,46 @@ pub const BeamNode = struct {
...
-                // we also tick validator per interval in case it would
-                // need to sync its future duties when its an independent validator
-                var maybe_validator_output = validator.onInterval(interval) catch |e| blk: {
-                    self.logger.err("error ticking validator to time(intervals)={d} err={any} (continuing tick)", .{ interval, e });
-                    zeam_metrics.metrics.lean_node_interval_error_total.incr(.{ .site = "validator.onInterval" }) catch |me| self.logger.warn("metric incr failed: {any}", .{me});
-                    break :blk null;
-                };
+                if (interval_in_slot == 0) {
+                    self.chain.submitProposeOnInterval(self, interval);
+                } else {
```
anshalshukla: Why is this required?

pkgs/node/src/stress.zig
```diff
@@ -561,7 +561,7 @@ fn runStress(allocator: Allocator, cfg: StressConfig) !StressSummary {
             .name = spec_name,
             .fork_digest = fork_digest,
             .attestation_committee_count = 1,
-            .max_attestations_data = 16,
+            .max_attestations_data = 8,
```
anshalshukla, line 555: use params directly instead of hardcoding

pkgs/node/src/stress.zig
```diff
@@ -1068,7 +1068,7 @@ fn runStressSaturation(allocator: Allocator, cfg: SaturationConfig) !SaturationS
...
```
anshalshukla, line 1060: use params directly instead of hardcoding
I think there are more test case above where I missed this, if params cannot be included let's have a common variable to define it in a file to reduce errors while updating

pkgs/node/src/testing.zig
```diff
@@ -205,84 +205,4 @@ pub const NodeTestContext = struct {
     pub fn genesisConfig(self: *NodeTestContext) types.GenesisSpec {
         return self.genesis_config;
     }
-
-    pub fn signBlockWithValidatorKeys(
-        self: *const NodeTestContext,
-        allocator: Allocator,
-        block: *types.SignedBlock,
```
anshalshukla, line 214: why is this no more required? we will still need to test properly formed blocks right?

pkgs/spectest/src/runner/ssz_runner.zig
```diff
@@ -120,31 +120,9 @@ const TestSignedAttestation = struct {
...
-    signature: TestBlockSignatures,
-
-    pub fn deinit(self: *TestSignedBlock) void {
-        self.block.deinit();
-        self.signature.deinit();
-    }
-};
+// Devnet5 `SignedBlock` carries a single opaque `proof: ByteList512KiB` (no SIGBYTES-sized
+// fields), so one type SSZ-roundtrips both prod and test fixtures — no separate test-scheme type
+// is needed. The old TestBlockSignatures/TestSignedBlock types were removed with BlockSignatures.
```
anshalshukla: not required ai slop

rust/hashsig-glue/src/lib.rs
```diff
@@ -1,13 +1,9 @@
...
-// Upstream renamed `Scheme...` → `SIG...` in `lifetime_2_to_the_32` only
-// (leanSig main, pulled in via leanMultisig `5eba3b1`). The `lifetime_2_to_the_8`
-// test module below is unchanged and still exports `Scheme...`.
+// Production config (default)
 #[cfg(not(feature = "test-config"))]
 mod config {
     pub use leansig::signature::generalized_xmss::instantiations_aborting::lifetime_2_to_the_32::{
         PubKeyAbortingTargetSumLifetime32Dim46Base8 as XmssPublicKey,
-        SIGAbortingTargetSumLifetime32Dim46Base8 as XmssScheme,
+        SchemeAbortingTargetSumLifetime32Dim46Base8 as XmssScheme,
```
anshalshukla: @ch4r10t33r can you explain why you changed this earlier? I think this was part of thin lto optimization on rust bindings

.github/workflows/ci.yml
```diff
@@ -428,7 +432,12 @@ jobs:
 
     - name: Generate LeanSpec fixtures
       working-directory: leanSpec
-      run: uv run fill --clean --fork=Lstar -n auto
+      # Matches the submodule leanSpec's own `just fill-ci` body. Runs SERIALLY
+      # (no -n auto) on purpose: per leanSpec/Justfile, pytest-xdist workers race
+      # on the lean_multisig_py prover's process-global setup and corrupt proofs
+      # intermittently. Restore -n auto only once that is per-process safe upstream.
+      # (This is the matching ci.yml note leanSpec/Justfile points to.)
+      run: uv run --group test fill --clean --fork=Lstar
```
anshalshukla: i think in the latest commit -n auto has been restored please look into that and update if that is the case

rust/multisig-glue/src/lib.rs
```diff
@@ -70,355 +69,417 @@ pub extern "C" fn xmss_setup_verifier() -> i32 {
...
     message_hash_ptr: *const u8,
     slot: u32,
     log_inv_rate: usize,
-) -> *const AggregatedXMSS {
-    if message_hash_ptr.is_null() {
-        return std::ptr::null();
+    out_buf: *mut u8,
+    out_cap: usize,
+    out_written: *mut usize,
+) -> i32 {
```
anshalshukla, line 129: Is there a reason why we are taking it as input instead of returning it?

pkgs/types/src/block.zig
```diff
@@ -389,52 +393,26 @@ pub const BeamBlock = struct {
...
-        return utils.jsonToString(allocator, json_value);
-    }
-};
-
+/// Devnet5 block envelope: the block plus a single aggregated proof.
+///
+/// `proof` is the SSZ-encoded form of a `TypeTwoMultiSignature` (a ByteList512KiB whose
+/// inner bytes are the merged Type-2 proof) binding every body attestation plus the
+/// proposer's signature over the block root. Replaces devnet4's `BlockSignatures`
+/// (per-attestation proof list + raw proposer signature).
```
anshalshukla: rm

pkgs/types/src/block.zig
```diff
@@ -506,17 +584,13 @@ pub const AggregatedAttestationsResult = struct {
...
-    /// FFI, run each prove on `thread_pool` after a serial prep phase (issue
-    /// #907 — worker processes ~2 `att_data` per slot window).
     pub fn computeAggregatedSignatures(
         self: *Self,
         validators: *const Validators,
         signatures_map: *const SignaturesMap,
         new_payloads: ?*const AggregatedPayloadsMap,
         known_payloads: ?*const AggregatedPayloadsMap,
         slot_filter: ?[]const Slot,
-        thread_pool: *ThreadPool,
```
anshalshukla, line 582: why is thread_pool no longer needed? We should ideally compact all the type1 proofs parallely and then create type2 proof in one of the thread.

pkgs/node/src/chain.zig
GrapeBaBa: Done. Removed the dead `propose_io`/`propose_group` `Io.Threaded` infra entirely (field, init alloc, struct literal, deinit). It was allocated/torn-down but never submitted to: `submitProposeOnInterval` runs `proposeImpl` synchronously (0f778f94 libxev diagnostic) and proposal proving runs on the shared `thread_pool`. Also fixed the stale `proposeImpl` comments that referenced the removed field.

pkgs/node/src/chain.zig
GrapeBaBa: Removed — same dead `propose_io` allocation block; gone with the field above.

pkgs/node/src/chain.zig
GrapeBaBa: Done in 320de8b6: removed the dead `propose_io`/`propose_group`. They were allocated and torn down but never submitted to (the #863 offload was reverted to a synchronous `proposeImpl`), so it was pure dead code.

pkgs/node/src/chain.zig
GrapeBaBa: Done in 320de8b6 — same `propose_io` dead-code removal (init + field + deinit).

pkgs/node/src/node.zig
GrapeBaBa: It routes the proposer's interval-0 through the chain proposal entry point (`submitProposeOnInterval`→`proposeImpl`→`maybeDoProposal`). With the worker offload removed it's now functionally equivalent to `validator.onInterval(0)`, so it can be collapsed — happy to simplify if you prefer.

pkgs/node/src/stress.zig
GrapeBaBa: Done in 320de8b6 — now uses `params.MAX_ATTESTATIONS_DATA`.

pkgs/node/src/stress.zig
GrapeBaBa: Done in 320de8b6 — both stress configs now reference `params.MAX_ATTESTATIONS_DATA` (the single source of truth) instead of the literal.

pkgs/node/src/testing.zig
GrapeBaBa: The removed helper built the obsolete devnet4 `BlockSignatures`. In devnet5, `genMockChain` produces valid Type-2 block proofs directly via `buildType2BlockProof` (state-transition/src/mock.zig), so well-formed blocks are still exercised — no separate signing helper needed.

.github/workflows/ci.yml
GrapeBaBa: Against the pinned leanSpec (`ac5f259`), `fill-ci` runs serially, which matches this ci.yml. If a newer leanSpec commit restored `-n auto`, which commit are you seeing? I'll bump the submodule pin and ci.yml together.

rust/multisig-glue/src/lib.rs
GrapeBaBa: The caller aggregates into a fixed-cap `ByteList512KiB`. The out-buffer + 0/-1/-2 protocol avoids a cross-FFI heap allocation that would need a separate free FFI call (the old API returned a pointer requiring `xmss_free_aggregate_signature`), eliminating an ownership/leak class.

pkgs/types/src/block.zig
GrapeBaBa: Already removed — only the `InvalidBlockSignatures` error name remains; the struct itself is gone.

pkgs/types/src/block.zig
GrapeBaBa: The per-AttestationData committee compaction of Type-1 proofs already runs in parallel via the thread pool (`compactAttestations`, #915). The recursive STARK proves run serially on purpose — fanning them across the pool, each spawning its own internal Rayon, oversubscribes CPU and inflated devnet aggregator p50/p95 (the regression #925 fixes on main). `buildType2BlockProof` at this line does only the single final `merge_many` over the already-compacted Type-1s, so there's no `thread_pool` here by design — which matches what you're describing: compact in parallel, then build the Type-2 in one pass.

pkgs/node/src/testing.zig
anshalshukla: Just to confirm we are still signing the block via proposer key and wrapping it in type2 proofs right?

.github/workflows/ci.yml
anshalshukla: https://github.com/leanEthereum/leanSpec/pull/772/changes

rust/multisig-glue/src/lib.rs
anshalshukla: I think we should move key generation and signing to a similar format as well to avoid leaks and avoid any additional rust allocations. I'll ask zclawz to do a PR on the current main and see if it works else we can do that as a separate PR or include it in this itself as per your wish

.github/workflows/ci.yml
GrapeBaBa: You're right — leanSpec #772 set the prover's rayon thread count to 1 and re-enabled parallel prod fill, so `fill-ci` is now `-n auto --dist=worksteal`. Bumped the submodule (ac5f259→352b558) and updated this step + comment to match. Done in d3c3572f.

pkgs/node/src/node.zig
GrapeBaBa: Keeping it as-is for now — this is the original devnet5 interval-0 routing; not collapsing it.

anshalshukla: I would prefer it to simplify as the slot interval and selection of functions should happen at a single place to avoid and any confusions

rust/hashsig-glue/src/lib.rs
ch4r10t33r: The latest devnet4 commit on leanMultisig had changed this. I had to rename it when I bumped up the version.

pkgs/node/src/node.zig
GrapeBaBa: Simplified as you suggested — collapsed the interval-0 special case so every interval routes through `validator.onInterval` (the slot-interval→duty mapping now lives only there: proposal at 0, attestation at 1). Dropped the now-redundant `submitProposeOnInterval`/`proposeImpl`. Done in 18bb72fb.

pkgs/node/src/testing.zig
GrapeBaBa: Yes — `genMockChain` signs the block root with the proposer key (`key_manager.signBlockRoot`) and wraps it via `buildType2BlockProof` (per-attestation Type-1s + proposer singleton merged into the single Type-2). Confirmed still proposer-signed + Type-2.

pkgs/types/src/state.zig
```diff
@@ -478,18 +478,16 @@ pub const BeamState = struct {
...
             if (3 * target_justifications_count >= 2 * num_validators) {
-                // The block becomes justified
-                //
-                // The chain now considers this block part of its safe head.
-                // Only advance the checkpoint forward.
-                // Attestations within a block can resolve in any order, and
-                // an earlier target processed after a later one must not
-                // drag latest_justified backwards.
+                // Record the justified target in the per-slot bitfield regardless of order.
+                try utils.setSlotJustified(finalized_slot, &self.justified_slots, target_slot, true);
```
anshalshukla: This is not present in spec, why do need it? Also we should set it after updating the latest_justified like we were doing it earlier.

pkgs/types/src/aggregation.zig
```diff
@@ -114,33 +118,131 @@ pub const AggregatedSignatureProof = struct {
...
+        try xmss.verifyType1(public_keys, message_hash, @intCast(epoch), &self.proof);
+    }
+};
+
+/// Compatibility alias: the per-attestation aggregate proof was renamed
+/// `AggregatedSignatureProof` → `SingleMessageAggregate` (same shape: participants + proof bytes,
+/// same `aggregate()` signature). Aliased so the per-att_data aggregation code keeps
+/// compiling against the new type without a wholesale rename. Field rename `proof_data`→`proof`
+/// and the `ByteList512KiB` proof-bytes container are reconciled at the call sites.
+pub const AggregatedSignatureProof = SingleMessageAggregate;
```
anshalshukla: Lets remove the compatibility alias it makes it confusing down the line

rust/multisig-glue/src/lib.rs
```diff
@@ -20,424 +19,475 @@ pub struct Signature {
...
+            init_aggregation_bytecode();
+        })
+        .is_ok()
+    });
+    if *ready {
+        0
+    } else {
+        -1
+    }
+}
```
anshalshukla: We don't need this

rust/multisig-glue/src/lib.rs
```diff
@@ -20,424 +19,475 @@ pub struct Signature {
...
-/// `init_aggregation_bytecode()` panics when the compiled prover bytecode file is missing.
-/// A panicking first call leaves the `OnceLock` empty, so a later call can retry.
+/// `init_aggregation_bytecode()` may panic (e.g. when the compiled prover bytecode file is
+/// missing — ENOENT). A Rust panic through an `extern "C"` boundary is UB, so we wrap the
+/// init body in `catch_unwind` and cache the boolean result in a `OnceLock`. The caller (Zig)
+/// checks the return code and surfaces a hard error instead of crashing.
 #[no_mangle]
-pub extern "C" fn setup_xmss_aggregation() -> i32 {
-    static INIT: OnceLock<()> = OnceLock::new();
-    match std::panic::catch_unwind(|| {
```
anshalshukla: I had refactored it such that setup is called just once every time a node starts and not repeatedly checked before every aggregation/ verification to avoid frequent locking and multiple instantiations

pkgs/xmss/src/aggregation.zig
```diff
@@ -5,209 +5,241 @@ const zeam_metrics = @import("@zeam/metrics");
...
+// --- multisig-glue FFI (leanMultisig Type-1/Type-2 surface) ---
+//
+// All wrappers are pure Zig + extern calls and are safe to call from any thread, but each one
+// drives the leanMultisig zk prover/verifier — they are heavy and must not be called from the
+// libxev thread. Run them on the chain worker / Io.Threaded pool instead.
+
+/// Returns 0 on success, -1 if the prover bytecode is missing or init failed (never panics).
+extern fn xmss_setup_prover() callconv(.c) c_int;
+/// Returns 0 on success, -1 on failure.
+extern fn xmss_setup_verifier() callconv(.c) c_int;
```
anshalshukla: I think we should stick with earlier pattern and setup while starting node and error out if the setup fails

pkgs/state-transition/src/transition.zig
```diff
@@ -54,513 +54,127 @@ pub fn apply_raw_block(allocator: Allocator, state: *types.BeamState, block: *ty
...
-
-    // All per-task scratch (pubkey handle arrays, pubkey_wrappers when cache is absent)
-    // lives in this arena and is freed with one call after the parallel phase returns.
-    var scratch = std.heap.ArenaAllocator.init(allocator);
-    defer scratch.deinit();
-    const scratch_alloc = scratch.allocator();
+    // Per-component scratch (pubkey handle arrays, message bindings) lives in this arena.
+    var arena = std.heap.ArenaAllocator.init(allocator);
+    defer arena.deinit();
+    const ar = arena.allocator();
```
anshalshukla: nit: call it allocator

pkgs/node/src/node.zig
```diff
@@ -2361,7 +2398,16 @@ pub const BeamNode = struct {
...
+                // Use the DB-PERSISTED finalized slot as the index/walk boundary, NOT the in-memory
+                // forkChoice value. processFinalizationAdvancement updates in-memory finalization,
+                // then (later) commits the finalized slot index + persisted-slot in one batch, then
+                // prunes forkChoice. During that lag the in-memory finalized is ahead of the DB
+                // index: trusting it makes the finalized loop below skip slots whose index is not yet
+                // written, producing a response that starts past start_slot → the requester sees a
+                // fork mismatch and aborts to the slow blocks_by_root walk. The persisted value, by
+                // contrast, always matches what the index actually holds, and during the lag the
+                // still-unpruned forkChoice walk covers the remaining (lower) slots contiguously.
+                const finalized_slot = self.chain.db.loadLatestFinalizedSlot(database.DbDefaultNamespace) orelse 0;
```
anshalshukla, line 2410: what's the problem in using the memory one?

pkgs/node/src/node.zig
GrapeBaBa: Resolved on this branch: the `interval_in_slot == 0` special case was removed (d005bd07). node.zig now calls `validator.onInterval` unconditionally, and the off-loop proposer dispatch (`submitProposeOnInterval`) moved into `validator_client.onInterval`, so interval->duty selection lives in the validator layer.

pkgs/node/src/testing.zig
GrapeBaBa: Resolved: `signBlockWithValidatorKeys` was restored and ported to Type-2 (fb73de77). It lives at testing.zig:211-280 and two node.zig tests use it to re-sign mutated mock blocks, so properly-formed (proposer-signed Type-2) blocks are still exercised.

pkgs/types/src/state.zig
GrapeBaBa: `justified_slots` / `setSlotJustified` is pre-existing zeam infra, not added by this PR: the per-slot bitfield dates to devnet1 (#203/#257/#331) and the `setSlotJustified` helper to #490 (rebase justifications off finalized_slot). It backs `isSlotJustified` + the finalization scan that zeam fork-choice needs (it differs from the block-tree walk), so it stays. The ordering point is valid and done in 27a60afb: `latest_justified` is set before `setSlotJustified` now, keeping the strictly-ahead guard. The two are independent, so it is behaviorally neutral.

pkgs/types/src/aggregation.zig
GrapeBaBa: Done (7b6c93a3): dropped the `AggregatedSignatureProof` alias and renamed all uses to the spec name `SingleMessageAggregate` (type-identical).

rust/multisig-glue/src/lib.rs
GrapeBaBa: Done (2c374f96): removed `xmss_setup_verifier` and its OnceLock; collapsed to a single `xmss_setup_aggregation`.

rust/multisig-glue/src/lib.rs
GrapeBaBa: Done (2c374f96): restored the single-setup design from #936 - one `xmss_setup_aggregation` behind one OnceLock, run once at node start; dropped the per-op lazy re-checks.

pkgs/xmss/src/aggregation.zig
GrapeBaBa: Done (2c374f96): single `setupXmssAggregation()` at node start that hard-errors on failure (AggregationSetupFailed); removed the lazy `ensureProverReady` and the per-op setup calls.

pkgs/state-transition/src/transition.zig
GrapeBaBa: Done (90e37dcb): renamed `ar` -> `arena_allocator` (plain `allocator` would shadow the function parameter).

pkgs/node/src/node.zig
GrapeBaBa: The DB finalized-slot read is deliberate (705f5130, late-joiner sync): the index and the persisted latest-finalized slot are written in one batch AFTER in-memory finalization advances, so the in-memory value can be ahead of the committed index. Using it for the loop boundary while the body reads the index would skip not-yet-written slots, the peer would see a parent mismatch and abort to the slow blocks_by_root walk. Reading the slot from the same namespace keeps boundary and index consistent. Rationale is inline at node.zig:2401-2409.

anshalshukla (approved):

PR #919, opened 2026-05-24
node, chain: parallel aggregate submission for #907

ch4r10t33r:
## Summary

- Merge [#916](https://github.com/blockblaz/zeam/pull/916) parallel aggregate work: replace serial `std.Io.Threaded` (`concurrent_limit=1`) with bounded non-blocking submits onto the shared `ThreadPool` (`aggregate_max_inflight=4`).
- Check the in-flight cap **before** the expensive state `cloneAndRelease`, so saturated workers do not waste deep clones.
- Add `zeam_aggregate_skip_total{reason="spawn_failed"}` when ThreadPool enqueue fails (never fall back to inline ~10s FFI on the libxev thread).
- Clarify Prometheus HELP text: `lean_pq_sig_aggregated_signatures_building_time_seconds` measures per-`att_data` wrap time inside `computeAggregatedSignatures`, **not** the XMSS STARK worker; operators should use `zeam_aggregate_worker_duration_seconds` for end-to-end FFI cost ([#907](https://github.com/blockblaz/zeam/issues/907)).

## Context

On the ansible devnet, zeam aggregators were dropping ~50% of slot triggers as `in_flight` because the XMSS FFI takes ~11s while the slot interval is ~6s and only one aggregate worker could run at a time ([#907](https://github.com/blockblaz/zeam/issues/907)). Live logs on `zeam_8` with a #916-era image showed the skip rate improved to ~10%, but publish lag and `timely=none` merge timing remain separate follow-ups.

This branch is based on current `main` (includes #908 trivial-input pre-filter) and supersedes the open #916 branch for review/merge.

## Test plan

- [ ] Deploy to devnet aggregator (`zeam_8`) and confirm `zeam_aggregate_skip_total{reason="in_flight"}` rate drops vs pre-#916 baseline
- [ ] Confirm `zeam_aggregate_worker_duration_seconds` p50 stays ~10s for multi-sig inputs (parallelism helps throughput, not per-run FFI cost)
- [ ] Confirm `zeam_aggregator_publish_aggregations_total` continues incrementing on duty subnet under load


zclawz (approved):
Approved. I re-checked the current head after the force-push; the tree matches the version I reviewed, with no new diff from my prior pass.

PR #920, opened 2026-05-24
types, node: parallelize per-att_data aggregate FFI (#907)

ch4r10t33r:
## Summary

- Refactor `computeAggregatedSignatures` to prepare each `AttestationData` serially, then run independent recursive STARK FFI calls in parallel when the aggregator thread pool is available and more than one entry needs proving.
- Sort `att_data` keys deterministically before building output (stable block/aggregate ordering across runs).
- Wire the aggregator path in `forkchoice.zig` through `computeAggregatedSignaturesWithPool`.

Complements #919 (parallel aggregate submission / `in_flight` skips). That PR improves slot throughput; this one targets the ~11s per-FFI wall clock when a slot window contains multiple `att_data` entries (~2 per worker call per #907 finding 1).

## What this does not fix

- Single `att_data` with 7–8 gossip sigs still pays one full ~11s STARK (cost is input-shape independent in leanMultisig). Trivial 1-sig cases remain covered by #908 aggregator pre-filter.
- Does not address #863 event-loop backlog or peer aggregate merge timing (`timely=none`).

## Test plan

- [ ] Devnet: `zeam_aggregate_worker_duration_seconds` p50 drops when slot window has 2 `att_data` (expect ~1× STARK time instead of ~2×)
- [ ] Cross-check with #919 on aggregator image for combined slot skip + worker latency improvement

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -214,6 +214,7 @@ pub fn build(b: *Builder) !void {
     zeam_types.addImport("@zeam/utils", zeam_utils);
     zeam_types.addImport("@zeam/metrics", zeam_metrics);
     zeam_types.addImport("@zeam/xmss", zeam_xmss);
+    zeam_types.addImport("@zeam/thread-pool", zeam_thread_pool);
 
     // add zeam-types
     const zeam_configs = b.addModule("@zeam/configs", .{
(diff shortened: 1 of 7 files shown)


zclawz (approved):
Approved after review. The parallel per-att_data aggregation refactor looks sound to me: prep stays serial, FFI work is scoped through the shared ThreadPool, output order is deterministic, and the aggregator path is wired through the pool while keeping outer aggregate in-flight capped. Local validation is blocked by this workspace Zig/dependency mismatch, but CI lint/macOS build are green and remaining Linux/test jobs are still running with no failures at approval time.

PR #925, opened 2026-05-25
node, types, xmss: keep aggregate worker under 4s slot budget (#907)

### LINKED ISSUE #907: node, aggregation: PR #902 metrics reveal real FFI worker cost is ~30x the standard build histogram, and ~50% of slot triggers are skipped as in_flight

ch4r10t33r:
## Summary
- Coalesce in-flight aggregation triggers instead of skipping them; one catch-up run when the worker finishes (`zeam_aggregate_coalesced_total`)
- On slow workers, refresh the slot window from the slot clock so we do not prove stale `att_data` (cuts coalesce storms after #923)
- Inline single-FFI in `runAggregateAttDataPreps` when only one `att_data` in the slot needs proving (avoids thread-pool spawn on the common case)
- Aggregator startup calls `ensureProverReady()` once (Rust init + Zig `prover_ready` flag)

Follow-up to #923: worker p50 was ~0.65s on steady devnet, but ~22% of slot triggers were skipped as `in_flight`. Tail proves (3–10s) could still exceed the 4s slot budget. This PR keeps the #923 snapshot/prove metrics wins and replaces skips with coalesce + stale-slot guards.

Dropped after review: `finishSingleRawAggregatePrep` and multisig-glue single-raw early path — marginal prep savings for 0 children + 1 gossip sig where the general path is already equivalent.

## Test plan
- [x] `zig fmt --check .`
- [x] `cargo fmt --check` / `cargo clippy -D warnings`
- [x] `zig build`
- [x] `zig build stress-quick`
- [x] `zig build stress-quick-saturation`
- [ ] Deploy `0xpartha/zeam:local` to devnet aggregators; confirm worker p50 ~ #923 levels, coalesced rate low, `in_flight` skips at 0

Closes #907


rust/multisig-glue/src/lib.rs
```diff
@@ -122,6 +122,21 @@ pub unsafe extern "C" fn xmss_aggregate(
...
+        let pk_ptr = unsafe { *raw_pub_keys };
+        let sig_ptr = unsafe { *raw_signatures };
+        if pk_ptr.is_null() || sig_ptr.is_null() {
+            return std::ptr::null();
+        }
+        let raw_xmss = vec![((*pk_ptr).inner.clone(), (*sig_ptr).inner.clone())];
+        let (_pub_keys, agg_sig) =
+            rec_xmss_aggregate(&[], raw_xmss, message_hash, slot, log_inv_rate);
+        return Box::into_raw(Box::new(agg_sig));
+    }
```
anshalshukla: what does this add? We already have this logic in place where in case the num_raw > 0 we check the pub keys and raw signatures

pkgs/xmss/src/aggregation.zig
```diff
@@ -85,6 +85,12 @@ fn ensureProverReady() !void {
     prover_ready.store(true, .release);
 }
 
+/// Idempotent prover warm-up for aggregators. Call at startup (before the
+/// first slot trigger) so the first real aggregate does not pay cold-init tail.
+pub fn prewarmProver() !void {
+    try ensureProverReady();
+}
```
anshalshukla: why not call ensureProverReady itself? we are wrapping a function into a functino if we want a public fucntion and `ensureProverReady` is not already than make it public

pkgs/types/src/block.zig
```diff
@@ -757,6 +758,74 @@ const AggregateAttDataFfiArgs = struct {
...
+        for (gossip_pks) |*pk| pk.deinit();
+        allocator.free(gossip_pks);
+    }
+    sigmap_pks.* = .empty;
+    sigmap_vids.deinit(allocator);
+
+    const empty_children = try allocator.alloc(aggregation.AggregatedSignatureProof, 0);
+    const empty_child_slices = try allocator.alloc([]*const xmss.HashSigPublicKey, 0);
+    const empty_child_allocs = try allocator.alloc([]*const xmss.HashSigPublicKey, 0);
+    const empty_child_wrappers = try allocator.alloc(xmss.PublicKey, 0);
```
anshalshukla: aggregateSignatures forms a similar payload in case of num_childrens == 0, how does this help?

pkgs/cli/src/node.zig
```diff
@@ -556,6 +556,9 @@ pub const Node = struct {
             xmss.setupProver() catch |err| {
                 self.logger.warn("xmss prover setup failed: {any}; aggregation may be unavailable", .{err});
             };
+            xmss.prewarmProver() catch |err| {
+                self.logger.warn("xmss prover prewarm failed: {any}", .{err});
+            };
```
anshalshukla: doesn't prewarn and setup same underlying calls?

rust/multisig-glue/src/lib.rs
ch4r10t33r: The null checks overlap with the general path — agreed. This early block skips the child proof setup (empty children_pks / children_proofs vecs and the child pointer walks) for the common 1-raw, 0-children shape. Small win on the Rust side; the bigger prep savings are in block.zig. Keeping it unless you prefer one unified path here.

pkgs/xmss/src/aggregation.zig
ch4r10t33r: Agreed — fixed in fe384fff. Dropped prewarmProver and exported ensureProverReady instead. Aggregator startup now calls that once.

pkgs/types/src/block.zig
ch4r10t33r: Rust still has to prove once for a valid wire format (#907 finding 4). This path saves Zig work before FFI: it skips child-selection loops in prepareAggregateAttData and runs a single FFI inline in runAggregateAttDataPreps when only one att_data needs proving. The empty child slices are the same shape the general path would end up with — we just avoid building them via the full child-merge path first.

pkgs/cli/src/node.zig
ch4r10t33r: Yes — both hit xmss_setup_prover on the Rust side. The difference was Zig-side: setupProver alone did not set prover_ready, so the first real aggregate could still re-enter setup. Fixed in fe384fff: one ensureProverReady() call at startup (sets the flag + init). Removed the duplicate setupProver + prewarmProver pair.

anshalshukla: What do you think if we skip the whole setupProver path before every aggregate and call it explicity when the node runs, I think the prep saving will be little but it will look simpler and we will def avoid some flag locks and checks before every aggregate/ verify run

pkgs/types/src/block.zig
anshalshukla: but why do we want to skip child selection loops? we might find a children that includes this signature in that case we should directly use it, or we can have a better aggregate with the children and gossip participant bit set

ch4r10t33r: We do not skip child selection — extendProofsGreedily runs first on new_payloads and known_payloads. This path only runs when that pass found zero children and exactly one gossip sig is left (validators already covered by a child are dropped at line 879). If a child covers the sig we either take the child-only done path or the general merge path. The shortcut is just less prep before the same single-raw FFI call.

pkgs/cli/src/node.zig
ch4r10t33r: Makes sense — init once at node startup and drop the per-aggregate ensureProverReady check would be cleaner. Can we do that as a follow-up after we verify this branch on devnet4? I want to confirm worker p50 and coalesce behaviour with the current fix before touching the hot path again.

anshalshukla: yeah I can create that separately

pkgs/types/src/block.zig
anshalshukla: > Rust still has to prove once for a valid wire format (https://github.com/blockblaz/zeam/issues/907 finding 4). This path saves Zig work before FFI: it skips child-selection loops in prepareAggregateAttData and runs a single FFI inline in runAggregateAttDataPreps when only one att_data needs proving. The empty child slices are the same shape the general path would end up with — we just avoid building them via the full child-merge path first.

I don't see then how this is helping, you can validate this in your next run but I don't see this will help much and makes a code little complex. I'm in favor of skipping this unless there is a huge performance improvement which i doubt, we can create an issue as well for this if you want to ship the PR in the current form itself

ch4r10t33r: @zclawz can you create an issue to track this? Let's address this issue/remove this code if there is no major performance improvement.

ch4r10t33r: Agreed on finishSingleRawAggregatePrep and the multisig-glue early path — dropped both in the latest commit. For 0 children + 1 gossip sig the general prep path is effectively the same, so that shortcut was not buying much.

Keeping the inline single-FFI in runAggregateAttDataPreps. That is a different thing: when the whole slot has only one att_data that needs proving, we run the FFI on the caller thread instead of spawning through the thread pool. On devnet most slots hit exactly one prove — avoiding pool setup/wakeup/sync on every trigger is real overhead next to a ~0.6s prove, and the change is ~10 lines with no duplicate prep logic. Happy to file an issue for the single-raw prep ideas if we want to revisit with prep-phase metrics later.

anshalshukla (dismissed):

pkgs/types/src/block.zig
zclawz: Created issue #928 to track this: https://github.com/blockblaz/zeam/issues/928

anshalshukla (dismissed):

zclawz (approved):
LGTM. Formally approving on GitHub as requested.

PR #927, opened 2026-05-25
node: refresh peer status when synced head lags wall clock

zclawz:
## Summary

Fixes #926 by making sync recovery independent of finalization-only sync status.

On early/devnet networks where peers keep `finalized_slot=0`, `getSyncStatus()` can correctly remain `.synced` even after gossip ingress stalls and the local head stops advancing. Previously the periodic status refresh was skipped in `.synced`, so no new peer status responses arrived to trigger the existing status-driven `blocks_by_root` / `blocks_by_range` catch-up path.

This PR:

- adds a canonical `wallGapSlots` helper reused by sync-gap and periodic-refresh policy
- keeps the periodic peer-status refresh decision as pure/tested logic in `blocks_by_range_sync.zig`
- adds `SYNC_STATUS_WALL_HEAD_LAG_THRESHOLD_SLOTS = 4`
- refreshes peer status while `.synced` when `wall_slot - our_head_slot >= 4`
- keeps `.no_peers` quiet and preserves the existing `.fc_initing` / `.behind_peers` refresh behavior

The result is that a node whose gossip ingress goes quiet will continue probing peers every `SYNC_STATUS_REFRESH_INTERVAL_SLOTS`; once a peer reports a higher head, the existing `.synced` status-response catch-up logic fetches the missing blocks over RPC.

## Tradeoff

The simpler fix would refresh status in every `.synced` interval too:

```zig
.fc_initing, .behind_peers, .synced => self.refreshSyncFromPeers(),
.no_peers => {},
```

This PR keeps the thresholded path to avoid steady-state status RPC fanout every 8 slots when the node is healthy. The cost is bounded recovery latency: lag must reach 4 wall-clock slots and then hit the 8-slot refresh cadence, so the first recovery probe can take just under 12 slots (~48s at 4s/slot) after the node stops advancing.

## Validation

- `zig fmt pkgs/node/src/constants.zig pkgs/node/src/blocks_by_range_sync.zig pkgs/node/src/node.zig`
- `git diff --check`
- Attempted `zig build test --summary all`, but the local runner has Zig 0.15.2 while `build.zig.zon` requires 0.16.0; dependency build scripts fail on removed/changed Build API fields before project tests compile.

```text
build.zig.zon: .minimum_zig_version = "0.16.0"
zig version: 0.15.2
error: no field named 'io' in struct 'Build.Graph'
```

### DIFF
--- a/pkgs/node/src/blocks_by_range_sync.zig
+++ b/pkgs/node/src/blocks_by_range_sync.zig
@@ -27,29 +27,38 @@ pub fn rangesOverlap(
     return a_start < b_end and b_start < a_end;
 }
 
-pub fn wallGapSlots(our_head_slot: types.Slot, wall_slot: types.Slot) u64 {
-    if (wall_slot <= our_head_slot) return 0;
-    return wall_slot - our_head_slot;
-}
-
 pub fn cappedSyncGapSlots(peer_head_slot: types.Slot, our_head_slot: types.Slot, wall_slot: types.Slot) u64 {
     if (peer_head_slot <= our_head_slot) return 0;
     const peer_gap: u64 = peer_head_slot - our_head_slot;
-    return @min(peer_gap, wallGapSlots(our_head_slot, wall_slot));
+    const wall_gap: u64 = if (wall_slot > our_head_slot) wall_slot - our_head_slot else 0;
+    return @min(peer_gap, wall_gap);
 }
 
-pub const PeriodicRefreshDecision = struct {
+pub const PeerStatusRefreshDecision = struct {
     refresh: bool,
     wall_head_lag_slots: u64,
 };
 
-pub fn periodicRefreshPeerStatusDecision(
+/// Pure periodic peer-status refresh policy.
+///
+/// Callers supply the current interval/slot cadence plus head snapshots; this helper
+/// owns the decision about whether a status refresh is useful. Wall-clock head lag is
+/// computed through `cappedSyncGapSlots` by treating `wall_slot` as the peer head,
+/// keeping all slot-gap arithmetic on one path.
+pub fn shouldRefreshPeerStatus(
     sync_status: anytype,
+    interval_in_slot: usize,
+    slot: types.Slot,
     our_head_slot: types.Slot,
     wall_slot: types.Slot,
+    refresh_interval_slots: u64,
     wall_head_lag_threshold_slots: u64,
-) PeriodicRefreshDecision {
-    const wall_head_lag_slots = wallGapSlots(our_head_slot, wall_slot);
+) PeerStatusRefreshDecision {
+    const wall_head_lag_slots = cappedSyncGapSlots(wall_slot, our_head_slot, wall_slot);
+    if (interval_in_slot != 0 or slot % refresh_interval_slots != 0) {
+        return .{ .refresh = false, .wall_head_lag_slots = wall_head_lag_slots };
+    }
+
     const refresh = switch (sync_status) {
         .fc_initing, .behind_peers => true,
         .synced => wall_head_lag_slots >= wall_head_lag_threshold_slots,
@@ -199,7 +208,7 @@ test "cappedSyncGapSlots limits range catch-up to wall-clock head" {
     try std.testing.expectEqual(@as(u64, 0), cappedSyncGapSlots(200, 150, 100));
 }
 
-test "periodicRefreshPeerStatusDecision handles sync state and wall lag" {
+test "shouldRefreshPeerStatus handles cadence sync state and wall lag" {
     const TestSyncStatus = union(enum) {
         synced,
         no_peers,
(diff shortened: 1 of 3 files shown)


pkgs/node/src/blocks_by_range_sync.zig
```diff
@@ -34,6 +34,15 @@ pub fn cappedSyncGapSlots(peer_head_slot: types.Slot, our_head_slot: types.Slot,
     return @min(peer_gap, wall_gap);
 }
 
+pub fn wallHeadLagSlots(our_head_slot: types.Slot, wall_slot: types.Slot) u64 {
```
ch4r10t33r: `wallHeadLagSlots` duplicates the wall-gap math already embedded in `cappedSyncGapSlots` (`wall_gap = if (wall_slot > our_head_slot) wall_slot - our_head_slot else 0`). Extract a single canonical helper, e.g. `wallGapSlots(our_head_slot, wall_slot)`, and have both `cappedSyncGapSlots` and the refresh policy call it. Otherwise we are one PR away from three copies of the same invariant.

pkgs/node/src/blocks_by_range_sync.zig
```diff
@@ -34,6 +34,15 @@ pub fn cappedSyncGapSlots(peer_head_slot: types.Slot, our_head_slot: types.Slot,
...
+    if (wall_slot <= our_head_slot) return 0;
+    return wall_slot - our_head_slot;
+}
+
+pub fn shouldRefreshStatusForWallHeadLag(our_head_slot: types.Slot, wall_slot: types.Slot, threshold_slots: u64) bool {
```
ch4r10t33r: This is a one-liner identity wrapper around `>= threshold`. It does not encode a distinct concept — just comparison syntax with a function name. Either inline at the policy call site or fold into a higher-level policy function (see `node.zig` comment). Thin pass-through helpers like this tend to accumulate.

pkgs/node/src/node.zig
```diff
@@ -2681,14 +2681,32 @@ pub const BeamNode = struct {
...
-            // was skipped (now fixed, but existing connections need a re-probe).
+            // Periodically re-send status to connected peers when sync may need
+            // recovery. Finalization-only sync state can report `synced` on early
+            // devnets while finalized_slot stays at zero; if gossip ingress stalls,
+            // the head then falls behind wall-clock slots but no new status response
+            // arrives to trigger RPC catch-up. Keep probing peers when the local
+            // head is materially behind wall clock so the `.synced` status-response
+            // arm can observe a peer head and start blocks_by_root/range recovery.
             if (interval_in_slot == 0 and slot % constants.SYNC_STATUS_REFRESH_INTERVAL_SLOTS == 0) {
                 switch (self.chain.getSyncStatus()) {
```
ch4r10t33r: This works, but it makes an already busy 4.2k-line file more tangled. The tick handler now owns sync policy: `.fc_initing`/`.behind_peers` → refresh, `.synced` + wall lag → refresh, `.no_peers` → quiet.

**Code-judo move:** push the whole decision into `blocks_by_range_sync.zig` (where `shouldCatchUpFromPeerStatus`, `cappedSyncGapSlots`, and `syncEndDecision` already live):

```zig
pub fn shouldPeriodicRefreshPeerStatus(
    sync_status: chain.SyncStatus,
    our_head_slot: types.Slot,
    wall_slot: types.Slot,
) bool { ... }
```

Then `node.zig` becomes a single `if (shouldPeriodicRefreshPeerStatus(...)) { self.refreshSyncFromPeers(); }` — deletes the growing `switch`, keeps #926 logic testable without spinning up `BeamNode`, and matches the module header ("pure gap/decision logic lives here").

pkgs/node/src/node.zig
```diff
@@ -2681,14 +2681,32 @@ pub const BeamNode = struct {
...
                     .fc_initing, .behind_peers => self.refreshSyncFromPeers(),
-                    .synced, .no_peers => {},
+                    .synced => {
```
ch4r10t33r: Before landing `SYNC_STATUS_WALL_HEAD_LAG_THRESHOLD_SLOTS` + two helpers + a new `.synced` arm, worth explicitly weighing the simpler one-line fix:

```zig
.fc_initing, .behind_peers, .synced => self.refreshSyncFromPeers(),
.no_peers => {},
```

Pros: no new concepts, no wall-lag threshold to tune. Cons: status RPC every 8 slots even when healthy (~32s at 4s/slot). If RPC load is the concern, the threshold approach is fine — but please state that tradeoff in the PR description.

If threshold stays, document worst-case recovery latency: lag must reach 4 slots **and** the tick must hit an 8-slot boundary → up to ~12 slots (~48s) before the first probe.

pkgs/node/src/blocks_by_range_sync.zig
```diff
@@ -175,6 +184,14 @@ test "cappedSyncGapSlots limits range catch-up to wall-clock head" {
     try std.testing.expectEqual(@as(u64, 0), cappedSyncGapSlots(200, 150, 100));
 }
 
+test "wall head lag triggers status refresh even when sync status is synced" {
```
ch4r10t33r: Test name says "even when sync status is synced" but only covers `wallHeadLagSlots` / `shouldRefreshStatusForWallHeadLag` arithmetic — no sync-status integration. Rename to something like `"wallHeadLagSlots and refresh threshold"`, or add a table-driven test for `shouldPeriodicRefreshPeerStatus` once that policy function exists.

pkgs/node/src/node.zig
```diff
@@ -2681,14 +2681,32 @@ pub const BeamNode = struct {
...
+                        const our_head_slot = self.chain.forkChoice.getHead().slot;
+                        const wall_slot = self.clock.wallSlotNow();
+                        if (blocks_by_range_sync.shouldRefreshStatusForWallHeadLag(
+                            our_head_slot,
+                            wall_slot,
+                            constants.SYNC_STATUS_WALL_HEAD_LAG_THRESHOLD_SLOTS,
+                        )) {
+                            self.logger.info(
+                                "head is {d} wall-clock slots behind while synced; refreshing peer status for catch-up",
+                                .{blocks_by_range_sync.wallHeadLagSlots(our_head_slot, wall_slot)},
```
ch4r10t33r: `shouldRefreshStatusForWallHeadLag(...)` already computes lag internally; the log calls `wallHeadLagSlots(...)` again. Compute lag once, or have the policy helper return it for logging.

ch4r10t33r (changes requested):
## Thermo-nuclear code quality review

The fix targets the right failure mode (#926): `.synced` suppresses periodic status refresh, so gossip stalls never produce status responses and the existing `.synced` catch-up path in `handleReqRespResponse` never fires. Behavior is plausible and aligned with the architecture in `chain.zig` (head-only gaps intentionally live outside `getSyncStatus`).

However, the implementation adds branching and near-duplicate logic in the wrong layer instead of consolidating sync policy. There is a clear code-judo move that preserves behavior while deleting most of the new surface area.

**Verdict: request changes.** Minimum bar: (1) deduplicate wall-gap computation with `cappedSyncGapSlots`, (2) collapse periodic refresh policy into one testable function in `blocks_by_range_sync.zig`, (3) update the stale `SYNC_STATUS_REFRESH_INTERVAL_SLOTS` comment (still says "when not synced"), (4) CI green.

Longer term (not necessarily this PR): we now have three independent stall-recovery triggers — `SlotDriverWatchdog` → `sync_refresh_pending` (#863), periodic refresh for `.fc_initing`/`.behind_peers`, and periodic refresh for `.synced` + wall lag (#926). A single "sync health / probe peers" policy would be easier to reason about.

ch4r10t33r (dismissed):
LGTM

ch4r10t33r (approved):
LGTM

PR #929, opened 2026-05-25
node: skip redundant work on locally produced aggregate publish

ch4r10t33r:
## Summary

Add a fast publish path for worker-committed aggregates. After `commitOneAggregateResult` stores the proof, publish only updates attestation trackers and gossips — skipping the redundant XMSS re-verify and second fork-choice store that `onGossipAggregatedAttestation` performs.

**Worker publish entry points (fast path):**
- `publishOneAggregate` — per-job callback (#925)
- `publishProducedAggregations` — batch fallback

Both route through `publishCommittedAggregation` → `publishLocalProducedAggregation`.

**Shared helpers extracted:**
- `publishAggregationGossip` — single gossip encode/publish/log path
- `recordAggregatorPublishMetric` — per-subnet counter in one place
- `applyAggregatedAttestationTrackers` — tracker updates; gossip ingress passes precomputed validator indices (one allocation)

**Unchanged / full-ingest path:**
- `publishAggregation` — still validate, verify, store, then gossip (validator duty output and other proofs not yet committed by the worker)
- `onGossipAggregatedAttestation` — same semantics for inbound gossip and chain-worker dispatch
- `storeAggregatedPayload` — **clone-on-insert**; callers that deinit the surrounding `SignedAggregatedAttestation` after store remain safe (see review note below)

## Why

The aggregate worker already deep-clones each proof into `latest_new_aggregated_payloads` in `commitOneAggregateResult`. Routing worker publish through `onGossipAggregatedAttestation` re-verified the proof and stored a duplicate — wasted CPU and heap on every locally produced aggregate.

## Review feedback addressed

Initial diff moved proof ownership into `storeAggregatedPayload`, which broke borrowed callers (gossip handler `defer message.deinit()`, chain-worker dispatch, `publishAggregation`) — double-free / use-after-move. Reverted to clone-on-insert; the performance win is localized to the worker publish fast path, not ingress store.

## Test plan

- [x] `zig fmt --check .`
- [ ] CI: `zig build test` (in progress on latest push)
- [ ] Devnet aggregator: confirm `lean_committee_signatures_aggregation_time_seconds` / worker duration improves vs baseline
- [ ] Verify gossip ingress still accepts remote aggregates


zclawz (changes requested):
I can't approve this as-is.

The new ownership contract for `ForkChoice.storeAggregatedPayload` appears to break all non-worker-owned aggregation ingestion paths. The function now takes ownership of `proof`, but callers still pass a proof that is owned by a surrounding `SignedAggregatedAttestation` which is deinitialized later:

- inbound gossip: `ethlibp2p.handleGossip` does `defer message.deinit()`. `chain.onGossipAggregatedAttestation(signed_aggregation)` now stores/moves `signed_aggregation.proof`; when control returns, `message.deinit()` frees the same proof again.
- chain-worker path: `dispatch` does `defer m.deinit()`. The thunk calls `onGossipAggregatedAttestation(agg)`, which moves `agg.proof`; `m.deinit()` then frees it again.
- worker-disabled inline path has the same issue through the borrowed `signed_aggregation` from the gossip message.
- `BeamNode.publishAggregation` also calls `onGossipAggregatedAttestation` and then publishes the same `signed_aggregation`, so after the move the gossip serialization is using a consumed proof.

The fast local-worker path itself is a good direction, but the ownership change needs to be localized: either keep `storeAggregatedPayload` cloning by default and add a separate `storeAggregatedPayloadOwned`/fast-path API for worker-produced proofs, or make every caller that passes a borrowed `SignedAggregatedAttestation` clone/transfer intentionally and suppress the later deinit. As written this is a double-free / use-after-move risk on normal gossip aggregation handling, so I’m requesting changes.

zclawz (approved):
Re-reviewed after the ownership fix. The previously flagged issue is addressed: `ForkChoice.storeAggregatedPayload` once again clones the proof and explicitly documents that callers retain ownership, so gossip/chain-worker message deinit remains safe. The local produced-aggregate fast path now avoids re-verification/re-store while preserving tracker updates, gossip publish, per-subnet publish metrics, and single deinit ownership for worker-produced aggregates.

Validation performed locally:
- `git diff --check origin/main...pr-929-review`
- `zig fmt --check pkgs/metrics/src/lib.zig pkgs/node/src/chain.zig pkgs/node/src/forkchoice.zig pkgs/node/src/node.zig`
- `zig build test --summary all` attempted but blocked by the workspace Zig 0.15.2 vs repo/dependency 0.16.0 expectation (`Build.Graph.io`).

GitHub Actions are still in progress on commit `137c70a`, so this approval is code-review approval, not a completed-CI signal.

PR #933, opened 2026-05-25
node, types: fix aggregator worker segfault on aggregate commit

ch4r10t33r:
## Summary

- Fix double-free in `computeSingleAggregatedSignature` when the single-child passthrough (`.done`) path returns: set `prep.outcome = .skip` before returning, matching the existing `.ffi` path guard.
- Fix use-after-corruption in `commitOneAggregateResult`: `ssz.serialize` corrupts the source proof in memory (documented in `sszCloneAndGetBytes`). Use one serialize pass to store the proof and deserialize a separate copy for the publish path, instead of returning the corrupted original.

These match the SIGSEGV stack seen on devnet-4 `zeam_8` (exit 139) in the aggregate worker at `sszClone` / `commitOneAggregateResult`.

## Test plan

- [x] Added regression test: `computeSingleAggregatedSignature: single-child passthrough survives prep deinit`
- [ ] `zig build test --summary all`
- [ ] Deploy to devnet-4 aggregator and confirm `zeam_8` no longer crash-loops after publishing aggregations


zclawz (dismissed):
All three concerns from my initial review are resolved:

- ✅ Fix 1 (`.done` double-free): correct from the start.
- ✅ Fix 2 (SSZ corruption): `sszSerializeAndGetBytes` is now the right primitive — `signature_live = false` is set at the exact point after serialize succeeds/corrupts the source, before any fallible deserialize step. The helper is well-documented, has a roundtrip test, and `sszCloneAndGetBytes` is updated to delegate through it.
- ✅ Fix 3 (RPC mutex): handler lifetime documented and verified — `deinit()` only frees `peer_id`; `handler.ptr` references the node.

Approved.

pkgs/node/src/forkchoice.zig
```diff
@@ -2187,20 +2190,41 @@ pub const ForkChoice = struct {
...
-        try self.buildAggregateSourceAttribution(att_data, result.signature, &source_payload_bits, &source_gossip_bits);
+        var owned_signature = signature;
+        var signature_live = true;
+        errdefer if (signature_live) owned_signature.deinit();
+        // ssz.serialize corrupts the source value; serialize once, deinit the
+        // worker proof, then deserialize separate copies for store and publish
+        // so no fallible step runs while signature_live is true.
+        var stored_proof: types.AggregatedSignatureProof = undefined;
+        var stored_proof_owned = true;
+        errdefer if (stored_proof_owned) stored_proof.deinit();
```
anshalshukla, line 2201: this will free uninitialized memory if sszSerializeAndGetBytes or deserialize errors

ch4r10t33r: Good catch — stored_proof_owned was true while stored_proof was still undefined, so the errdefer could run deinit on garbage if serialize or deserialize failed.

Fixed in 12b7693f: start with stored_proof_owned = false and only set it true after deserialize succeeds.

anshalshukla (approved):

PR #935, opened 2026-05-26
hashsig-glue: replace Box allocs with Zig-owned placement-init pattern

zclawz:
Addresses the review comment https://github.com/blockblaz/zeam/pull/918#discussion_r3301879001 (ref PR #918) requesting that key-generation and signing be refactored to avoid additional Rust heap allocations, consistent with the caller-supplies-buffer pattern already used in multisig-glue.

## Problem

`hashsig-glue` was Box-allocating every `KeyPair`, `Signature`, and `PublicKey` on the Rust heap and handing raw pointers to Zig. Zig then had to call back into Rust (`hashsig_keypair_free` etc.) to free them. This created a split-ownership model where allocation and deallocation crossed the FFI boundary in opposite directions, complicating lifetime tracking and adding unnecessary alloc/free pairs.

## Solution: placement-init pattern

Rust exports layout queries so Zig can pre-allocate correctly-sized storage, then passes that storage to Rust for in-place initialisation. Cleanup runs Rust Drop in-place, then Zig frees its own buffer. The Rust heap is never involved for the outer struct.

### `rust/hashsig-glue/src/lib.rs`

**New layout queries:**
- `hashsig_sizeof_keypair / _signature / _public_key` — Zig queries at runtime to know how much to allocate
- `hashsig_alignof_keypair / _signature / _public_key` — required alignment (C `malloc` always satisfies this for `#[repr(C)]` types)

**New placement-init (replaces Box-returning functions):**
- `hashsig_keypair_generate_into(out: *mut KeyPair, ...)` — uses `std::ptr::write`, returns `i32` (0 / -1)
- `hashsig_keypair_from_ssz_into`, `hashsig_sign_into`, `hashsig_signature_from_ssz_into`, `hashsig_public_key_from_ssz_into` — same pattern

**New drop-in-place (replaces `Box::from_raw` frees):**
- `hashsig_keypair_deinit` / `hashsig_signature_deinit` / `hashsig_public_key_deinit` — call `std::ptr::drop_in_place`, do **not** free the buffer (caller owns it)

**Removed** (all were Box-allocating):
`hashsig_keypair_generate`, `hashsig_keypair_from_ssz`, `hashsig_keypair_free`, `hashsig_sign`, `hashsig_signature_free`, `hashsig_signature_from_ssz`, `hashsig_public_key_from_ssz`, `hashsig_public_key_free`

### `pkgs/xmss/src/hashsig.zig`

- `KeyPair`, `Signature`, `PublicKey` wrappers now own their storage via `cAlloc`/`cFree` (thin wrappers over `std.c.malloc`/`std.c.free`)
- C `malloc` always returns memory aligned to `max_align_t` (>=8 bytes on LP64), satisfying the alignment of any Rust `#[repr(C)]` type used here
- **Public API is unchanged** — same `generate`, `fromSsz`, `sign`, `verify`, `toBytes`, `deinit` methods; no changes needed in `aggregation.zig` or any other caller
- `PublicKeyCache` slot semantics unchanged; `deinit` correctly reconstructs `_buf` + `handle` from the stored integer pointer

## What this does NOT eliminate

- Inner leansig `Vec<u8>` data inside the XMSS key/signature types — these are owned by the leansig library and freed correctly when `drop_in_place` runs Drop
- The intermediate `Vec<u8>` inside the `to_bytes()` serialisation helpers — a leansig API constraint; short-lived temporaries that never escape the FFI call

## Verification

- `cargo check -p hashsig-glue` clean
- `zig ast-check` on `hashsig.zig`, `aggregation.zig`, `lib.zig` all pass
- No remaining callers of any removed Box-based FFI symbol in the Zig tree

### DIFF
--- a/pkgs/xmss/src/hashsig.zig
+++ b/pkgs/xmss/src/hashsig.zig
@@ -3,50 +3,103 @@ const Allocator = std.mem.Allocator;
 
 pub const aggregate = @import("aggregation.zig");
 
-/// Opaque pointer to the Rust KeyPair struct
+// Opaque Zig types that map to the corresponding Rust structs.
+// Zig never looks inside these; sizes come from the hashsig_sizeof_* queries.
 pub const HashSigKeyPair = opaque {};
-
-/// Opaque pointer to the Rust Signature struct
 pub const HashSigSignature = opaque {};
-
-/// Opaque pointer to the Rust PublicKey struct
 pub const HashSigPublicKey = opaque {};
-
-/// Opaque pointer to the Rust PrivateKey struct
 pub const HashSigPrivateKey = opaque {};
 
-/// Generate a new key pair
-extern fn hashsig_keypair_generate(
+// ─── Layout queries ────────────────────────────────────────────────────────────
+// Rust exports these so Zig can pre-allocate the right amount of space.
+// Call once at startup (or comptime-cache them); values are stable for the
+// lifetime of a given Rust build.
+
+extern fn hashsig_sizeof_keypair() callconv(.c) usize;
+extern fn hashsig_alignof_keypair() callconv(.c) usize;
+extern fn hashsig_sizeof_signature() callconv(.c) usize;
+extern fn hashsig_alignof_signature() callconv(.c) usize;
+extern fn hashsig_sizeof_public_key() callconv(.c) usize;
+extern fn hashsig_alignof_public_key() callconv(.c) usize;
+
+// ─── Placement-init (no Rust Box allocation) ──────────────────────────────────
+// Each `_into` function writes a fully-initialised struct into caller-supplied
+// storage (allocated by Zig/C malloc, which always satisfies Rust's alignment).
+// The matching `_deinit` runs Rust Drop in-place WITHOUT freeing the buffer;
+// the caller owns the buffer and must free it afterwards.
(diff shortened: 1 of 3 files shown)


zclawz:
**CI fix pushed** — commit `4aced8e`

**Root cause:** `pkgs/xmss/src/lib.zig` has its own copy of `PublicKeyCache` (separate from the one in `hashsig.zig`). When `PublicKey` gained the `_buf` field in this PR, `hashsig.zig`'s `deinit` was correctly updated to supply both fields:
```zig
const raw: [*]u8 = @ptrFromInt(ptr_int);
var pk = PublicKey{ ._buf = raw, .handle = @ptrCast(raw) };
```
But `lib.zig`'s `deinit` was left with the old single-field init:
```zig
var pk = PublicKey{ .handle = @ptrFromInt(ptr_int) }; // missing _buf
```
This caused `error: missing struct field: _buf` on all three macOS checks (build / test / Dummy prove).

**Fix:** port the same corrected pattern to `lib.zig`'s `deinit`. Both `_buf` and `handle` point to the same C-heap allocation start, so the stored atomic `usize` (which is `@intFromPtr(handle)`) is sufficient to reconstruct both fields.

🤖 zclawz

PR #936, opened 2026-05-26
rm multiple setup/ verify call sites

anshalshukla:
removes multiple locks from the hot path, performs single aggregation setup at the startup. aggregation twidles are required even in case of non aggregator nodes from compaction during block production

node returns an error it setupXmssAggregation fails


rust/multisig-glue/src/lib.rs
```diff
@@ -18,55 +18,19 @@ pub struct Signature {
...
-            init_aggregation_bytecode();
-        })
-        .is_ok()
-    });
-    if *ready {
-        0
-    } else {
-        -1
+pub extern "C" fn setup_xmss_aggregation() -> i32 {
+    match std::panic::catch_unwind(|| {
```
ch4r10t33r, line 35: This dropped the `OnceLock` guards we had on `xmss_setup_prover` / `xmss_setup_verifier`. Every call now reruns `init_aggregation_bytecode()` and the DFT twiddle precompute.

Zig is supposed to call this once, but if it gets called twice (test harness, refactor, dual entry point) we can double-init global state and crash. Please keep a single `OnceLock<bool>` here like before — same `catch_unwind`, but init runs at most once.

pkgs/xmss/src/aggregation.zig
```diff
@@ -98,17 +84,11 @@ pub fn setRayonThreads(num_threads: usize) void {
...
-    if (xmss_setup_prover() != 0) return error.ProverSetupFailed;
-}
-
-/// Initialize the XMSS verifier (idempotent — only runs once).
-pub fn setupVerifier() error{VerifierSetupFailed}!void {
-    if (xmss_setup_verifier() != 0) return error.VerifierSetupFailed;
+/// Initialize XMSS aggregation (both prove and verify state). Must be called
+/// exactly once at node startup, before any aggregation or verification work
+/// begins.
+pub fn setupXmssAggregation() error{XmssAggregationSetupFailed}!void {
```
ch4r10t33r, line 90: The contract is now "call once at startup", but several code paths never go through CLI startup:

- `pkgs/node/src/testing.zig` `initTestThreadPool()` — used by forkchoice tests that call real aggregation FFI
- `pkgs/types/src/block.zig` local test thread pool
- other xmss tests in this file that call `aggregateSignatures` without setup (recursive aggregate test, etc.)

Before, `ensureProverReady()` / `setupVerifier()` on the hot path papered over that. Now those tests can hit uninitialized FFI and segfault.

Suggest calling `setupXmssAggregation()` (and `setRayonThreads(1)` for tests) from a shared test helper, e.g. inside `initTestThreadPool()`.

pkgs/xmss/src/aggregation.zig
```diff
@@ -134,8 +114,6 @@ pub fn aggregateSignatures(
         return AggregationError.AggregationFailed;
     }
 
-    try ensureProverReady();
-
     const num_children = children_pub_keys.len;
```
ch4r10t33r, line 117: Removing `ensureProverReady()` here is fine for production — startup already paid the cost. Just flagging that every caller must have run `setupXmssAggregation()` first. No fallback anymore.

pkgs/xmss/src/aggregation.zig
```diff
@@ -219,8 +197,6 @@ pub fn verifyAggregatedPayload(public_keys: []*const hashsig.HashSigPublicKey, m
     // Get bytes from aggregated signature
     const sig_bytes = agg_sig.constSlice();
```
ch4r10t33r, line 200: Same as `aggregateSignatures`: no lazy verifier setup anymore. `verifyAggregatedPayloadBatch` in state-transition and block import paths rely on startup having run setup. OK for the node binary; not OK for unit tests unless the test helper is updated.

pkgs/xmss/src/aggregation.zig
```diff
@@ -319,7 +293,7 @@ test "aggregateSignatures and verifyAggregatedPayload with valid and invalid pub
     var signature = try keypair.sign(&message_hash, epoch);
     defer signature.deinit();
 
-    try setupProver();
+    try setupXmssAggregation();
```
ch4r10t33r, line 299: Only this xmss test calls setup. Tests further down (recursive aggregate, missing child key, etc.) still call `aggregateSignatures` / `verifyAggregatedPayload` without setup. They only work if this test ran first in the same binary. Please add setup to each test that hits the FFI, or do it once in a shared test init.

pkgs/cli/src/node.zig
```diff
@@ -561,11 +555,15 @@ pub const Node = struct {
         }
         xmss.setRayonThreads(rayon_threads);
 
-        if (options.is_aggregator) {
-            xmss.ensureProverReady() catch |err| {
-                self.logger.warn("xmss prover init failed: {any}; aggregation may be unavailable", .{err});
-            };
-        }
+        // Single XMSS aggregation setup for both prover and verifier paths
+        try xmss.setupXmssAggregation();
```
ch4r10t33r, line 559: Behavior change: aggregators used to log a warning and keep running if prover init failed (`ensureProverReady` catch). Now `try setupXmssAggregation()` exits the process for every node type.

If that's intentional (missing bytecode = don't start), fine — worth a line in the PR description so ops knows validators won't boot without aggregation artifacts either.

pkgs/cli/src/main.zig
```diff
@@ -633,23 +633,21 @@ fn mainInner(init: std.process.Init) !void {
             const reserved_system_threads: usize = 4; // main, p2p, api server, metrics server
             const desired_workers = @max(@as(usize, 1), cpu_count -| reserved_system_threads);
             const worker_count = @min(desired_workers, @as(usize, 4));
+
+            // Single XMSS aggregation setup before the Zig thread pool exists so workers
+            // cannot observe uninitialised FFI state
+            xmss.setupXmssAggregation() catch |err| {
```
ch4r10t33r, line 641: Good that setup runs before `ThreadPool.init` here.

`node.zig` still calls `setRayonThreads()` with the tuned split; this sim path does not. Pre-existing gap, but now this path also pays full twiddle precompute. Sim aggregation may not match production rayon sizing on aggregator-like workloads.

ch4r10t33r (changes requested):
Moving setup before the thread pool is the right fix for the init races. Removing lazy setup from the hot path should help aggregator throughput too.

Two things to fix before merge: Rust init should stay idempotent (OnceLock), and test harnesses need to call setup now that aggregate/verify no longer self-init.

pkgs/cli/src/node.zig
anshalshukla: Yeah its intentional without the precomputes it will be very slow to aggregate and aggregators should fix the setup as it leaves little point to run them without the setup

pkgs/xmss/src/aggregation.zig
anshalshukla: right, I think missing it in tests is not a big problem and this simplifies the code and locks in hot path so current flow is preferable

ch4r10t33r: Can we document this somewhere? Just so that we do not forget?

pkgs/xmss/src/aggregation.zig
```diff
@@ -98,17 +84,11 @@ pub fn setRayonThreads(num_threads: usize) void {
...
-pub fn setupProver() error{ProverSetupFailed}!void {
-    if (xmss_setup_prover() != 0) return error.ProverSetupFailed;
-}
-
-/// Initialize the XMSS verifier (idempotent — only runs once).
-pub fn setupVerifier() error{VerifierSetupFailed}!void {
-    if (xmss_setup_verifier() != 0) return error.VerifierSetupFailed;
+/// Initialize XMSS aggregation (both prove and verify state). Must be called
+/// exactly once at node startup, before any aggregation or verification work
+/// begins.
```
anshalshukla, line 89: It's already documented here? Do you want me to add it in PR description as well or you want more comments across the codebase? @ch4r10t33r

ch4r10t33r: I think this will do. I had missed this.

ch4r10t33r (approved):
LGTM

PR #938, opened 2026-05-26
node, network: recover from gossip ingress stall (#926)

ch4r10t33r:
## Summary

Addresses the five recovery gaps from [blockblaz/zeam#926](https://github.com/blockblaz/zeam/issues/926) when zeam nodes stop receiving inbound gossip but still report `synced` on early devnets (`finalized_slot = 0`):

- **Sync state**: `getSyncStatus()` returns `behind_peers` when wall-clock head lag exceeds the threshold while finalization is still at zero.
- **Rate-limited status refresh**: `refreshSyncFromPeers()` rotates through peers in batches of 8 instead of blasting every connected peer each tick.
- **Proactive catch-up**: when gossip has been silent for 2+ slots and head lags wall clock, start `blocks_by_range` / `blocks_by_root` from the best cached peer status without waiting for a status RPC response.
- **Gossipsub mesh healing**: re-send gossipsub subscriptions after QUIC timeout disconnects and periodically when mesh peers are low or gossip ingress stalls.
- **RPC chunk import (#894)**: when the chain-worker is enabled, missing-parent RPC chunks cache + fetch parents instead of inline `onBlock` on libxev.

## Test plan

- [ ] Deploy to devnet4 and verify zeam nodes recover from gossip stall without container restart
- [ ] Confirm `lean_node_sync_status{status="syncing"}` when head lags wall clock with `finalized_slot=0`
- [ ] Monitor status refresh logs show batched peer counts (not 55-at-once timeout storms)
- [ ] Verify proactive catch-up logs (`proactive catch-up: gossip silent...`) trigger `blocks_by_range` when gossip stops


zclawz (approved):
## Review — zclawz

**XMSS unification**: Clean. Rust-side merges `PROVER_READY` + `VERIFIER_READY` into a single `INIT: OnceLock<()>` with `catch_unwind` outside `get_or_init`, so a panicking first call leaves the lock empty and allows retry. Old code called `init_aggregation_bytecode()` twice (once in prover path, once in verifier path); now it's once. Correct.

**`stored_proof_owned = false`** (forkchoice.zig): Improvement over PR #933 — initialised to `false`, set `true` only after successful `ssz.deserialize`. Prevents the errdefer firing on an undefined `stored_proof` if `sszSerializeAndGetBytes` fails early.

**`getSyncStatus()` wall-clock lag**: Atomic `wall_head_lag_slots` with `.monotonic` load/store — appropriately relaxed for a sync-status heuristic. Check order (finalized gaps first, wall lag last) is correct.

**Rate-limited refresh**: `peerBatchWindow` is pure, well-tested. Wrap-around `(batch.start + i) % total` in the batch loop is correct.

**`findBestCatchUpPeerStatus`**: Peer-id duped under lock, freed on lock release — correctly guards against UAF if `onPeerDisconnected` races and frees the map key. Logic is sound.

**Gossip mesh healing**: `subscribed_gossip_topics` + lock, `refreshGossipMeshSubscriptions` copies under lock then sends. Lock held during `deinit` is fine.

**`cacheMissingParentRpcChunk`**: Clean extraction; missing-parent check before `chain.onBlock` is the key behavioral fix for #894/#926.

**Minor observation**: `genMockChain` now calls `setupXmssAggregation()` before `setRayonThreads` has run. Tests that call `genMockChain` before `setupTestPrimitives` will let rayon initialise with its default thread count — the `setRayonThreads(1)` in `setupTestPrimitives` is then a silent no-op. Not a correctness issue, but worth knowing if test concurrency ever matters.

Approved.

PR #941, opened 2026-05-26
node, types, metrics, xmss: cut recursive-STARK aggregation tail + per-call instrumentation (#940)

ch4r10t33r:
Addresses #940 (and #899) — drives the aggregate-build p95 down by changing the
aggregator from *eager-aggregate-then-recursively-merge* to *defer-and-flat-re-prove
over growing raws*, plus the metric instrumentation that made the diagnosis
possible.

## TL;DR

- **Root cause** (measured, not guessed): every successful aggregate deleted the
  raw signatures it consumed. The next aggregator tick on the same `att_data`
  saw the freshly-stored aggregate as a *child* + any new gossip arrivals as
  raws, forcing `rec_xmss_aggregate` to take the recursive code path at
  **~4.5 s mean** per call. On the live devnet, 60 % of proves were recursive.
- **Fix**: keep raw sigs live; pre-check whether any stored aggregate already
  covers them (skip if yes); pass raw vids as `gossip_available` to greedy so
  children dominated by raws are not picked. The FFI now runs flat
  (`num_children = 0`) on the vast majority of ticks. lean-bench extrapolation
  for flat on 16-core Hetzner: ~0.5 s for 8-validator committees.
- **Instrumentation**: new `zeam_xmss_rec_aggregate_phase_seconds{phase}` and
  `zeam_xmss_rec_aggregate_prove_by_input_seconds{num_raw, num_children}`
  histograms. Together they answered the #940 investigation plan's decision
  table from production data (see *How we got here* below) and will be the
  before/after evidence on the next devnet restart.

## Commits, in narrative order

| Commit | What | Why |
|--------|------|-----|
| `80e5d8fd` | `node`: fix missing `setupTestPrimitives` reference in `commitOneAggregateResult` test | Unblocks `zig build` / `zig build test` on `origin/main`; the helper introduced in #933 referenced an undeclared name |
| `743c42a0` | `metrics, xmss`: new `zeam_xmss_rec_aggregate_prove_by_input_seconds{num_raw, num_children}` (low-cardinality buckets) | #940 plan item 1 — input-size attribution on the prove histogram so the tail can be classified |
| `7be975b4` | `metrics, xmss, multisig-glue`: split the FFI timer 3-ways via Rust out-pointers; new `zeam_xmss_rec_aggregate_phase_seconds{phase="marshal"\|"stark"\|"post"}` | #940 plan item 2 (hypothesis A vs B) — needed to tell wrapper overhead from prover overhead |
| `cf336cf3` | `types`: drop subsumed children before STARK (`pruneSubsumedChildren`) | First-attempt perf fix targeting the two-pass greedy's subsumption edge case; left in place since it is harmless and orthogonal to commit 5 |
| `7931e696` | `node, types`: deferred aggregation — keep raw sigs live, skip when stored proof dominates raws, pass `raw_vids` to greedy, prune dominated stored aggregates on commit | The actual fix once the metrics in commits 2–3 had localized the cost to the recursive STARK driven by the eager-delete-raws lifecycle |

## How we got here (live-devnet evidence)

Snapshot of `zeam_8` (16-core Hetzner, log_inv_rate=2) before the fix:

```
zeam_xmss_rec_aggregate_phase_seconds:
  marshal  0.4 ms / call    (0.02 %)
  stark    1830 ms / call   (99.97 %)
  post     0.2 µs / call    (0.00001 %)

zeam_xmss_rec_aggregate_prove_by_input_seconds:
  num_children=0   flat        250–470 ms / call
  num_children=3-4 recursive   ~4500 ms / call
  recursive share:             60 % of proves
```

The phase split (commit 3) eliminated hypothesis A — zeam wrapper overhead is
not the bottleneck; marshal + post sum to <1 ms. The input-size labels
(commit 2) eliminated the cross-client / hardware speculation: flat proves at
this scale already meet the lean-bench extrapolation (~3.6 ms/sig). The tail
is entirely from `num_children > 0` cases at a flat ~4.5 s each.

The subset-prune (commit 4) was the natural first guess at what creates those
recursive cases — two greedy passes over `new_payloads` / `known_payloads`
that cannot cross-compare candidates. On the devnet that hypothesis was
disproved by the **fast-path delta** metric staying at 1 across runs: greedy
almost never produces strictly-dominated children at this topology, so the
prune is essentially a no-op. (Kept anyway — harmless, defensive.)

Reading the data again with that ruled out: the recursive cases arise because
the local store always contains the previously-published aggregate as a
"child" candidate, while any newly-arrived raws appear separately — and that
is structural, not a greedy bug. Commit 5 changes the lifecycle so this no
longer happens.

### Before

```text
slot S    : raws = {0..3}             → flat prove → aggregate A {0..3}
            DELETE raws from live
slot S+1  : raws = {4,5}              ← only the NEW arrivals
            stored = [A]              ← from S
            extendProofsGreedily → picks A
            sigmap iter → picks {4,5}
            recursive STARK over (A + raws)    ~4.5 s ❌
            DELETE raws → aggregate B {0..5}
slot S+2  : same shape with newer raws → another 4.5 s recursive merge
```

### After

```text
slot S    : raws = {0..3}             → flat prove → aggregate A {0..3}
            KEEP raws live
slot S+1  : raws = {0..5}             ← cumulative
            stored = [A]
            anyStoredProofCoversAll(raws ⊆ A.participants)? no
            extendProofsGreedily(gossip_available = raws) → A's bits all in raws → coverage 0 → A NOT picked
            sigmap iter → picks {0..5}
            flat STARK over raws (num_children = 0)    ~0.5 s ✅
            store aggregate B {0..5}; drop A as subsumed (commit-time prune)
slot S+2  : raws = {0..5} (no new)
            anyStoredProofCoversAll(raws ⊆ B.participants)? yes → SKIP entirely
slot S+3  : raws = {0..7} (more new)  → flat prove → aggregate C {0..7}
            ...
```

Three coupled changes:

1. **`forkchoice.zig` `commitOneAggregateResult`** — stop deleting consumed raw
   sigs from `attestation_signatures`. Drain only the snapshot's private inner
   map. Finalization-driven prune (`pruneStaleSignatures`) still bounds memory.

2. **`block.zig` `prepareAggregateAttData`** — compute `raw_vids` up front;
   `anyStoredProofCoversAll` short-circuit to `skip` when a stored aggregate
   already covers them; pass `raw_vids` as `gossip_available` to
   `extendProofsGreedily` so dominated children are not selected.

3. **`forkchoice.zig` `commitOneAggregateResult`** — before appending the new
   aggregate to `latest_new_aggregated_payloads[att_data]`, drop any existing
   entry whose participants are contained in the new one. Otherwise the list
   would accumulate one strictly-dominated entry per tick.

`shouldSuppressDuplicateAggregateCommit` is left unchanged — its predicate
(`none of snap's vids are in live`) now normally evaluates false, so it stops
firing. That is safe: `anyStoredProofCoversAll` plays that role earlier in the
pipeline, before the worker produces a duplicate proof.

`participantsContainsAll` is promoted to `pub` in `block.zig` and re-exported
via `types/lib.zig` so `forkchoice.zig` can use it for the dominated-stored-
aggregate prune.

## Metrics changes (commits 2–3, recap)

Both are zeam-internal (`zeam_*`); the cross-client spec metric
`lean_pq_sig_aggregated_signatures_building_time_seconds` is **not modified**,
so existing dashboards keep working.

- `zeam_xmss_rec_aggregate_prove_by_input_seconds{num_raw, num_children}` —
  same wall-time observation as the unlabeled prove histogram, additionally
  labeled by coarse input-size buckets.
  - `num_raw` buckets: `0`, `1`, …, `8`, `9-15`, `16-31`, `32+`
  - `num_children` buckets: `0`, `1`, `2`, `3-4`, `5-7`, `8+`
- `zeam_xmss_rec_aggregate_phase_seconds{phase}` — per-phase split inside the
  Rust FFI, reported back via out-pointers from `xmss_aggregate`:
  - `marshal` — `raw_xmss` Vec build + `.inner.clone()` × 2N + child deserialize
  - `stark` — `rec_xmss_aggregate` itself (the leanMultisig STARK)
  - `post` — `Box::into_raw` of the returned aggregate
- The bare `zeam_xmss_rec_aggregate_prove_seconds` is preserved unchanged so
  lean-bench `aggregate.flat_*_r2` comparisons stay apples-to-apples.

FFI signature change in `rust/multisig-glue/src/lib.rs`: `xmss_aggregate` now
takes three additional `*mut u64` out-params (`out_marshal_ns`, `out_stark_ns`,
`out_post_ns`). All nullable; the only caller is `pkgs/xmss/src/aggregation.zig`
which passes three stack u64s.

## Behaviour contracts

- **Validator coverage is preserved**, with or without the fix. Both the
  subset-prune (commit 4) and the dominated-stored-aggregate prune (commit 5
  step 3) drop entries that are fully covered by another entry — the
  union of validators in the kept set is unchanged.
- **Memory** stays bounded: finalization still prunes both
  `attestation_signatures` (via `pruneStaleSignatures`) and
  `latest_new_aggregated_payloads` (via `prunePayloadMapBySlot`).
- **Cross-client wire format** is unchanged. Aggregates produced by the new
  flow are ordinary `SignedAggregatedAttestation`s; peers cannot tell the
  difference. The number of aggregates published per slot may go up
  modestly (deferred path re-publishes as raws grow), but each is strictly
  larger than any prior one for the same `att_data` and the commit-time
  prune ensures we don't store redundant copies locally.

## Test plan

- [x] `zig build -Dprover=dummy` — clean.
- [x] `zig build test -Dprover=dummy` — 310 tests pass across the test
      binaries.
- [x] Test `aggregate prunes attestation signatures` updated: previously
      asserted `attestation_signatures.count() == 0` after commit
      (encoding the old delete-on-commit behaviour); now asserts the new
      contract (count stays at the original raw count).
- [ ] Deploy `0xpartha/zeam:local` to ansible devnet `zeam_8`; verify on
      the new image:
      - `zeam_xmss_rec_aggregate_prove_by_input_seconds_count{num_children!="0"}`
        drops sharply (target: <10 % of proves vs 60 % baseline)
      - `zeam_xmss_rec_aggregate_phase_seconds{phase="stark"}` mean drops
        toward lean-bench extrapolation (~30–500 ms depending on raw count)
      - `lean_pq_sig_aggregated_signatures_building_time_seconds` mean
        below 0.5 s, p95 below 1 s
      - Fast-path delta (counter − prove_count) grows: skips when no new
        coverage to add
      - No duplicate-aggregate gossip storms (peer dedup behaviour
        verified via `lean_pq_sig_aggregated_signatures_verification_time_seconds`
        and gossip volume metrics)

## Out of scope (deliberately)

- Renaming or re-scoping `lean_pq_sig_aggregated_signatures_*` for cross-
  client comparison. The leanMetrics spec text is too thin to pin
  semantics; the divergence is the same one documented in #940's
  cross-client analysis. Separate spec work.
- Upstream leanMultisig changes (persistent prover state, batched FFI
  accepting multiple `att_data` per call). The structural change in this
  PR avoids the recursive code path almost entirely at this topology, so
  the upstream cost reduction is no longer on the critical path.
- Changing `min_aggregation_inputs` or `isAggregatorTrivialInput`. Those
  remain as the early-bail filters; deferred aggregation just changes
  what the worker does once those filters say "yes, aggregate."

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -103,6 +103,14 @@ const Metrics = struct {
     /// Buckets are deliberately low-cardinality — see `numRawBucket` /
     /// `numChildrenBucket` below for the exact mapping.
     zeam_xmss_rec_aggregate_prove_by_input_seconds: XmssRecAggregateProveByInputHistogram,
+    /// Per-phase breakdown inside `xmss_aggregate` reported back from the
+    /// Rust FFI via out-pointers. The three phases ("marshal", "stark",
+    /// "post") sum to a value very close to one observation on
+    /// `zeam_xmss_rec_aggregate_prove_seconds` (delta is the FFI call/return
+    /// overhead, sub-microsecond). This lets us tell whether zeam's per-call
+    /// cost lives in argument deserialization or in the leanMultisig STARK
+    /// itself — the two have very different fixes. See #940.
+    zeam_xmss_rec_aggregate_phase_seconds: XmssRecAggregatePhaseHistogram,
     lean_pq_sig_aggregated_signatures_verification_time_seconds: PQSigAggregatedVerificationHistogram,
     lean_pq_sig_aggregated_signatures_valid_total: PQSigAggregatedValidCounter,
     lean_pq_sig_aggregated_signatures_invalid_total: PQSigAggregatedInvalidCounter,
@@ -450,6 +458,11 @@ const Metrics = struct {
     const XmssRecAggregateProveHistogram = metrics_lib.Histogram(f32, &[_]f32{ 0.05, 0.1, 0.25, 0.5, 0.75, 1.0, 1.5, 2.0, 3.0, 5.0, 10.0, 15.0 });
     const XmssRecAggregateProveByInputLabel = struct { num_raw: []const u8, num_children: []const u8 };
     const XmssRecAggregateProveByInputHistogram = metrics_lib.HistogramVec(f32, XmssRecAggregateProveByInputLabel, &[_]f32{ 0.05, 0.1, 0.25, 0.5, 0.75, 1.0, 1.5, 2.0, 3.0, 5.0, 10.0, 15.0 });
+    // Buckets span sub-millisecond (post phase = pointer wrap, typically ns)
+    // through multi-second STARK proves. Phase label values are
+    // "marshal" | "stark" | "post" — see XmssRecAggregatePhaseLabel below.
+    const XmssRecAggregatePhaseLabel = struct { phase: []const u8 };
+    const XmssRecAggregatePhaseHistogram = metrics_lib.HistogramVec(f32, XmssRecAggregatePhaseLabel, &[_]f32{ 0.0001, 0.001, 0.005, 0.01, 0.05, 0.1, 0.25, 0.5, 1.0, 2.0, 4.0, 8.0 });
     const PQSigAggregatedVerificationHistogram = metrics_lib.Histogram(f32, &[_]f32{ 0.1, 0.25, 0.5, 0.75, 1, 1.25, 1.5, 2, 4 });
     const PQSigAggregatedValidCounter = metrics_lib.Counter(u64);
     const PQSigAggregatedInvalidCounter = metrics_lib.Counter(u64);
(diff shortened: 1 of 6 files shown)


zclawz (approved):
Approved based on the review posted at https://github.com/blockblaz/zeam/pull/941#issuecomment-4552997865. LGTM.

PR #944, opened 2026-05-27
feat(xmss): shadow-sim CPU-cost injection for aggregation/verification

GrapeBaBa:
## Summary

Ports qlean-mini's Shadow-simulator CPU-cost injection to zeam. Under the [Shadow](https://shadow.github.io/) network simulator the virtual clock advances on I/O and timer syscalls, not CPU work — so XMSS signature aggregation (the network's heaviest CPU cost) appears **instant** in simulated time, hiding the exact bottleneck shadow simulations need to study. This injects a modeled `sleep = n_signatures / rate` into aggregation and verification so the cost shows up in virtual time.

- **New `pkgs/xmss/src/shadow_cost.zig`** — pure `computeDelayNs(rate, n)` delay math + rate config, resolved once at node startup with precedence **CLI flag > env var > off**.
- **Two CLI flags** (`--shadow-xmss-aggregate-signatures-rate`, `--shadow-xmss-verify-aggregated-signatures-rate`, in signatures/sec) with matching env vars `ZEAM_SHADOW_XMSS_AGGREGATE_SIGNATURES_RATE` / `ZEAM_SHADOW_XMSS_VERIFY_AGGREGATED_SIGNATURES_RATE`. Names mirror qlean 1:1 for cross-client comparison.
- Sleep injected at the end of `aggregateSignatures` / `verifyAggregatedPayload` (success paths only) via `zeam_utils.sleepNs`, **guarded** so the default (feature-off) path does no syscall.
- **Default off everywhere** → zero behavior change to production and existing sims unless a rate is set. Real crypto is kept (sleep-only model); a fake-crypto path is left as a future seam.

### Why blocking sleep is faithful

Aggregation runs synchronously on the libxev event-loop thread, so real heavy aggregation already occupies the node. A blocking sleep models the node being busy during aggregation — the same semantics as qlean's `sleep_for`.

## Test plan

- [x] `zig build test-shadow-cost` — unit tests for the delay math (null/≤0/non-finite → 0; `n=0` → 0; `n=100, rate=22.704` → ≈4.4 s). 3/3 pass.
- [x] Full `zig build` (Zig 0.16, rust glue) succeeds; `zeam node --help` shows both flags as `FLOAT` with correct help + env hints.
- [ ] (Follow-up) Exercise under the lean-quickstart Shadow harness: set a slow rate on one node and confirm its finality lag / aggregation duration increases relative to a feature-off node.

## Notes

- The cost model is intentionally rate-only (`n / rate`), matching qlean; a fixed STARK-setup term can be added later if `bench-xmss` shows a non-trivial intercept.
- `verifyAggregatedPayloadBatch` is not instrumented in this PR (kept scope to the two functions in the original design); can be added if the batch path becomes the primary verify route.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -201,6 +201,7 @@ pub fn build(b: *Builder) !void {
         .target = target,
         .optimize = optimize,
     });
+    zeam_xmss.link_libc = true; // shadow_cost reads env via libc getenv
     zeam_xmss.addImport("ssz", ssz);
     zeam_xmss.addImport("@zeam/metrics", zeam_metrics);
     zeam_xmss.addImport("@zeam/utils", zeam_utils);
@@ -657,6 +658,8 @@ pub fn build(b: *Builder) !void {
     setTestRunLabelFromCompile(b, run_node_test, node_tests);
     test_step.dependOn(&run_node_test.step);
 
+    // Build shadow_cost as its own module for focused pure-function tests. This
+    // gives it independent globals from the copy imported through @zeam/xmss.
     const zeam_shadow_cost = b.createModule(.{
         .root_source_file = b.path("pkgs/xmss/src/shadow_cost.zig"),
         .target = target,
--- a/build.zig
+++ b/build.zig
@@ -657,6 +657,21 @@ pub fn build(b: *Builder) !void {
     setTestRunLabelFromCompile(b, run_node_test, node_tests);
     test_step.dependOn(&run_node_test.step);
 
+    const zeam_shadow_cost = b.createModule(.{
+        .root_source_file = b.path("pkgs/xmss/src/shadow_cost.zig"),
+        .target = target,
+        .optimize = optimize,
+    });
+    zeam_shadow_cost.link_libc = true; // shadow_cost reads env via libc getenv
+    const shadow_cost_tests = b.addTest(.{
+        .root_module = zeam_shadow_cost,
+    });
+    const run_shadow_cost_test = b.addRunArtifact(shadow_cost_tests);
+    setTestRunLabelFromCompile(b, run_shadow_cost_test, shadow_cost_tests);
+    test_step.dependOn(&run_shadow_cost_test.step);
+    const shadow_cost_test_step = b.step("test-shadow-cost", "Run shadow sim-cost unit tests");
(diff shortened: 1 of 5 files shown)


build.zig
```diff
@@ -657,6 +657,21 @@ pub fn build(b: *Builder) !void {
     setTestRunLabelFromCompile(b, run_node_test, node_tests);
     test_step.dependOn(&run_node_test.step);
 
+    const zeam_shadow_cost = b.createModule(.{
+        .root_source_file = b.path("pkgs/xmss/src/shadow_cost.zig"),
+        .target = target,
+        .optimize = optimize,
```
ch4r10t33r, line 666: Heads up: this builds `shadow_cost.zig` as its own module, separate from the one xmss imports relatively. That means two independent copies of the `agg_rate` / `verify_rate` globals. Fine for the current pure-function tests, but any future test that calls `init()` and then drives aggregation through xmss won't see the state it set. Worth a one-line comment here so the next person isn't surprised.

ch4r10t33r (commented):
Two notes from a local review.

pkgs/xmss/src/shadow_cost.zig
```diff
@@ -0,0 +1,63 @@
...
+var agg_rate: ?f64 = null;
+var verify_rate: ?f64 = null;
+
+const ENV_AGG = "ZEAM_SHADOW_XMSS_AGGREGATE_SIGNATURES_RATE";
+const ENV_VERIFY = "ZEAM_SHADOW_XMSS_VERIFY_AGGREGATED_SIGNATURES_RATE";
+
+// Zig 0.16 removed `std.process.getEnvVarOwned`; use libc `getenv` for a simple
+// process-env read (no allocation, no free). Returns null when unset/unparseable.
+fn readEnvRate(key: [*:0]const u8) ?f64 {
+    const raw = std.c.getenv(key) orelse return null;
```
ch4r10t33r, line 27: `std.c.getenv` needs libc. The test module sets `link_libc = true`, but in the real build the xmss module relies on libc being pulled in transitively (rust glue / crypto). If that ever stops being true, the link error lands here. Either set `link_libc = true` on the xmss module too, or leave a `// requires libc via <dep>` note next to this call.

ch4r10t33r (approved):
LGTM

PR #948, opened 2026-05-28
node, metrics: defer gossip-block hashTreeRoot to chain-worker + libxev callback histogram (#942)

ch4r10t33r:
## Summary

- **Step 1 – libxev callback duration histogram**: adds `zeam_libxev_callback_duration_seconds{site=...}` (HistogramVec) to `pkgs/metrics/src/lib.zig` and instruments `onGossip.block`, `onGossip.attestation`, `onGossip.aggregation`, `onGossip.block.hash_tree_root`, `onGossip.block.dispatch`, and `onInterval.tick` call-sites. Exposes via `/metrics` scrape so we can attribute slow drain passes to specific sites.

- **Step 2 – defer hashTreeRoot off libxev**: the unconditional `hashTreeRoot(BeamBlock)` that ran on the libxev thread for every inbound gossip block is removed from the common path. When a chain-worker is available and the parent is already in fork-choice (the hot path during normal sync), the block is dispatched to the worker immediately with `precomputed_block_root = null`; the worker computes the root itself. hashTreeRoot is still computed on libxev only for: (a) orphan blocks whose parent is not yet in fork-choice, and (b) future-slot blocks queued for later replay — both already rare and bounded.

## Why this matters

During devnet catch-up, gossip delivers 20–30 blocks in a burst. Each `hashTreeRoot(BeamBlock)` took ~5–15 ms; with 25 blocks in a single libxev drain pass that adds up to 375 ms — explaining the `zeam_xev_clock_until_done_slow_ge_500ms_total` counter incrementing ~70/minute during the stall.

## Correctness notes

- `protoArray.onBlock` (forkchoice.zig:97–101) returns early (no-op, no error) when `block.blockRoot` is already present, making the libxev-side `hasBlock` dedup an optimisation, not a correctness requirement.
- Gossipsub message-id dedup at the Rust libp2p layer means most actual duplicates never reach `onGossip`.
- `handleChainImportedBlock` (node.zig:631–678) fires `processCachedDescendants` with the resolved root after the worker completes, so the `processed_block_root = null` return on the fast path is safe.

## What to observe after deployment

Scrape `zeam_libxev_callback_duration_seconds` from `/metrics`. If `onGossip.block.dispatch` p99 is still >100 ms after this change, the next candidate is moving `sszClone` off libxev (step 3).

## Test plan

- [x] `zig build test` — all test suites green (metrics histogram appears in /metrics output test added)
- [ ] Deploy to devnet, scrape `/metrics`, confirm histogram buckets populate
- [ ] Verify `zeam_xev_clock_until_done_slow_ge_500ms_total` no longer increments rapidly during catch-up

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -198,6 +198,12 @@ const Metrics = struct {
     zeam_xev_clock_until_done_slow_ge_500ms_total: ZeamXevClockUntilDoneSlowGe500msCounter,
     /// Count of clock-loop `run(.until_done)` drains taking ≥1s wall time.
     zeam_xev_clock_until_done_slow_ge_1s_total: ZeamXevClockUntilDoneSlowGe1sCounter,
+    /// Wall time spent inside individual libxev callbacks, labeled by callsite.
+    /// Used to attribute slow `zeam_xev_clock_until_done_drain_seconds` to a
+    /// specific synchronous chunk of work executed on the libxev thread (#942).
+    /// Each callsite name is a fixed low-cardinality string; see
+    /// `observeLibxevCallback` for the canonical site list.
+    zeam_libxev_callback_duration_seconds: LibxevCallbackDurationHistogram,
     // Fork-choice tick interval duration: actual elapsed time between forkchoice tickIntervalUnlocked calls
     zeam_fork_choice_tick_interval_duration_seconds: ForkChoiceTickIntervalDurationHistogram,
     /// Wall time for the per-slot aggregation tick (interval 2): `maybeAggregateOnInterval`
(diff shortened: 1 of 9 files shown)


zclawz (approved):
Approved after review.

I checked the changed paths around gossip ingress/decode metrics, libxev callback histograms, deferred block root computation through the chain worker, future-slot/orphan handling, and the Rust libp2p yamux-only/message-id changes. The direction looks correct and the async ownership paths are accounted for.

Validation:
- GitHub checks on 0f3cefcb4b69ace1c8bebe25b1b41eecce849cc9 are green for lint/build/test/docker/provers.
- Local `cargo test --manifest-path rust/libp2p-glue/Cargo.toml --quiet` passed: 17/17.
- Local `zig build test` could not be used as a signal in this workspace because the installed Zig is 0.15.2 while cached deps expect a newer Build API (`b.graph.io`), but CI covered the Zig gates successfully.

No blocking findings from my pass.

PR #949, opened 2026-05-28
sync: collapse fetchBlockByRoots / fetchBlockByRootsFromPeer into a single function

markolazic01:
## Description

- Remove `fetchBlockByRootsFromPeer`, rename the remaining function to `fetchBlockByRoots` with an optional `preferred_peer` parameter
- Update all call sites to pass `null` where no peer preference exists

Closes #913


ch4r10t33r (approved):
LGTM. Clean behavior-preserving refactor: collapses fetchBlockByRoots / fetchBlockByRootsFromPeer into one function with an optional preferred_peer. Verified no leftover references to the removed function and all call sites pass the third arg correctly. Closes #913.

PR #950, opened 2026-05-29
feat: add min_slot filter to selectPeer with fallback

markolazic01:
## Motivation

Peers reporting `head_slot=0` are either freshly connected (no status exchanged yet) or genuinely empty. Routing `blocks_by_root` requests to them wastes a round-trip and a timeout cycle.

## Description

- Add optional `min_slot` parameter to `selectPeerExcluding` that filters out peers whose last-known `head_slot` is below the threshold
- Peers with no status default to `head_slot=0`
- Falls back to the full candidate set if no peer meets the threshold
- Thread the parameter through `selectPeerCopy`, `selectPeer`, and `selectPeerExcluding` on `Network`
- Existing callers pass `null` to preserve current behavior

Closes #911

Related: #909


pkgs/node/src/locking.zig
```diff
@@ -1006,16 +1001,28 @@ pub fn ConnectedPeersImpl(comptime PeerInfo: type) type {
...
+                        }
+                    }
+                    if (range_capable_only and @hasField(PeerInfo, "blocks_by_range_unavailable")) {
+                        if (entry.value_ptr.blocks_by_range_unavailable) continue;
+                    }
+                    try candidates.append(allocator, entry.value_ptr.peer_id);
                 }
-                try candidates.append(allocator, entry.value_ptr.peer_id);
+                if (candidates.items.len > 0 or effective_min_slot == null) break;
+                effective_min_slot = null;
```
ch4r10t33r: **High.** When the filter yields zero candidates, this retries with `null` and routes to head_slot=0 peers anyway — the exact behavior #911 wanted to prevent. The caller has no signal that fallback fired. At least log a warn + bump a metric so operators can see when filtering degrades, and consider returning null so the caller can decide (retry later vs accept any peer).

pkgs/node/src/network.zig
```diff
@@ -1089,8 +1089,8 @@ pub const Network = struct {
         // returns the backend error and the existing retry paths handle it.
         const peer = if (preferred_peer) |peer_id| blk: {
```
ch4r10t33r, line 1090: **Critical.** This is the `blocks_by_root` peer pick — the call site #911 was about. Still passes `null`. The whole point of this PR is to let this call avoid head_slot=0 peers; without the actual `min_slot` value here the new param is dead code. At minimum pass the caller's expected slot.

pkgs/node/src/network.zig
```diff
@@ -429,16 +429,16 @@ pub const Network = struct {
...
 
-    pub fn selectPeerExcluding(self: *Self, exclude: ?[]const u8) !?[]u8 {
-        return self.connected_peers.selectPeerExcluding(self.allocator, exclude, false);
+    pub fn selectPeerExcluding(self: *Self, exclude: ?[]const u8, min_slot: ?u64) !?[]u8 {
+        return self.connected_peers.selectPeerExcluding(self.allocator, exclude, false, min_slot);
     }
 
     pub fn selectPeerForRangeSyncExcluding(self: *Self, exclude: ?[]const u8) !?[]u8 {
-        return self.connected_peers.selectPeerExcluding(self.allocator, exclude, true);
+        return self.connected_peers.selectPeerExcluding(self.allocator, exclude, true, null);
```
ch4r10t33r: **Medium.** Same here for the range-sync retry path. `selectPeerForRangeSyncExcluding` hardcodes `null` for `min_slot`. The range-sync caller knows what slot it's syncing toward — thread it through.

pkgs/node/src/locking.zig
```diff
@@ -1006,16 +1001,28 @@ pub fn ConnectedPeersImpl(comptime PeerInfo: type) type {
...
-            var it = self.map.iterator();
-            while (it.next()) |entry| {
-                if (exclude) |ex| {
-                    if (std.mem.eql(u8, entry.key_ptr.*, ex)) continue;
-                }
-                if (range_capable_only and @hasField(PeerInfo, "blocks_by_range_unavailable")) {
-                    if (entry.value_ptr.blocks_by_range_unavailable) continue;
+            var effective_min_slot: ?u64 = min_slot;
+            while (true) {
+                var it = self.map.iterator();
```
ch4r10t33r: **Medium.** No test asserts the filter actually filters. The existing test was updated for the signature change but only ever passes `null`. Add a test with peers at slots 0, 5, 10 and `min_slot=7` → only slot-10 peer eligible; with `min_slot=20` → fallback fires.

ch4r10t33r (changes requested):
Mechanism added but nothing uses it every production caller still passes `null`. This PR claims to close #911 but the bug (routing `blocks_by_root` to head_slot=0 peers) is unchanged. Also: the fallback silently disables the filter on the exact case that matters (only head_slot=0 peers available), and no test covers the new behavior.

Please can you address these comments? @markolazic01

pkgs/node/src/network.zig
markolazic01: applied

pkgs/node/src/locking.zig
markolazic01: Thanks for all the feedback and apologies for the inconvenience.

Removed the fallback, now returning null. What do you exactly prefer when it comes to the expected slot, does not seem to be easily accessible from that point? Is latest finalized slot alright? Maybe get latest slot + 1?
Please let me know your preference.

Also on the fallback part, do you want me to introduce retries in places where the function is used, or how otherwise should the user decide if he wants to retry with `null` or wait?

pkgs/node/src/locking.zig
markolazic01: Will surely apply this, just once all the other points are properly fixed.

pkgs/node/src/network.zig
markolazic01: This is tackled inside of this comment here:
https://github.com/blockblaz/zeam/pull/950#discussion_r3352473367

pkgs/node/src/locking.zig
ch4r10t33r: **Q1 (`min_slot` value):** prefer `our_head_slot + 1` — peer must be at least as advanced as us to be useful for a `blocks_by_root` fetch. That's the accurate predicate.

@g11tech — would like your view here too. Trade-off: tighter `our_head_slot + 1` vs minimal `1` (just filter the never-statused peers from #911) vs `our_finalized_slot` (weaker but always satisfiable). Which is the right default for the `blocks_by_root` call site?

**Q2 (fallback / retry):** let null bubble up; existing retry paths already handle it.

- `blocks_by_root`: `ensureBlocksByRootRequest` returning null → caller logs `NoPeersAvailable`, bumps existing dedup-bucket metric, returns. The fetch is re-scheduled on the next tick via the existing pending-roots queue (gossip / parent-walk machinery already retries). No new retry loop needed.
- `selectPeerForRangeSyncExcluding`: null already falls back to head-by-root per existing logic. Same — propagate null.

Visibility additions worth including in this PR:
- Metric `zeam_select_peer_filter_drops_total{reason="head_slot_below_min"}` so operators see when the filter is biting.
- Rate-limited WARN log per "no peer matched min_slot" event so log spam stays bounded under sustained no-good-peer windows.

No new retry semantics needed in this PR — pure additive filter + observability.

markolazic01: Thank you @ch4r10t33r, this is very useful. Will also await the response from @g11tech.
I think maybe ideally we could try with `our_head_slot + 1`, then fallback to `1`(minimal), then fallback to `null`? Btw, does `our_finalized_slot + 1` do us any good?

g11tech: our_head_slot + 1 may sense, no need to fallback below our_finalized_slot + 1

g11tech: but i don't think fallback is needed, if there is a new block somewhere on a chain that we are not on, it will be on a greater slot than our head slot

pkgs/node/src/node.zig
```diff
@@ -2610,7 +2609,7 @@ pub const BeamNode = struct {
         if (missing_roots.items.len == 0) return;
 
         const handler = self.getReqRespResponseHandler();
-        const maybe_request = self.network.ensureBlocksByRootRequest(missing_roots.items, depth, handler, preferred_peer) catch |err| blk: {
+        const maybe_request = self.network.ensureBlocksByRootRequest(missing_roots.items, depth, handler, preferred_peer, self.chain.forkChoice.getCurrentSlot() + 1) catch |err| blk: {
```
ch4r10t33r:
```suggestion
        const maybe_request = self.network.ensureBlocksByRootRequest(missing_roots.items, depth, handler, preferred_peer, self.chain.forkChoice.getHead().slot + 1) catch |err| blk: {
```
`getCurrentSlot()` is wall-clock; peers can't be ahead of wall-clock, so this filters out every peer. Use our head slot instead.

pkgs/node/src/node.zig
```diff
@@ -1797,8 +1797,7 @@ pub const BeamNode = struct {
         range_unavailable: bool,
     ) void {
         const node_name = self.node_registry.getNodeNameFromPeerId(snap.peer_id_copy);
-
-        const next_peer_opt = self.network.selectPeerForRangeSyncExcluding(snap.peer_id_copy) catch null;
+        const next_peer_opt = self.network.selectPeerForRangeSyncExcluding(snap.peer_id_copy, self.chain.forkChoice.getCurrentSlot() + 1) catch null;
```
ch4r10t33r:
```suggestion
        const next_peer_opt = self.network.selectPeerForRangeSyncExcluding(snap.peer_id_copy, self.chain.forkChoice.getHead().slot + 1) catch null;
```
Same fix — `getCurrentSlot()` (wall-clock) → `getHead().slot` (our head).

ch4r10t33r (changes requested):
Critical: `getCurrentSlot()` returns the wall-clock slot, not our head slot. No peer can have `head_slot > wall_clock`, so `wall_clock + 1` rejects every peer → sync stalls. Swap to `getHead().slot + 1` at both sites.

pkgs/node/src/node.zig
ch4r10t33r:
```suggestion
        const next_peer_opt = self.network.selectPeerForRangeSyncExcluding(snap.peer_id_copy, self.chain.forkChoice.getHead().slot + 1) catch null;
```
Still `getCurrentSlot()` (wall-clock) here — same starvation as the blocks_by_root site. Swap to `getHead().slot + 1`.

ch4r10t33r (changes requested):
Other call site at node.zig:2612 fixed — thanks. This one was missed; same fix.

ch4r10t33r (approved):
LGTM

ch4r10t33r (approved):
LGTM. Both call sites now use `getHead().slot + 1` — peer filter behaves correctly under sync. Metric + tests in place.

PR #953, opened 2026-05-29
node, cli, types: blocks_by_range threshold 64→4 + per-call aggregator-children cap (#942, #940)

ch4r10t33r:
Two independent observability/perf follow-ups, bundled because PR #943 became orphaned (PR #941 it stacked on was already merged into main).

---

## (1) `BLOCKS_BY_RANGE_SYNC_THRESHOLD` 64 → 4 (#942 follow-up)

Lowers `BLOCKS_BY_RANGE_SYNC_THRESHOLD` from 64 to 4 so realistic post-#948 devnet sync gaps (5–59 slots) hit the bulk `blocks_by_range` fast path instead of the slower per-block `blocks_by_root` parent-walk.

### Devnet measurement that motivated this

After the PR #948 merge, on the 2026-05-29 devnet, the upstream ethlambda snappy encoder bug is still emitting broken block payloads on the wire. On zeam_8 after ~1 h:

```
blocks received via gossip (decoded ok)              = 16
blocks dropped by snappy decode                      = 56
zeam_libxev_callback_duration_seconds_count{site="onGossip.block"} = 16
zeam_8 head_slot                                     = 960
```

16 of 960+ slot blocks reached the app via gossip; the rest filled in by catch-up RPC. Per-node behindness then settled at:

| node | gap to wall-clock |
|---|---:|
| zeam_15 | 5 |
| zeam_14 | 12 |
| zeam_13 | 20 |
| zeam_12 | 28 |
| zeam_11 | 35 |
| zeam_10 | 43 |
| zeam_9 | 51 |
| zeam_8 | 59 |

Every gap is under the old `BLOCKS_BY_RANGE_SYNC_THRESHOLD = 64`, so the catch-up gate at `pkgs/node/src/node.zig:1517` always picked `blocks_by_root` — one round-trip per block. Lowering the cutoff to 4 flips the gate at gap ≥ 5: gaps 1–4 stay on the cheaper per-block path; anything larger hits `blocks_by_range` and recovers in a single round trip.

Once the upstream encoder bug at ethlambda is fixed and gossip block delivery rises back to ~100 %, this constant becomes largely dormant — typical catch-up gaps under healthy gossip are 0–1 slot.

Pre-existing test `shouldCatchUpFromPeerStatus small gaps use by-root not threshold gate` updated to preserve its assertion at the new threshold value (inputs `(50, 0, 100)` → `(3, 0, 10)`) without changing the property being tested.

---

## (2) Per-call aggregator-children cap (#940, cherry-pick of `cda85a3f`)

Cherry-picked the single functional commit from #943 (now closed as orphaned — PR #941 it was stacked on was already merged). Adds `--max-aggregation-children` CLI flag (default `0` = uncapped) that caps the number of child STARK proofs `prepareAggregateAttData` passes into the recursive aggregator-worker. Issue #940 is still open and main does not yet carry this mechanism; rationale and bucket-bucket prove-time data are in the original commit message.

Auto-merge cleanly resolved the rebase across the 4 files that have moved on main since `cda85a3f` was written (`pkgs/cli/src/main.zig`, `pkgs/node/src/{chain,forkchoice,node}.zig`).

---

## Test plan

- [x] `zig build test -Dprover=dummy` clean — all tests pass
- [ ] Deploy to devnet, observe `peer X is far ahead (gap=N slots), initiating bulk catch-up via blocks_by_range` log line firing on gaps as small as 5
- [ ] Re-measure per-node behindness — expect gap to wall-clock to shrink for laggard nodes
- [ ] Validate `--max-aggregation-children` cap clamps `prepareAggregateAttData` invocations with `num_children > cap`

## Related

Follow-up to #948 / #942 (threshold change). Picks up #943 / #940 (cap-children).


zclawz:
Review pass from zclawz:

I went through the threshold change, the `max_aggregation_children` plumbing, and the Rust libp2p-glue follow-up changes. The functional direction looks good to me, but I’m not comfortable marking this ready/approved while the current lint check is red.

Blocking before merge:

```text
libp2p-glue/src/lib.rs:595:5
error: consider using `sort_unstable_by_key`
rows.sort_unstable_by(|a, b| b.0.cmp(&a.0));

suggested:
rows.sort_unstable_by_key(|b| std::cmp::Reverse(b.0));
```

This is from the GitHub `lint` job on head `840cbf028bf16f7b423ea71d9da453b20df46b44` (`-D warnings` makes the clippy lint fatal). Once that one-line fix lands and CI goes green, I don’t see a code-level blocker in the diff.

Small non-blocking doc nit: the PR body says `--max-aggregation-children` default `0 = uncapped`, but the code/docs implement default `0` as flat-only / no child recursive merges. The code-side behavior is internally consistent; the PR description should be corrected so operators don’t read it backwards.

Validation note: I couldn’t run local Rust checks in this environment because `cargo` is not installed. Local `zig build` also can’t run here with the installed Zig 0.15.2 because deps expect the project CI toolchain (CI logs show Zig 0.16.0). I used the live GitHub check logs as the validation source.

zclawz (approved):
Approved after re-checking the latest head. The prior clippy blocker is fixed by 2011d36; lint is green, and the follow-up diff is the expected one-line sort_unstable_by_key change.

zclawz (approved):
Approved after re-checking the latest head. The prior clippy blocker is fixed by 2011d36; lint is green, and the follow-up diff is the expected one-line sort_unstable_by_key change.

PR #954, opened 2026-05-29
node: detect stuck-mesh-cluster and force full peer-status refresh (#942 follow-up)

ch4r10t33r:
## Summary

Adds a one-shot full-fanout `sendStatusToPeer` burst when the cached `latest_status.head_slot` across all visible peers tops out far below wall-clock — the signature of a peer-pool that is all stuck against the same upstream gossip-ingress problem. Rate-limited to one fanout per `SYNC_STATUS_STUCK_CLUSTER_REFRESH_COOLDOWN_SLOTS` slots so the unbatched RPC burst can't re-create the en-masse status-RPC-timeout cascade from #926.

## What this addresses

Observed on the 2026-05-29 devnet run of PR #953: `zeam_8` was 254 slots behind wall-clock, but every cached `latest_status.head_slot` topped out at 45 — a peer (`lantern_2`) that was itself stuck. `findBestCatchUpPeerStatus` correctly picked the best-of-stale option each interval, fired `blocks_by_root` against it, made no progress for 25+ minutes.

Cross-client check confirmed it was a peer-info problem, not a chain problem: `grandine_4` was at `head=377` against wall=378. The chain was fine. The peers `zeam_8` had cached status for were just all behind.

## The fix

1. **New pure helper** `shouldForceFullPeerStatusRefresh` in `blocks_by_range_sync.zig`. Fires when `best_peer_head + threshold < wall_slot` AND a cooldown has elapsed.
2. **Refactor** `refreshSyncFromPeers` into `refreshSyncFromPeersImpl(RefreshKind)` — `.batched` keeps the existing #926 behaviour; `.full_fanout` walks every connected peer in one tick.
3. **Detector** `maybeForceFullPeerStatusRefresh` wires the helper to a `last_stuck_cluster_refresh_slot` atomic, logs the detection, and triggers the fanout.
4. **Wired** into `runSyncRecoveryOnInterval` on the same per-slot cadence as `maybeInitiateProactiveCatchUp`.

Two new constants, both at 16 (~64 s on 4 s slots):
- `SYNC_STATUS_STUCK_CLUSTER_PEER_LAG_THRESHOLD_SLOTS = 16`
- `SYNC_STATUS_STUCK_CLUSTER_REFRESH_COOLDOWN_SLOTS = 16`

## What this is NOT

This is **not** peer discovery. It can't help when zeam is connected to a genuinely stuck peer pool with zero healthy peers anywhere on the connection list. Peer-discovery / re-bootstrap when stuck is a separate gap, suitable for a follow-up.

## Test plan

- [x] 5 new unit tests in `blocks_by_range_sync.zig` (fires on stuck condition; off when best peer at wall; off under threshold; rate-limited; safe under non-monotonic slot)
- [x] `zig build test -Dprover=dummy` — clean
- [ ] Deploy to devnet, reproduce the PR #953 stuck condition, observe `"stuck mesh cluster detected"` warning + `"status refresh full fanout"` info
- [ ] Confirm cooldown rate-limit suppresses subsequent firings within 16 slots
- [ ] Compare per-node gap-to-wall-clock against the prior run

## Related

Follow-up to PR #953 (which lowered `BLOCKS_BY_RANGE_SYNC_THRESHOLD` 64 → 4 — that change worked but wasn't enough on its own when the visible mesh is stuck). Both are derivatives of #942.


zclawz (dismissed):
Reviewed the stuck-mesh-cluster recovery changes. The pure predicate and cooldown behavior are covered by focused tests, the node wiring preserves the existing batched refresh path while keeping full-fanout rate-limited, and the current lint/test/build checks are green. No code-level blockers from me.

zclawz (changes requested):
Re-reviewed the two new commits (`77d21e5`, `c4a9661`). I’m not re-approving this head yet — I think the orphan-cache replacement drops the child block without leaving a path to import/refetch it.

Blocking concern:

In both updated paths:

- `processBlockByRootChunk` computes `block_root`, then immediately `removePendingBlockRoot(block_root)` before calling `cacheMissingParentRpcChunk`.
- `cacheMissingParentRpcChunk` now no longer caches `block_root`; it only enqueues `signed_block.block.parent_root` and flushes that fetch.
- When that parent later imports, `processCachedDescendants(parent_root)` only sees entries from `network.block_cache` / child links. The child block was intentionally not cached, and its original pending root was already removed, so there is no obvious mechanism that refetches/imports the child.

Same shape exists in the gossip orphan path: it removes any pending `block_root`, skips caching the orphan, queues only the parent, and then returns. The comment says “When the parent arrives, the chain will re-fetch this block via blocks_by_root”, but I can’t find the state that would cause that re-fetch after the pending root is removed and no descendant cache entry is stored.

So this looks like it fixes the invalid-free crash by dropping the orphan child on the floor, which can stall the parent walk / lose the requested descendant. I’d be happy with either:

1. keep a non-destructive child-refetch marker so importing the parent re-queues the child root, or
2. don’t remove `block_root` from pending until there is a replacement retry path, or
3. add a test that exercises: receive child via blocks_by_root with missing parent → fetch/import parent → child is fetched/imported afterward.

CI note: lint/test/build mostly look fine, but the current macOS build check is red due to a rustup DNS failure and Ubuntu test is still in progress at the time of this review. That looks infra-ish; the code-path issue above is the blocker for my approval.

zclawz (approved):
Re-reviewed latest head `c535f8a` after the orphan-dependent fix. This resolves my prior blocker.

The new `orphan_dependents` map records `parent_root → child_root` for dropped orphan chunks, drains it when the parent imports, and re-queues the child root for `blocks_by_root` refetch. That supplies the missing state/path I was worried about in the previous review. The duplicate suppression, deinit of stored `ArrayList`s, and no-early-return in `processCachedDescendants` all look right to me.

One tiny non-blocking nit: the new regression test’s final batch-pending check is effectively observational rather than assertive because `processCachedDescendants` flushes immediately, but it does still assert the key invariant that the dependency is tracked and then drained on parent import.

CI is still running on this head as I write this, so please wait for it before merge, but I’m satisfied with the code changes.

PR #957, opened 2026-05-31
aggregator and proposer aggregation improvements

anshalshukla:
- Aggregator: does a single attestation_data aggregation based on the current forkchoice
- Proposer: enables partial block production as per configurable timeout parameterr


pkgs/node/src/validator_client.zig
```diff
@@ -137,25 +137,35 @@ pub const ValidatorClient = struct {
             }
 
             self.logger.debug("constructing block for slot={d} proposer={d}", .{ slot, slot_proposer_id });
-            const produced_block = try self.chain.produceBlock(.{ .slot = slot, .proposer_index = slot_proposer_id });
-            self.logger.info("produced block for slot={d} proposer={d} with root={x}", .{ slot, slot_proposer_id, &produced_block.blockRoot });
+            // Spawn the deadline-aware build worker, then block until it
+            // self-truncates and exits. The worker bounds its own runtime
+            // by `chain.proposal_deadline_pct`, so this wait is bounded.
+            self.chain.submitBlockBuildOnInterval(slot, slot_proposer_id);
+            self.chain.thread_pool.waitAndWork(&self.chain.proposal_build_wg);
```
ch4r10t33r, line 146: **Critical.** Submit then immediately `waitAndWork` is synchronous — same wall clock as the old `produceBlock`. The PR description says interval-aware build/finalize split, but both happen in one tick here. To actually get the claimed semantics, call `submitBlockBuildOnInterval` from an interval-0 callback and only call `finalizeProposalIfReady` at interval-1. Otherwise drop the split, the `proposal_deadline_pct` knob, and the `PartialProposalBody` plumbing.

pkgs/node/src/forkchoice.zig
```diff
@@ -2130,7 +2151,12 @@ pub const ForkChoice = struct {
 
         const compute_start_ns = zeam_utils.monotonicTimestampNs();
 
-        for (att_data_keys) |data| {
+        // Aggregate only the single AttestationData most likely to advance
+        // `fcStore.latest_justified`. All gossip sigs and child payloads
+        // (new + known) registered against the chosen key are still folded
+        // into the aggregate; only the choice of key is narrowed.
+        const latest_justified = self.anchorState.latest_justified;
```
ch4r10t33r: **Critical.** `anchorState` is fixed at fork-choice init — its `latest_justified` is the genesis/checkpoint value and never advances. On a long-running aggregator this stays at slot 0 forever, so the preferred-tier match never moves and the fallback tier always wins. Use `self.fcStore.latest_justified` instead.

pkgs/node/src/forkchoice.zig
```diff
@@ -2130,7 +2151,12 @@ pub const ForkChoice = struct {
...
+        if (selectJustificationAdvancingKey(&snap, att_data_keys, latest_justified)) |data| {
```
ch4r10t33r, line 2159: **Critical.** Cuts aggregator from N keys/tick to 1 key/tick. If real workload has multiple distinct `AttestationData` per slot (cross-subnet, late attestations), the unpicked keys never get aggregated — tiebreak biases toward the same key, no round-robin, no backlog drain. Quantify on devnet first; if N>1 keys/slot are legitimate, this is throughput regression hidden as latency win.

pkgs/node/src/network.zig
```diff
@@ -568,11 +568,6 @@ pub const Network = struct {
             }
             return err;
         };
-        // Cache took the inner SignedBlock value (struct copy). Free the
-        // outer heap pointer; the inner allocations are now owned by the
-        // cache and will be freed via `removeFetchedBlock` /
-        // `BlockCache.deinit`.
-        self.allocator.destroy(block_ptr);
     }
```
ch4r10t33r, line 571: **Critical.** The PR removes the `self.allocator.destroy(block_ptr)` from the success path (with its explanatory comment) but the diff has no caller-side change. On success the caller's heap pointer is now never freed. Either restore the destroy, or update the caller to own the free and explain it in the commit message.

pkgs/network/src/interface.zig
```diff
@@ -843,23 +861,43 @@ pub const PeerEventHandler = struct {
...
     }
 
     pub fn subscribe(self: *Self, handler: OnPeerEventCbHandler) !void {
+        self.mutex.lock();
+        defer self.mutex.unlock();
         try self.handlers.append(self.allocator, handler);
+        // Replay peers that connected before this handler subscribed.
+        var it = self.connected.iterator();
+        while (it.next()) |entry| {
+            handler.onPeerConnected(entry.key_ptr.*, entry.value_ptr.*) catch |e| {
```
ch4r10t33r: **High.** `handler.onPeerConnected(...)` runs while still holding `self.mutex`. If the handler ever calls back into `PeerEventHandler` (or schedules work that does), recursive lock attempt → deadlock. Snapshot `self.connected` under the lock, release, then replay.

pkgs/node/src/chain.zig
```diff
@@ -4628,6 +4650,303 @@ pub const BeamChain = struct {
...
+        // Re-acquire pre-state from the state store (immutable per root).
+        var pre_borrow = self.statesGet(body.parent_root) orelse {
+            self.logger.warn(
+                "finalize failed for slot={d}: missing pre-state for parent_root",
+                .{slot},
+            );
+            return null;
+        };
+        defer pre_borrow.assertReleasedOrPanic();
+        const pre_snapshot = try pre_borrow.cloneAndRelease(self.allocator);
```
ch4r10t33r: **High.** `pre_snapshot` was already cloned in the build worker. Cloning a `BeamState` is an MB-scale SSZ round-trip — paying that twice per proposal. Carry `pre_snapshot` (or the computed `post_state`) through `PartialProposalBody` instead.

ch4r10t33r (dismissed):
6 inline blockers — 4 critical, 2 high. PR bundles too many concerns; consider splitting (rust libp2p panic fix, `is_aggregator` propagation, and peer-event replay are independently mergeable).

pkgs/node/src/network.zig
anshalshukla: This was changed in the clone PR itself but was missed here earlier which leads to panic due to double free

pkgs/node/src/forkchoice.zig
anshalshukla: `selectJustificationAdvancingKey` priorities the attestation_data that can help move current justified and finalized instead of a naive aggregation of everything in store. It restricts to a single aggregation per interval as the validator votes continue to change and the most favorable attestation_data should be prioritized. 

We can continue to make this smarter with additional attestation_data that can build upon the justification movement based on the aggregation earlier but it is not super critcal

pkgs/node/src/validator_client.zig
anshalshukla: proposal_deadline_pct (default 90%) is to imply that the block building and aggregation stuff should happen within this time period of the proposal interval (720ms) and rest 80% is for state transition and proposal signature over the full/partial block. 

I think the proposal_deadline_pct is little misleading and I'll rename it

pkgs/node/src/chain.zig
anshalshukla: fixed, although it is no longer ssz round trip and significantly faster

g11tech (approved):

pkgs/node/src/forkchoice.zig
g11tech: the right solution is to configure a cli flag with default value of 1 which is the aggregator data concurrency
pls do a a followup PR for the same

build.zig
```diff
@@ -85,6 +85,9 @@ pub fn build(b: *Builder) !void {
     // LTO option (disabled by default for faster builds)
     const enable_lto = b.option(bool, "lto", "Enable Link Time Optimization (slower builds, smaller binaries)") orelse false;
 
+    // ThreadSanitizer for the `zeam` cli exe (debug data-race hunting, e.g. node3 sync).
+    const enable_tsan = b.option(bool, "tsan", "Build the zeam cli with ThreadSanitizer") orelse false;
```
anshalshukla, line 89: @ch4r10t33r here

PR #959, opened 2026-05-31
node, libp2p-glue: downgrade per-chunk reqresp + per-attestation gossip logs to debug

ch4r10t33r:
## Summary

Three log statements that fire many times per slot at info level were pushing useful info-level signal off the screen on busy devnet hosts. All three are routine per-event tracing — valuable for debugging but not for normal operation. Moved to debug:

- \`rust/libp2p-glue/src/lib.rs:3120\` — \`[reqresp] Sent response chunk on channel ...\` (fires once per response chunk; a single \`blocks_by_range\` reply can emit dozens)
- \`rust/libp2p-glue/src/lib.rs:3131\` — \`[reqresp] Sent end-of-stream on channel ...\` (paired with the chunk log; once per response stream)
- \`pkgs/node/src/node.zig:463\` — \`received gossip attestation for slot=... validator=...\` (once per attestation per validator per slot — the loudest line in node logs)

No behaviour change. Paired aggregation / published / processed lines were intentionally left at info per scope clarification.

## Test plan

- [x] \`zig build test\` — green
- [x] \`cargo build -p libp2p-glue\` — green
- [ ] Devnet smoke after deploy — confirm logs no longer flood with these three lines at info level

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -3211,7 +3211,7 @@ pub const BeamChain = struct {
                         return err;
                     };
                 }
-                self.logger.info("processed gossip attestation for slot={d} validator={d}{f}", .{
+                self.logger.debug("processed gossip attestation for slot={d} validator={d}{f}", .{
                     slot,
                     validator_id,
                     validator_node_name,
--- a/pkgs/node/src/node.zig
+++ b/pkgs/node/src/node.zig
@@ -460,7 +460,7 @@ pub const BeamNode = struct {
                 const validator_node_name = self.node_registry.getNodeNameFromValidatorIndex(validator_id);
 
                 const sender_node_name = self.node_registry.getNodeNameFromPeerId(sender_peer_id);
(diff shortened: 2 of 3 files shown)


zclawz (dismissed):
Reviewed. This is a narrow logging-level-only change: the noisy per-attestation gossip line and per-response-chunk / EOS reqresp lines move from info to debug, with no control-flow or state changes. That matches the PR intent and preserves error/warn signal for actual send failures.

Lint is already green on this head; the broader build/test matrix is still running as I write this, so please wait for CI before merge. No code-level blockers from me.

zclawz (approved):
Re-reviewed latest head `eec71aa`. The new commit is consistent with the PR’s scope: it also downgrades the matching per-attestation chain “processed gossip attestation” line from info to debug. Still logging-only, no behavior/state change.

Lint is green and the macOS build has passed; the remaining matrix jobs are still running as I write this, so wait for full CI green before merge. No code-level blockers from me.

Issue #723, opened 2026-04-13
snappyz decode fails with TooLarge error on block gossip topic


ch4r10t33r:
## Description

During the devnet4 run (Apr 13), zeam fails to decode a block received via gossipsub on topic `/leanconsensus/devnet0/block/ssz_snappy` with a `TooLarge` error from the snappyz decoder.

## Logs

```
Apr-13 21:08:09.397 [s=0 i=0] [error] (zeam): [network] Error in snappyz decoding the message for topic=/leanconsensus/devnet0/block/ssz_snappy: error.TooLarge
Apr-13 21:08:09.403 [s=0 i=0] [warning] (zeam): [network] SSZ deserialization failed for snappyz_decode message - written 9374750 bytes to debug file: deserialization_dumps/failed_snappyz_decode_1776114489.bin
Apr-13 21:08:09.403 [s=0 i=0] [error] (zeam): [network] Snappyz decode failed - debug file created
```

## Context

- The decoded payload is ~9.37 MB, which likely exceeds the snappyz max decode buffer
- This was a block message from another client on the devnet4 network (slot ~174)
- A debug dump was written to `deserialization_dumps/failed_snappyz_decode_1776114489.bin`

## Expected behavior

Zeam should be able to decode valid snappy-compressed blocks from other clients. The max decode size limit may need to be increased or made configurable to handle large blocks (e.g. blocks with many attestations or large post-quantum signature payloads).

PR #961, opened 2026-05-31
network: bump MAX_RPC_MESSAGE_SIZE 4 MB → 50 MB to match gossip cap (fixes non-finalising devnet)

ch4r10t33r:
## Summary

Devnet finalisation is currently broken: every \`blocks_by_range\` response that carries a real block fails inside \`buildResponseFrame\` with \`error.PayloadTooLarge\`, observed at **~1300 rejects per zeam node in 17 min** across **all peer kinds** (zeam, gean, qlean, grandine — so this is not peer-specific). Without RPC catch-up, fork choice diverges into many parallel branches and no validator majority accumulates → **finalized stays at slot 0**.

### Devnet evidence (zeam_0, 17 min uptime)

| Peer client | \`PayloadTooLarge\` rejects |
|-|-|
| zeam | **783** |
| gean | 501 |
| qlean | 99 |
| grandine | 24 |

Fork-choice tree at slot 246 shows 200+ nodes across 4+ branches at every depth; head=246, finalized=0.

## Root cause

\`MAX_RPC_MESSAGE_SIZE\` was 4 MB while \`MAX_GOSSIP_BLOCK_SIZE\` is 50 MB. \`blocks_by_range\` and \`blocks_by_root\` carry the same XMSS-heavy blocks as the gossip block topic, so the RPC cap must accommodate the same per-block payload. The doc on \`MAX_GOSSIP_BLOCK_SIZE\` already notes devnet4 reached ~9.37 MB blocks (issue #723); the matching RPC cap was missed when the gossip cap was bumped.

## Fix

One-constant bump: \`MAX_RPC_MESSAGE_SIZE\` 4 MB → 50 MB, matching \`MAX_GOSSIP_BLOCK_SIZE\`. The same blocks flow through both transports — they share the same cap.

The pre-existing PR #855 review #9 TODO (split into per-protocol caps so Status / Goodbye / attestations can stay tighter) is preserved verbatim and can land separately.

## Test plan

- [x] \`zig build test\` green (existing \`validateRpcSnappyHeader rejects oversized declared size\` test still works — uses \`MAX_RPC_MESSAGE_SIZE + 1\` symbolically)
- [ ] Deploy to devnet; expect \`PayloadTooLarge\` rejects to drop to ~0 and finalised slot to advance past 0 within a few minutes

### DIFF
--- a/pkgs/network/src/ethlibp2p.zig
+++ b/pkgs/network/src/ethlibp2p.zig
@@ -31,10 +31,22 @@ const ServerStreamError = error{
     InvalidResponseVariant,
 };
 
-/// General RPC message size limit (4 MB). Used for req/resp protocol messages
-/// (BlocksByRoot, Status, etc.) and as a baseline gossip limit for small messages
-/// such as attestations and aggregations.
-const MAX_RPC_MESSAGE_SIZE: usize = 4 * 1024 * 1024;
+/// General RPC message size limit. Used for req/resp protocol messages
+/// (BlocksByRoot, BlocksByRange, Status, etc.) and as a baseline gossip limit
+/// for small messages such as attestations and aggregations.
+///
+/// Sized to match `MAX_GOSSIP_BLOCK_SIZE`: `blocks_by_range` and
+/// `blocks_by_root` carry the same XMSS-heavy blocks as the gossip block
+/// topic, so the RPC cap must accommodate the same per-block payload. A
+/// lower RPC cap (was 4 MB before #960) caused `error.PayloadTooLarge`
+/// during `buildResponseFrame` for every legitimately-sized block that
+/// fit in gossip but not in the RPC frame check, breaking blocks_by_range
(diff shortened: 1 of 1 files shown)


zclawz (approved):
Reviewed. The change is intentionally narrow: bumping `MAX_RPC_MESSAGE_SIZE` to match the existing 50 MB gossip block cap so block-carrying req/resp paths (`blocks_by_root` / `blocks_by_range`) can transport the same XMSS-heavy blocks that gossip already accepts.

The pre-decode/frame guards remain in place and the oversized-header test is symbolic (`MAX_RPC_MESSAGE_SIZE + 1`), so it continues to cover the boundary after the constant bump. The per-protocol tighter-cap TODO is preserved for a follow-up; for the current finalisation-blocking devnet issue, sharing the block-sized cap across RPC is the right emergency fix.

Lint is green; the broader build/test matrix is still running as I write this, so please wait for full CI green before merge. No code-level blockers from me.

PR #964, opened 2026-06-01
Report peers causing behind sync status

zclawz:
## Summary
- extend `BeamChain.SyncStatus.behind_peers` with a bounded snapshot of peers whose finalized slots make the node report `behind_peers`
- keep returning `max_peer_finalized_slot` and include `behind_peers=[peer(finalized_slot=...)]` in all behind-peers duty/aggregation skip logs
- copy peer ids into fixed-size status storage so callers do not borrow connected-peer map memory after releasing its lock

## Validation
- `zig fmt pkgs/node/src/chain.zig pkgs/node/src/validator_client.zig`
- `zig ast-check pkgs/node/src/chain.zig && zig ast-check pkgs/node/src/validator_client.zig`
- `/tmp/zig-x86_64-linux-0.16.0/zig build test -Dprover=dummy` is running locally; earlier run got through compile after the modified code, then failed only on the now-fixed Zig 0.16 `std.io` API use.

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -2352,7 +2352,7 @@ pub const BeamChain = struct {
         {
             const sync_status = self.getSyncStatus();
             zeam_metrics.metrics.lean_node_sync_status.set(.{ .status = "idle" }, if (sync_status == .no_peers or sync_status == .fc_initing) 1 else 0) catch {};
-            zeam_metrics.metrics.lean_node_sync_status.set(.{ .status = "syncing" }, if (sync_status == .behind_peers) 1 else 0) catch {};
+            zeam_metrics.metrics.lean_node_sync_status.set(.{ .status = "syncing" }, if (sync_status == .peers_materially_ahead) 1 else 0) catch {};
             zeam_metrics.metrics.lean_node_sync_status.set(.{ .status = "synced" }, if (sync_status == .synced) 1 else 0) catch {};
         }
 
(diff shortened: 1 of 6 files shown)


pkgs/node/src/chain.zig
```diff
@@ -4481,11 +4481,14 @@ pub const BeamChain = struct {
                 self.logger.info("aggregating for slot={d} with no peers (local only)", .{slot});
             },
             .behind_peers => |info| {
-                self.logger.warn("skipping aggregation production for slot={d}: behind peers (head_slot={d}, finalized_slot={d}, max_peer_finalized_slot={d})", .{
+                var behind_peers_buf: [1024]u8 = undefined;
+                const behind_peers = info.peerListForLog(&behind_peers_buf);
+                self.logger.warn("skipping aggregation production for slot={d}: behind peers (head_slot={d}, finalized_slot={d}, max_peer_finalized_slot={d}, behind_peers={s})", .{
```
g11tech: @zclawz rather than forming a big behind_peers_buf, log the peer using logger.debug for each of the behind_peers from the BehindPeerInfo, separate debug line for each of the peer's and also use node_registery to lookup getNodeNameFromPeerId and add their names in that debug log

and here in the warn log, just log the count for behind_peers

zclawz: Addressed in eb26e80:

- removed the aggregate `behind_peers=[...]` warn payload
- warn logs now include only `behind_peer_count` alongside `max_peer_finalized_slot`
- added one debug log line per behind peer, including `getNodeNameFromPeerId(...)` output and that peer's finalized slot
- applied the same pattern to aggregation, block-production, and attestation behind-peers logs

Validation so far:
- `zig fmt pkgs/node/src/chain.zig pkgs/node/src/validator_client.zig`
- `/tmp/zig-x86_64-linux-0.16.0/zig ast-check pkgs/node/src/chain.zig`
- `/tmp/zig-x86_64-linux-0.16.0/zig ast-check pkgs/node/src/validator_client.zig`
- `/tmp/zig-x86_64-linux-0.16.0/zig build test -Dprover=dummy` is still running locally.

pkgs/node/src/chain.zig
```diff
@@ -4901,14 +4951,24 @@ pub const BeamChain = struct {
...
+        if (behind_peer_finalization) {
+            var info = self.behindPeersStatus(our_head_slot, our_finalized_slot, max_peer_finalized_slot);
+            var cause_guard = self.connected_peers.iterateLocked();
+            defer cause_guard.deinit();
+            var cause_iter = cause_guard.iter;
+            while (cause_iter.next()) |entry| {
+                const peer_info = entry.value_ptr;
+                if (peer_info.latest_status) |status| {
+                    if (status.finalized_slot > our_head_slot or status.finalized_slot > our_finalized_slot) {
+                        info.behind_peers.addPeer(peer_info.peer_id, status.finalized_slot);
```
g11tech: @zclawz 
instead of naming it behind_peers, name it peers_materially_ahead to correctly name it

zclawz: Done — renamed to `peers_materially_ahead` throughout.

g11tech (approved):

PR #966, opened 2026-06-01
perf + diag: batched XMSS verify across blocks + #942 peer_id attribution

ch4r10t33r:
## Summary

Four commits, one PR. End-to-end batched-XMSS-verify-across-blocks (stages 1–3 + metric rename) plus a small diagnostic fix that surfaces the real peer_id on #942 gossip-decode failures (cherry-picked from #967, which is now closed).

## Commits

| # | Commit | Layer |
|-|-|-|
| 1 | \`a9c3853c\` | \`stf.verifySignaturesParallelMultiBlock\` API + batch-size histogram + 3 new stage labels |
| 2 | \`248d6aa4\` | \`chain.onBlocksBatch\` + \`preVerified\` opt on \`onBlock\` |
| 3 | \`487c9081\` | Chain-worker drain coalesces ≥2 \`.on_block\` messages → batched dispatch |
| 4 | \`d9ad5f4d\` | metric rename: \`lean_chain_worker_block_dispatch_total\` → \`zeam_chain_worker_block_dispatch_total\` |
| 5 | \`57c4bae2\` | libp2p-glue: log \`propagation_source\` instead of \`"unknown_peer"\` on gossip decode failures (#942) |

### Devnet data motivating this (PR #963 deployed image)

| Stage inside \`step="verify_signatures"\` | Mean per block | Share |
|-|-|-|
| **\`phase2_batch_verify\` (rayon XMSS)** | **147 ms** | **99.0%** |
| \`proposer_xmss_verify\` | 1.3 ms | 0.9% |
| \`phase1_prep\` | 0.1 ms | 0.07% |
| \`proposer_block_root\` (after #963 \`e795c0a3\` fix) | 3.6 μs | 0.002% |

A 4-block sync sweep is 4 × ~150 ms ≈ 600 ms of wall-clock verify work today, all of it inside one block at a time. Batching means **one rayon call** across every attestation in all 4 blocks — same parallel pool, 4× the concurrent work to schedule.

### Stage 1 — new \`stf\` API (commit 1)

\`\`\`zig
pub const VerifyMultiBlockTask = struct {
    signed_block: *const types.SignedBlock,
    precomputed_block_root: *const [32]u8,
};

pub fn verifySignaturesParallelMultiBlock(
    allocator: Allocator,
    tasks: []const VerifyMultiBlockTask,
    pubkey_cache: ?*xmss.PublicKeyCache,
    thread_pool: anytype,
) !void
\`\`\`

- Phase 1 (serial): per-block pubkey-cache warm + index check + message-hash compute, accumulating tasks into one batch.
- Phase 2 (parallel): **ONE** rayon call across the combined batch.
- Proposer signatures: K serial \`xmss.verifySsz\` (single XMSS each — too small to batch).
- K=1 fast-path delegates to the existing \`verifySignaturesParallel\` so single-block callers (gossip, locally-produced blocks) keep identical metric labels and behaviour.
- \`precomputed_block_root\` REQUIRED on the multi-block path so we don't re-introduce the duplicated \`hashTreeRoot(BeamBlock)\` that PR #963's \`e795c0a3\` fix eliminated.

New metrics:
- \`zeam_stf_verify_signatures_stage_duration_seconds{stage="phase1_prep_multiblock" | "phase2_batch_verify_multiblock" | "proposer_xmss_verify_multiblock"}\`
- \`zeam_stf_verify_signatures_batch_size\` (buckets 1, 2, 4, 8, 16, 32)

### Stage 2 — \`chain.onBlocksBatch\` (commit 2)

\`\`\`zig
pub const BatchBlockResult = struct { missing_roots: ?[]types.Root };

pub fn onBlocksBatch(
    self: *Self,
    signedBlocks: []const types.SignedBlock,
    block_roots: []const types.Root,
    pruneForkchoice: bool,
) ![]BatchBlockResult
\`\`\`

1. Look up FIRST block's parent state (sufficient — lean consensus validator set is fixed at genesis, so pubkeys are identical across all parent states).
2. Build K \`VerifyMultiBlockTask\` entries.
3. ONE \`stf.verifySignaturesParallelMultiBlock\` call.
4. Per-block sequential STF via the regular \`onBlock\` path with new opts flag \`preVerified = true\` — skips the duplicated verify call inside \`onBlock\` while every other invariant (STF, fork choice, persistence) stays unchanged.

Failure handling: if the batched verify returns \`InvalidBlockSignatures\`, falls back to per-block \`onBlock(.{ preVerified = false })\` so valid blocks still import and only the bad block fails individually.

K=0 → empty slice. K=1 → routes through \`onBlock\` for identical semantics.

\`onBlock\` opts addition: \`preVerified: bool = false\` on \`CachedProcessedBlockInfo\`. Set ONLY by \`onBlocksBatch\`; bare callers leave it false.

### Stage 3 — chain-worker drain (commit 3)

\`\`\`
if (block_queue.outstandingDepth() > BLOCK_BATCH_THRESHOLD) {
    collect up to BLOCK_BATCH_MAX .on_block messages
    dispatch via on_blocks_batch vtable        ← NEW
} else {
    one .on_block via on_block vtable           ← existing single-block path
}
\`\`\`

| Constant | Value | Why |
|-|-|-|
| \`BLOCK_BATCH_THRESHOLD\` | **2** | Below this, latency-sensitive gossip / locally-produced blocks keep going through the single-block path. |
| \`BLOCK_BATCH_MAX\` | **4** | Matches \`BLOCKS_BY_RANGE_SYNC_THRESHOLD\` — a single catch-up sweep fills the batch exactly. |

Non-\`.on_block\` variants encountered mid-collection flush the partial batch and route through single-shot dispatch for ordering fidelity. \`markProcessed()\` called once per popped message.

New thunk \`chainWorkerOnBlocksBatchThunk\`: calls \`chain.onBlocksBatch\`, runs per-block bookkeeping (\`onBlockFollowup\` + \`replayPendingAttestations\` + \`imported_block\` backchannel) once per block in iteration order.

New counter: \`zeam_chain_worker_block_dispatch_total{path="single" | "batch"}\` — proves whether batching engaged.

### Stage 4 — metric prefix correction (commit 4)

Renamed from \`lean_chain_worker_block_dispatch_total\` to \`zeam_chain_worker_block_dispatch_total\`. \`lean_*\` is reserved for protocol-shape metrics; this counter measures a zeam implementation detail (drain-loop dispatch decision) with no analogue in other lean clients.

### Known trade-off in the batched dispatch path

TOCTOU rejection paths (\`MissingPreState\` / \`PreFinalizedSlot\`) that the single-block thunk routes to the \`rejected_block\` callback are NOT exercised from the batched thunk: \`onBlocksBatch\` swallows per-block failures into \`missing_roots = null\` so we lose the error tag. Acceptable because:
1. The TOCTOU window is narrow (the producer just checked \`hasBlock(parent)\` under a shared lock).
2. Gossip-rate blocks at queue depth ≤ threshold still go through the single-block thunk with full TOCTOU.
3. Catch-up blocks (where this PR helps most) rarely hit TOCTOU.

## Part 2 — #942 peer_id attribution (commit 5)

Pulled in from PR #967 (now closed) so it ships alongside this work.

The decode-failure log at \`pkgs/network/src/ethlibp2p.zig:638\` prints \`from peer={sender_peer_id_slice}\` — and that slice has been uniformly \`"unknown_peer"\` because of this on the Rust side (\`rust/libp2p-glue/src/lib.rs:2761\`):

\`\`\`rust
let sender_peer_id_string = message.source
    .map(|p| p.to_string())
    .unwrap_or_else(|| "unknown_peer".to_string());
\`\`\`

\`message.source\` is \`Some(PeerId)\` only when the gossip message is signed by its original publisher. zeam runs gossipsub with \`ValidationMode::Anonymous\`, which disables signed messages — so \`message.source\` is ALWAYS \`None\`, and the literal "unknown_peer" string ended up in every \`error.Corrupt\` log entry.

That destroyed our ability to attribute byte-corrupt arrivals to a specific upstream peer. Today's investigation revealed:
- ~14 corrupt/min/node, fleet-wide
- **Byte-identical corrupted payloads on every zeam node** (same SHA across 4 nodes for the same timestamp) — not local buffer aliasing, deterministic from somewhere upstream
- Python reference snappy ALSO rejects the bytes → genuinely malformed snappy, not a zig-snappy edge case
- But the network keeps progressing → other clients accept these bytes → some producer is emitting non-conformant snappy that lenient decoders accept

### Fix

When \`message.source\` is \`None\`, fall back to \`propagation_source\` — the peer who handed us these bytes on this hop. \`propagation_source\` is always populated by libp2p-gossipsub (it's the immediate forwarder), so we always have a real peer_id (\`"16Uiu2HAm..."\` form) in the log now.

Trade-off: \`propagation_source\` is the FORWARDER, not the original producer (the publisher's identity is genuinely unrecoverable under Anonymous mode). But every byte-identical corrupt arrival traces back through some forwarding hop, and now we can score / disconnect / report that specific peer instead of guessing "unknown_peer".

## Expected impact

| Workload | Before | After (K=4) |
|-|-|-|
| 4-block sync sweep verify wall-clock | ~600 ms (4 × 150 ms) | **~150 ms** (one rayon call) |
| Catch-up rate per peer-sweep | ~7 blocks/s | **~25-30 blocks/s** |
| zeam_15-class 100-slot deficit closing time | ~30 s | **~5 s** |
| #942 corruption attribution | "from peer=unknown_peer" → useless | real peer_id → traceable to a specific client |

## Test plan

- [x] \`zig build test\` — all 350+ tests across 14 groups pass
- [x] \`cargo build -p libp2p-glue\` green for both changes
- [ ] Deploy with a fresh image; expect:
  - \`zeam_chain_worker_block_dispatch_total{path="batch"}\` to be non-zero during catch-up
  - \`zeam_stf_verify_signatures_batch_size_bucket{le="4"}\` to dominate over \`le="1"\` during catch-up
  - \`zeam_stf_verify_signatures_stage_duration_seconds_sum{stage="phase2_batch_verify_multiblock"}\` substantially less than 4× the single-block stage sum
  - lagging-node behind delta to close roughly 4× faster than today
  - \`error.Corrupt\` log lines now showing a real \`peer=16Uiu2HAm...\` for #942 attribution

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,12 +79,26 @@ COPY .git/HEAD .git/HEAD
 COPY .git/refs .git/refs
 
 # Get git commit hash and build the project with optimizations
-# Use cache mount for Zig build cache as well
+# Use cache mounts for both the Zig and Cargo build caches.
+#
+# Without the three `/root/.cargo/{registry,git}` and `/app/rust/target`
+# mounts, every Docker build (even for a 1-line Rust change) re-downloads
+# the full Cargo crate graph and re-compiles the ~200 transitive deps
+# (libp2p, gossipsub, quinn, tokio, ...) from scratch. On Mac/Docker
+# Desktop that's 5-15 minutes of pure Cargo compile time burned per
+# build. With these mounts: ~10s incremental rebuilds for typical changes.
+#
+# `rust/target` is where `cargo -C rust build` writes artifacts (see
+# build.zig: `cargo_build = b.addSystemCommand({..., "-C", "rust", ...})`).
+#
 # CPU targeting per architecture:
 #   amd64: x86-64-v3 (AVX2, no AVX512) — portable across modern x86_64 servers
 #   arm64: generic CPU — QEMU-compatible when cross-building
 ARG TARGETARCH
 RUN --mount=type=cache,target=/root/.cache/zig \
+    --mount=type=cache,target=/root/.cargo/registry \
+    --mount=type=cache,target=/root/.cargo/git \
+    --mount=type=cache,target=/app/rust/target \
     EXTRA_ZIG_FLAGS="" && \
     if [ "$TARGETARCH" = "amd64" ]; then \
         EXTRA_ZIG_FLAGS="-Dcpu=x86_64_v3 -Drust-target-cpu=x86-64-v3"; \
@@ -98,14 +112,20 @@ RUN --mount=type=cache,target=/root/.cache/zig \
     else \
         GIT_VERSION=$(echo "$GIT_VERSION" | head -c 7); \
     fi && \
-    zig build --seed 0 -Doptimize=ReleaseSafe -Dgit_version="$GIT_VERSION" $EXTRA_ZIG_FLAGS
+    zig build --seed 0 -Doptimize=ReleaseSafe -Dgit_version="$GIT_VERSION" $EXTRA_ZIG_FLAGS && \
+    REC_AGG_DIR=$(find /root/.cargo/git/checkouts -path "*/rec_aggregation" -type d | head -1) && \
+    cp -r "$REC_AGG_DIR" /tmp/rec_aggregation_src && \
+    echo "$REC_AGG_DIR" > /tmp/rec_aggregation_path
 
-# rec_aggregation's compilation.rs reads .py source files at runtime to verify
-# a bytecode fingerprint (via env!("CARGO_MANIFEST_DIR") baked at compile time).
-# Stage the crate source so we can recreate the path in the runtime image.
-RUN REC_AGG_DIR=$(find /root/.cargo/git/checkouts -path "*/rec_aggregation" -type d | head -1) \
-    && cp -r "$REC_AGG_DIR" /tmp/rec_aggregation_src \
-    && echo "$REC_AGG_DIR" > /tmp/rec_aggregation_path
+# Note: rec_aggregation extraction is folded into the build RUN above
+# (not a separate RUN) because we cache-mount /root/.cargo/git on that
+# RUN — cache mounts are ephemeral per-RUN, so a subsequent layer cannot
+# see /root/.cargo/git/checkouts at all. /tmp/rec_aggregation_src lands
+# on the real image layer and is picked up by the runtime-prep stage.
+# rec_aggregation's compilation.rs reads .py source files at runtime to
+# verify a bytecode fingerprint (via env!("CARGO_MANIFEST_DIR") baked at
+# compile time), so the crate source has to be present at the same path
+# in the runtime image.
 
 # Intermediate stage to prepare runtime libraries
 FROM ubuntu:24.04 AS runtime-prep
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -339,7 +339,7 @@ const Metrics = struct {
     /// dispatch. Sum the latter with the K from
     /// `zeam_stf_verify_signatures_batch_size_count` to recover the
     /// total number of blocks that took the batched path.
-    lean_chain_worker_block_dispatch_total: LeanChainWorkerBlockDispatchCounter,
+    zeam_chain_worker_block_dispatch_total: ZeamChainWorkerBlockDispatchCounter,
     /// Tripwire counter (zclawz review on PR #890): bumped from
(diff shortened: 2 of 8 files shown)


g11tech (approved):
while I agree that devnet5 will make it redundant as @anshalshukla says, we need to pass devnet4 stage acceptability as of now, so approving it

PR #968, opened 2026-06-02
network: replace rust libp2p-glue with pure-zig zig-libp2p (QUIC)

ch4r10t33r:
## Summary

Replaces zeam's Rust `libp2p-glue` FFI stack with a pure-Zig networking path built on [`zig-libp2p`](https://github.com/blockblaz/zig-libp2p) **v0.1.97** (transitively **zquic v1.7.40**). `zeam node beam`, simtest, and the Docker devnet entry point now use `EthLibp2p` over QUIC + libp2p TLS.

Net change: **−~8,000 LOC** of Rust glue removed; **+~1,900 LOC** of native Zig networking (`ethlibp2p.zig`, `gossip_codec.zig`).

> **History note:** this branch was squashed from 66 incremental commits into a single commit for review. The per-release upstream history is preserved in the collapsible section below.

### Networking (`pkgs/network/src/ethlibp2p.zig`)

- **QUIC transport** via `zig-libp2p` `QuicRuntime`: listen/dial, multistream-select, gossipsub, req/resp.
- **Host identity**: secp256k1 private key read from `--node-key` (lean-quickstart `.key` format), matching `eth-beacon-genesis` ENR-derived `/p2p/...` peer ids.
- **In-memory TLS PEM** (no cert files on disk; works in `FROM scratch` containers).
- **Gossipsub**: SSZ + snappy encode/decode, publish/subscribe, `#942` snappy hardening reused from `gossip_codec.zig`.
- **Req/resp**: status, blocks_by_root, blocks_by_range — SSZ framing, callback dispatch, in-flight failure on peer disconnect.
- **Peer events**: connect/disconnect/dial-failure with base58 PeerIds matching the legacy Rust path.
- **Bootnodes**: registered as `host.registerKnownPeer` **and** `host.addDirectPeer` so they are always mesh-eligible and GRAFTed on the next gossipsub heartbeat instead of waiting for a SUBSCRIBE/GRAFT round trip.
- **Metrics parity**: `lean_gossip_mesh_peers`, `zeam_libp2p_swarm_command_dropped_total`.
- **Observability**: gossip / SUBSCRIBE / mesh-status logs land on the `debug` channel (only `registered bootnode` stays at `info`).

### Removed

- `rust/libp2p-glue/` (~5,700 LOC); the ~2,400 LOC legacy Rust-FFI `EthLibp2p` was rewritten in place as the native `pkgs/network/src/ethlibp2p.zig` above.
- `zeam-glue` libp2p feature flag from the Rust prover workspace (prover staticlibs unchanged).

### Other zeam fixes folded in on this branch

- **`node/chain`**: fix use-after-free in `produceBlock` / `finalizeProposalIfReady` (`wrapOwnedStateIntoRc` freed the `post_state` wrapper while `forkChoice.onBlock` still held the dangling pointer — segfault at first block proposal under ReleaseSafe).
- **`node`**: publish blocks *before* the redundant local `onBlock`; fix mesh-peer metric registration; fix gossip-source attribution (proposer-coverage no longer reports `gossip=none`).
- **`node`/`chain`**: aggregator `onInterval` signing offload + lock-scoping (fixes the aggregator stall); catch-up finalization-gap fix.
- **`cli`, `build`**: restore the mock simtest harness and params-scaled timeouts.
- **`leanSpec`**: submodule bumped for devnet5 SSZ fixtures.
- **Docker**: exclude `**/target/` from build context.

## Transport hardening (since v0.1.34)

The persistent-`/meshsub` instability that earlier showed up as "chain forks / wedge under mixed-client load" was **root-caused to the QUIC transport, not consensus**, and fixed in the dependency. Highlights of the v0.1.35 → v0.1.93 progression:

| Area | Fix | Tracking |
|------|-----|----------|
| **Per-stream receive flow control** | zquic never proactively emitted `MAX_STREAM_DATA`, so a long-lived `/meshsub` gossip stream stalled at the peer's initial 10 MiB send window → connection teardown → req/resp reset → **stuck-at-slot-48 wedge**. Now extends per-stream + conn-level windows proactively on both recv paths. | zquic#172/#173 (v1.7.38) |
| **cwnd collapse** | `onAck` exited congestion recovery on the first ACK and cleared the once-per-flight loss gate, so one loss episode cut cwnd to the floor (`cwnd=13500`, `cong_events` in the thousands) and throttled gossip → outbox-stuck redials. Fixed per RFC 9002 §7.3.2. | zquic#175 (v1.7.39) |
| **Spurious cwnd collapse on loopback** | local UDP `EWOULDBLOCK`/`ENOBUFS` drops were mis-counted as network loss; added bounded retry + `local_send_drops` instrumentation. | zquic v1.7.36 |
| **Req/resp half-close (FIN)** | go-libp2p responders read the request to EOF before replying; zeam never half-closed its send side, so status RPC to a go-libp2p peer timed out. Now sends a STREAM FIN after the request. | zig-libp2p v0.1.93 |
| **Gossip over any connection** | publish over inbound or outbound leg so a single-direction connection still propagates gossip. | zig-libp2p#214 (v0.1.90) |
| Log noise | per-tick CC backpressure trace demoted `warn` → `debug`. | zquic#174 |

<details>
<summary>Earlier upstream pins (zig-libp2p v0.1.11 → v0.1.34) — initial bring-up & multistream/req-resp parity</summary>

| Release | What zeam needed |
|---------|------------------|
| v0.1.11–v0.1.17 | zquic v1.6.13–v1.6.15; zig↔go QUIC ping; Initial retransmit replay; inbound gossip publish; cert UTCTime fix; raised wire caps for hash-sig blocks; TCP+TLS+Yamux interop (#186); persistent per-peer `/meshsub` stream (#189) |
| v0.1.18–v0.1.27 | delimited req/resp + `/ipfs/id` + `/ipfs/ping` inbound; persist multistream accumulator across drive ticks; framing-collision fixes; reply to `/ipfs/id/push`; release inbound req/resp slots after response; surface remote-initiated streams on outbound connections; drain GRAFT/PRUNE to mesh streams; accept `/meshsub/{1.0..1.3}.0` |
| v0.1.28–v0.1.34 | recreate wedged persistent gossip streams; single persistent `/meshsub` per peer (mark `broken` instead of opening a second stream); surface remote `CONNECTION_CLOSE` to `onConnectionClosed`; forward-compat unknown `SubOpts` protobuf fields; one-tick persistent handshake + eager first dial + `to=null` SUBSCRIBE broadcast fan-out |

</details>

## Dependency update: v0.1.93 → v0.1.97

The pin moved to **zig-libp2p v0.1.97** (transitively **zquic v1.7.40**, which fixes a `u5` 1-RTT PN-sweep integer-overflow panic). The intervening releases add libp2p surface that zeam does not consume yet but is now available: **mDNS** LAN discovery, **rendezvous** (`/rendezvous/1.0.0`), **gossipsub v1.1 peer scoring**, **Kademlia DHT** lifecycle + record validators (spec-compliant IPNS), **AutoNAT** active probing, **DCUtR** auto-trigger, **Circuit Relay v2** reservation refresh, and **Identify-Push** auto-trigger. The v0.1.96 release also reorganized the library layout (`core/` · `primitives/` · `protocols/` · `transport/`) — additive public API, so `@import("zig_libp2p")` is unchanged.

zeam-side changes for the bump:
- `pkgs/network/src/ethlibp2p.zig`: handle the `swarm.Event` variants added across the range (`peer_discovered`, `reachability_changed`, `dcutr_succeeded` / `dcutr_failed`, `autonat_probe_peer`, `relay_reservation`, `identify_push_peer`) as no-ops in the network event loop — zeam does not act on them yet (kept as an explicit list so a future variant still forces a compile-time review).
- Merged `main` and reconciled `node.fetchBlockByRoots` with #949 (which collapsed `fetchBlockByRoots` / `fetchBlockByRootsFromPeer` into one function taking a `preferred_peer` argument): the branch's one remaining 2-arg call in the locally-produced-block reprocess path now passes `null`.

Standalone `zig build` is clean.

### zeam ↔ ethlambda (rust-libp2p) — ✅ validated
16-minute fresh-genesis run on `lean-quickstart` `local-devnet`:

| Check | Result |
|-------|--------|
| QUIC/TLS handshake (both directions) | ✅ |
| Persistent `/meshsub` stream | ✅ stable for the connection lifetime |
| Cross-client gossip (blocks + attestations) | ✅ |
| Cross-client req/resp (status, blocks_by_*) | ✅ both directions |
| Long-lived gossip past 10 MiB / slot 48 | ✅ **no wedge** (per-stream flow-control fix) |
| cwnd / outbox health | ✅ no cwnd-floor stalls, zero outbox-stuck redials |
| **Chain progress + finality** | ✅ head 8 → 249, **finalized = 244**, head == wall-clock throughout |

### zeam ↔ gean (go-libp2p) — ✅ status RPC fixed
- Status req/resp `StreamTimedOut` → **fixed** by the req/resp half-close FIN; gean now replies (`received status response … head_slot=0`).
- 3-node finalization is **not yet validated** — see limitations.

## Known limitations (follow-up, not blockers for this PR)

- **3-node devnet is RAM-bound, not code-bound.** On an 8 GiB Docker/OrbStack VM, `zeam_0` alone holds ~4 GiB (~52%); with ethlambda + gean + bootnodes the VM oversubscribes and **gean gets OOM-killed** (its `--rm` container vanishes, hiding the exit). Run a 3-node devnet on a ≥16 GiB VM, or run 2 nodes (zeam + ethlambda, which finalizes cleanly). Environmental — no transport regression.
- **zeam steady-state footprint (~4 GiB)** is worth profiling (hash-sig/XMSS + chain state vs. a slow leak). Separate from this PR.
- **Hash-sig block size**: blocks reach ~3 MiB snappy-compressed; wire caps raised to 16–64 MiB headroom upstream.

## Test plan

- [x] `zig build simtest` — three nodes over real QUIC (all-to-all mesh, blocks propagate)
- [x] `EthLibp2p` unit tests (protocol mapping, peer events, bootnode parsing, QuicRuntime bring-up)
- [x] Docker devnet: zeam ↔ ethlambda peering, gossip, block production
- [x] Persistent `/meshsub` stream survives cross-client GRAFT/PRUNE under sustained load
- [x] **zeam ↔ ethlambda continuous finality** (head 8→249, finalized 244 — flow-control + cwnd-collapse fixes)
- [ ] **3-node cross-client finality** (zeam + ethlambda + gean) — gated on a ≥16 GiB devnet VM (gean OOMs under the constrained VM)


pkgs/cli/src/node.zig
```diff
@@ -376,15 +391,33 @@ pub const Node = struct {
...
+        self.listen_addresses_str = if (addresses.listen_addresses.len > 0)
+            try addresses.listen_addresses[0].toString(allocator)
+        else
+            try allocator.dupe(u8, "");
+        errdefer allocator.free(self.listen_addresses_str);
+        self.connect_peers_str = try joinMultiaddrsCsv(allocator, addresses.connect_peers);
+        errdefer allocator.free(self.connect_peers_str);
 
-        self.network = try networks.EthLibp2p.init(allocator, &self.loop, .{
+        self.network = try networks.EthLibp2pV2.init(allocator, &self.loop, .{
```
g11tech: why have a v2 version and file, why not replace it
or if we have v2, then keep both files if we wanna have option of switching the rustglue back

ch4r10t33r: Good call — went with "replace it". Since this PR deletes the Rust `libp2p-glue` path entirely, there's no v1 left for the `V2` suffix to disambiguate, and we don't want to keep the Rust glue switchable (the whole crate is removed). So I dropped the suffix rather than keep two files:

- `EthLibp2pV2` → `EthLibp2p`, `EthLibp2pV2Params` → `EthLibp2pParams`
- `ethlibp2p_v2.zig` → `ethlibp2p.zig` (the old file was deleted in this PR, so the name is free)
- scrubbed vestigial `v2` prefixes from log lines / doc comments, refreshed the module header (it still described a skeleton-alongside-Rust with QUIC "deferred"), and removed the now-complete `MIGRATION.md`

No behavior change; `zig build` + `zig build test` (220/220) pass. Done in 96e7113f.

pkgs/state-transition/src/transition.zig
```diff
@@ -379,6 +381,6 @@ fn logBeamStatePerFieldRoots(
     const fmt = "{s} per-field block_slot={d} state_slot={d} hbh_len={d} js_len={d} jr_len={d} | slot={x} lbh={x} lj={x} lf={x} hbh={x} js={x} v={x} jr={x} jv={x}";
     switch (level) {
         .err_level => logger.err(fmt, args),
-        .info_level => logger.info(fmt, args),
```
g11tech, line 382: why remove info level, it can still stay even if unused

g11tech (approved):
i am good with taking a risk and moving to zig libp2p
lets move that to blockblaz repo as well

g11tech (changes requested):
Image

beam run on the shadow server isn't finalizing and producing a fractured forkchoice
(which it doesn't on normal beam sim run on main )

pkgs/state-transition/src/transition.zig
ch4r10t33r: Restored `info_level` as a variant of `PerFieldLogLevel` with its `logger.info` switch arm in 744da88a. The PostStateMatch call site still passes `.debug_level` (keeps the per-field baseline out of info logs), but `.info_level` is back as an option for callers that want it surfaced — even if currently unused. Doc comment updated to match.

g11tech (approved):

PR #972, opened 2026-06-03
chore: bump leanMultisig multisig-glue dependency

zclawz:
## Summary
- bump `rust/multisig-glue` leanMultisig git dependencies from `5eba3b141455349d7cdbf0f5d3ccfb2e640b02aa` to `f66d4a974eced803574eb0ea43d812e523c8d7ad`
- refresh `rust/Cargo.lock` for the new leanMultisig/Plonky3 transitive pins
- adapt `multisig-glue` to upstream `AggregatedXMSS::{compress,decompress}` and fallible aggregation API
- log `xmss_aggregate` upstream `AggregationError`s before returning null to the Zig caller

## Compatibility / rollout note
This is **not** a rolling-compatible aggregate wire/proof-format bump.

The outer aggregate byte encoding is still postcard + lz4, but the payload/schema and proof internals changed between `5eba3b1` and `f66d4a9`:
- old `AggregatedXMSS`: `{ proof: Proof<F>, bytecode_point, metadata }`
- new `AggregatedXMSS`: `{ info: AggregatedXMSSInfo { without_pubkeys, pubkeys }, proof: ExecutionProof }`
- upstream also includes verifier/prover transcript and circuit changes such as Fiat-Shamir duplex sponge, bus/fingerprint changes, type-2 aggregation support, stricter decode/proof validation, and Plonky3 git rev movement (`82cfad73` -> `5ef05b2a`).

So old and new nodes should be treated as separate aggregate-protocol versions. Rollout plan: coordinated devnet cutover only; do not run a mixed aggregator fleet expecting cross-version aggregate reuse. Discard/let-prune any cached `latest_*_aggregated_payloads` from the old version during the upgrade rather than deserializing/reusing them after restart.

## Transitive lockfile audit
Expected from the upstream bump:
- leanMultisig crates move to `f66d4a9`
- Plonky3 crates remain `0.5.1` but move git rev `82cfad73` -> `5ef05b2a`
- new upstream helper deps: `include_dir`, `system-info`, `zk-alloc`, and macOS-only `objc2*` crates
- Cargo resolver also rewrites a few `windows-sys` lock references; no Linux code path impact expected

These can change prover bytes, so this PR should be merged only with the coordinated devnet/prover cutover above.

## Validation
- `cargo check -p multisig-glue`
- `cargo fmt --check`
- `zig build test` with Zig `0.16.0` (`/home/node/.local/zig/zig-x86_64-linux-0.16.0/zig`)

## Follow-up required before merge
- run hive/devnet validation for the coordinated aggregate-format cutover

### DIFF
--- a/rust/multisig-glue/src/lib.rs
+++ b/rust/multisig-glue/src/lib.rs
@@ -162,7 +162,7 @@ pub unsafe extern "C" fn xmss_aggregate(
                 return std::ptr::null();
             }
             let proof_bytes = slice::from_raw_parts(proof_ptrs[i], proof_lens[i]);
-            let proof = match AggregatedXMSS::deserialize(proof_bytes) {
+            let proof = match AggregatedXMSS::decompress(proof_bytes) {
                 Some(p) => p,
                 None => return std::ptr::null(),
             };
@@ -180,13 +180,16 @@ pub unsafe extern "C" fn xmss_aggregate(
     let t_marshal_done = Instant::now();
 
     // Call rec_aggregation
-    let (_pub_keys, agg_sig) = rec_xmss_aggregate(
+    let (_pub_keys, agg_sig) = match rec_xmss_aggregate(
         &children_with_keys,
         raw_xmss,
         message_hash,
         slot,
         log_inv_rate,
-    );
+    ) {
+        Ok(result) => result,
+        Err(_) => return std::ptr::null(),
+    };
 
     let t_stark_done = Instant::now();
 
@@ -235,7 +238,7 @@ pub unsafe extern "C" fn xmss_verify_aggregated(
 
     // Deserialize aggregate signature
     let bytes = slice::from_raw_parts(agg_sig_bytes, agg_sig_len);
-    let agg_sig = match AggregatedXMSS::deserialize(bytes) {
+    let agg_sig = match AggregatedXMSS::decompress(bytes) {
         Some(sig) => sig,
         None => return false,
     };
@@ -339,7 +342,7 @@ pub unsafe extern "C" fn xmss_verify_aggregated_batch(
 
     tasks.par_iter().all(|task| {
         let sig_bytes = slice::from_raw_parts(task.sig_ptr as *const u8, task.sig_len);
-        let agg_sig = match AggregatedXMSS::deserialize(sig_bytes) {
+        let agg_sig = match AggregatedXMSS::decompress(sig_bytes) {
             Some(sig) => sig,
             None => return false,
         };
@@ -374,7 +377,7 @@ pub unsafe extern "C" fn xmss_aggregate_signature_to_bytes(
     }
 
     let agg_sig_ref = &*agg_sig;
-    let serialized = agg_sig_ref.serialize();
+    let serialized = agg_sig_ref.compress();
 
     if serialized.len() > buffer_len {
         return 0;
@@ -403,7 +406,7 @@ pub unsafe extern "C" fn xmss_aggregate_signature_from_bytes(
 
     let input_slice = slice::from_raw_parts(bytes, bytes_len);
 
-    match AggregatedXMSS::deserialize(input_slice) {
+    match AggregatedXMSS::decompress(input_slice) {
         Some(agg_sig) => Box::into_raw(Box::new(agg_sig)),
         None => std::ptr::null_mut(),
     }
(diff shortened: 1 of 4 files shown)


rust/multisig-glue/src/lib.rs
```diff
@@ -162,7 +162,7 @@ pub unsafe extern "C" fn xmss_aggregate(
                 return std::ptr::null();
             }
             let proof_bytes = slice::from_raw_parts(proof_ptrs[i], proof_lens[i]);
-            let proof = match AggregatedXMSS::deserialize(proof_bytes) {
+            let proof = match AggregatedXMSS::decompress(proof_bytes) {
```
ch4r10t33r, line 165: **Critical.** `serialize/deserialize` → `compress/decompress` strongly suggests upstream changed encoding (lz4? bincode variant?). If bytes differ, nodes on this PR cannot decode aggregates from peers still on 5eba3b1 (and vice versa). Devnet aggregator state cached in `latest_known_aggregated_payloads` becomes unreadable after restart. Confirm encoding is byte-compatible, OR call out the network split + state-migration plan in the PR description.

rust/multisig-glue/src/lib.rs
```diff
@@ -180,13 +180,16 @@ pub unsafe extern "C" fn xmss_aggregate(
...
+    let (_pub_keys, agg_sig) = match rec_xmss_aggregate(
         &children_with_keys,
         raw_xmss,
         message_hash,
         slot,
         log_inv_rate,
-    );
+    ) {
+        Ok(result) => result,
+        Err(_) => return std::ptr::null(),
```
ch4r10t33r: **High.** `rec_xmss_aggregate` is now fallible and we drop the `Err` discriminant. Operator sees the aggregator silently return null with no idea why. At minimum log the error before returning null.

ch4r10t33r (changes requested):
Mechanical dep bump, but the API rename suggests a wire-format change. PR description doesn't say WHAT changed between 5eba3b1 and f66d4a9 (bug fix? perf? format change?) — that's the load-bearing question. Also: aggregation now returns Result and the Err is discarded silently. Cargo.lock changes ~150 lines — audit transitive bumps (plonky3, sha3, p3_monty_31) for unexpected major/minor flips that could change prover bytes. Validation only ran `cargo check` + `cargo fmt` — `zig build test` + hive devnet must pass before merge; `cargo check` doesn't catch ABI mismatches with the Zig caller.

rust/multisig-glue/src/lib.rs
zclawz: The PR description's "Compatibility / rollout note" section covers this explicitly. Summary:

- The outer wire format (postcard + lz4) is unchanged — only the inner `AggregatedXMSS` struct changed.
- Old: `{ proof: Proof<F>, bytecode_point, metadata }`
- New: `{ info: AggregatedXMSSInfo { without_pubkeys, pubkeys }, proof: ExecutionProof }`

This is intentionally **not** a rolling-compatible bump — the PR description calls out the coordinated devnet cutover requirement and the instruction to discard cached `latest_*_aggregated_payloads` from the old version on restart. The `serialize`/`deserialize` → `compress`/`decompress` rename is a pure upstream API rename; the lz4 framing is preserved.

rust/multisig-glue/src/lib.rs
zclawz: The `Err` is not silently dropped in the current code — see the match arm in `xmss_aggregate`:

```rust
Err(err) => {
    eprintln!("xmss_aggregate failed: {err}");
    return std::ptr::null();
}
```

The error message is emitted to stderr before returning null, so the operator will see it in node logs.

ch4r10t33r (approved):
All three inline blockers addressed.

| # | Status |
|---|---|
| Wire-format break (line 165) | DOCUMENTED — "not rolling-compatible" + coordinated cutover plan + drop cached aggregates on upgrade |
| Err silently discarded (line 191) | FIXED — `eprintln!("xmss_aggregate failed: {err}")` before null return (`ffcda94`) |
| Bump rationale (Cargo.toml) | DOCUMENTED — schema change, Fiat-Shamir duplex sponge, type-2 aggregation, stricter decode, Plonky3 git rev `82cfad73 → 5ef05b2a` |
| Cargo.lock audit | DONE — plonky3 stays 0.5.1, no surprise major/minor flips |
| `zig build test` | Ran with Zig 0.16.0 |

CI: build + test + dummy-prove + build-all-provers PASS on both OS.

Approving the code change. **Do not merge until the hive devnet validation + coordinated aggregator cutover are ready** (already in PR's "Follow-up required before merge"). Mixed-version aggregator fleet will break aggregate reuse across nodes.

PR #976, opened 2026-06-05
cli: size worker pool by cgroup-aware cpu count

GrapeBaBa:
## Motivation

`std.Thread.getCpuCount()` only reads the CPU affinity mask, so under a cgroup CPU quota (`docker --cpus` / k8s `limits.cpu`) it reports the host core count rather than the quota, over-sizing the shared worker thread pool and causing CFS throttling in CPU-limited containers. (`node.zig` already has a comment acknowledging this exact issue near the `--rayon-threads` guard.)

## Changes

- Add `pkgs/cli/src/cpu_count.zig` — `getNumCpus(gpa, io)` = `min(cgroup quota, affinity)`, via `/proc/self/{cgroup,mountinfo}` + `cpu.max` / `cpu.cfs_quota_us` (cgroup v1 + v2). Ported from the `num_cpus` crate.
- Use it when sizing the worker pool in both paths: the simulation path (`main.zig`) and the single-node path (`node.zig`). Both pass the process `Io` (`init.io`); `Node.init` gains an `io` param so it uses the caller's io rather than a global/throwaway one.
- `/proc` files report size 0 → streamed to EOF (`readerStreaming` + `allocRemaining`); `ceilDiv` overflow-safe; non-Linux falls back to affinity.

## Testing

- 15 unit tests in `cpu_count.zig`; vectors adapted byte-for-byte from `seanmonstar/num_cpus` (MIT).
- `zig build install` builds the cli; OrbStack Linux container: `--cpus=2`→2, `--cpus=4`→4, unlimited→host count.

---
🤖 Developed with AI assistance (Claude Code).


g11tech (dismissed):

ch4r10t33r (commented):
Three medium notes inline — none blocking. The rayon thread sizing also benefits from this change (rayon_threads derives from desired_workers), which is worth calling out in the PR body since it's a bigger win than "worker pool sizing" alone.

pkgs/cli/src/cpu_count.zig
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    /// First cpu controller across all lines; a v1 entry takes precedence over v2.
+    fn load(content: []const u8) ?Subsys {
+        var result: ?Subsys = null;
+        var lines = std.mem.splitScalar(u8, content, '\n');
+        while (lines.next()) |raw| {
+            const line = std.mem.trimEnd(u8, raw, "\r");
+            const s = Subsys.parseLine(line) orelse continue;
+            if (result != null and s.version == .v2) continue; // v2 never overwrites
```
ch4r10t33r, line 220: **Medium: hybrid-cgroup edge case can silently miss a real quota.**

When both v1 and v2 lines appear in `/proc/self/cgroup`, v1 wins here. On hybrid systems where the cpu controller has actually moved to v2 but a v1 stub line is still present, the code then looks for a v1 mount, finds nothing, and falls back to affinity — missing the real v2 quota.

Rare on modern Linux but possible. Two options:
1. Comment that this matches `num_cpus` behavior and may miss v2-only quotas on hybrid systems.
2. If the v1 mount lookup fails in `cgroupsNumCpus`, fall through and retry with v2.

pkgs/cli/src/cpu_count.zig
```diff
@@ -0,0 +1,460 @@
...
+    const result = if (cgroupsNumCpus(gpa, io)) |quota| @min(quota, logical) else logical;
+
+    assert(result >= 1);
+    assert(result <= logical);
+    return result;
+}
+
+/// Logical CPU count from the affinity mask.
+fn logicalCpus() usize {
+    const n = std.Thread.getCpuCount() catch 1;
```
ch4r10t33r: **Medium: `getCpuCount` failure caps the result at 1.**

If `std.Thread.getCpuCount()` errors, `logical` becomes 1. Then `min(quota, 1) = 1` regardless of the cgroup quota — the pool gets sized to a single worker.

`getCpuCount` almost never fails in practice, but when it does the failure is invisible. A one-line `log.warn` on the error path would surface it in startup logs.

pkgs/cli/src/cpu_count.zig
```diff
@@ -0,0 +1,460 @@
...
+    // them on Linux means a real quota could be silently missed (the pool may
+    // over-size in a CPU-limited container). Parse-level misses below are the
+    // benign "no limit applies" cases (no cpu controller, cpu.max == "max").
+    const cgroup = readProcFile(gpa, io, cwd, "/proc/self/cgroup", .limited(1 << 20)) orelse {
+        log.warn("failed to read /proc/self/cgroup; sizing by affinity CPU count", .{});
+        return null;
+    };
+    defer gpa.free(cgroup);
+
+    const mountinfo = readProcFile(gpa, io, cwd, "/proc/self/mountinfo", .limited(8 << 20)) orelse {
```
ch4r10t33r: **Medium: 8 MB cap on `mountinfo` is generous — worth a comment.**

Containers with thousands of mounts can produce multi-MB `mountinfo`. 8 MB is safe, but the number looks arbitrary without context. A short comment explaining the rationale ("sized for pathological container mount counts") would stop a future reader from tightening it without thinking.

pkgs/cli/src/cpu_count.zig
GrapeBaBa: It is impossible

PR #988, opened 2026-06-10
fix(node): propose-path layering + publish latency + coverage report

GrapeBaBa:
Addresses three propose-path issues raised in review of the current architecture (delayed publishing, redundant STF for locally produced blocks, and the always-empty coverage report).

## 1. Proposer duty decision back in the validator client

`ValidatorClient.onInterval` now resolves `getSlotProposer` itself and calls the renamed `chain.submitPropose(node, slot, proposer_id)`; the chain keeps only chain-state gating (sync status) and the off-loop dispatch. The off-loop move itself (0f778f94) stays — the multi-second Type-2 merge must not run on the slot loop — but the duty decision no longer lives in the chain.

## 2. No redundant verify+STF on the local-block publish hop

**Archaeology:** until the per-resource locking migration (5fac5fd2, #805), `publishBlock` passed the produce-time cached post state to `onBlock`, which skipped recompute entirely for locally produced blocks. #805 removed that shortcut (a raw map pointer could no longer be carried across `states_lock.exclusive`), and as an undesigned side effect every locally produced block since has paid: parent state clone ×2 + **full proof re-verify of the proof this node just built** + **full state transition re-run** — all on the publish critical path (the network gossip send happens only after), with the recomputed state then **discarded** by `statesCommitKeepExisting` because produceBlock had already cached it.

`onBlock` now has a self-validating fast path: when BOTH the fork-choice entry and the cached post state for the block already exist (locally produced blocks on the publish hop; also duplicate re-imports, whose first import already verified them), it persists + confirms using the cached state and skips the recompute. Any probe miss falls through to the unchanged full path. This restores the pre-#805 publish semantics with borrow-discipline that satisfies the locking model (probe-and-release; no borrow held across exclusive sections). Preventing a corrupt proof from being *built* is the build path's responsibility and is being root-caused separately (instrumented runs in progress) — re-verifying our own freshly built proof on every block is not the fix for that.

## 3. Coverage report: window the data slots

The per-slot `last round attestation aggregate coverage` report filtered the payload maps by data slot `== slot-1`, but the aggregation pipeline matures payloads 2-3 slots after the attestation slot — measured result: **0 report emissions in 599 slots** while the same maps demonstrably held 31/32-coverage payloads (the proposal-side log showed them). The report now windows `(slot-4, slot]`, and the pre-merge snapshot is stamped with the newest data slot present instead of an arbitrary first hash-map iterator key. Unit test covers in-window / boundary / out-of-window / future cases.

## Validation

- `zig build` green; full `zig build test` green on the pre-rework revision; final-revision test run + CI in flight.
- Companion liveness fix: #986 (material-lead sync gating).

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -456,7 +456,7 @@ pub const BeamChain = struct {
 
     /// The block proof is a recursive-STARK Type-2 merge that takes seconds at the
     /// prod scheme — far longer than an 800ms interval. Running it on the libxev slot loop would
-    /// freeze gossip/tick handling, so `submitProposeOnInterval` offloads the whole propose
+    /// freeze gossip/tick handling, so `submitPropose` offloads the whole propose
     /// (produceBlock + buildBlockProof + publishBlock) to a `thread_pool` worker, mirroring the
     /// aggregate worker. At most one propose runs at a time; a second trigger is dropped (the next
     /// slot re-proposes). `proposeImpl` decrements on exit.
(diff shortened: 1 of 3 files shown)


pkgs/node/src/chain.zig
```diff
@@ -3636,6 +3636,88 @@ pub const BeamChain = struct {
...
+            }
+
+            // SSZ bytes for persistence — same live-serialize caveat as the
+            // full path: never pass the live SignedBlock to ssz.serialize.
+            var fp_fallback_ssz: std.ArrayList(u8) = .empty;
+            defer fp_fallback_ssz.deinit(self.allocator);
+            var fp_clone: types.SignedBlock = undefined;
+            var fp_clone_initialized = false;
+            defer if (fp_clone_initialized) fp_clone.deinit();
+            const fp_ssz_for_db: []const u8 = if (blockInfo.sszBytes) |precomputed| precomputed else blk: {
```
g11tech: why are we cloning block?

pkgs/node/src/chain.zig
```diff
@@ -3636,6 +3636,88 @@ pub const BeamChain = struct {
...
+        // produced blocks. This path restores those semantics: a proof this
+        // node just built is not re-verified against itself — preventing a
+        // corrupt proof from being built is the build path's job — and the
+        // transition result is not recomputed only to be discarded.
+        //
+        // The two probes make the path self-validating: it activates only when
+        // BOTH the fork-choice block and the cached post state already exist
+        // (which also covers duplicate re-imports, whose first import already
+        // verified them). Any miss falls through to the full path below.
+        fastpath: {
```
g11tech: why is this complication needed?
you should make sure that node.publishBlock should lookup and pass postState in the blockinfo so this is already on fast path

g11tech (commented):
also i don't understand where this PR fixes the issue that the pubishing is happening in delayed manner despite block production in time

pkgs/node/src/chain.zig
GrapeBaBa: You are right, reworked — the separate block is gone (0b153605). Now it is a ~15-line probe at the top of the computedstate branch: if forkchoice has the block and the states map has its post state, clone our own cached state and skip the parent clone + verify + transition. Everything downstream is the normal single flow.

Re passing postState from publishBlock: that is exactly what we had before 5fac5fd2, and it was removed for a reason — statesGet now returns a BorrowedState, and holding its read side across the exclusive commit inside onBlock deadlocks. Also the caller-supplied postState path deep-clones the state a second time inside onBlock, so it would cost two clones instead of one. Doing the lookup inside the branch gets the same effect with one clone and no lock games.

pkgs/node/src/chain.zig
GrapeBaBa: Gone with the rework. That was the fast path duplicating the existing clone-then-serialize fallback from the main path (serializing the live SignedBlock corrupts List/Bitlist state — see the comment on the main-path fallback). Only the original copy remains now.

ch4r10t33r (approved):
LGTM

PR #990, opened 2026-06-10
fix(forkchoice): prefer fresh high-target proposal attestations

zclawz:
## Summary

Fixes #989.

Block proposal attestation selection now ranks eligible `AttestationData` keys by:

1. highest unjustified `target.slot` first,
2. newest attestation `slot` when target ties,
3. payload count and lexicographic att-data order as deterministic tie-breakers.

This prevents stale low-target keys from filling the 8-entry `MAX_ATTESTATIONS_DATA` cap before fresh/high-target aggregates are considered, which was excluding subnet 0 from proposals during a justification stall.

## Edge cases covered

- high-target key beats a stale low-target key even when the stale key has a newer attestation slot;
- equal-target keys prefer the newest attestation slot;
- the 8-key proposal cap is not filled exclusively by stale low-target entries when fresh high-target entries exist.

## Validation

- `zig fmt --check pkgs/node/src/forkchoice.zig` ✅
- `zig build test --summary all` with `PATH=/usr/local/cargo/bin:$PATH` ran through the new forkchoice tests successfully:
  - `proposal attestation selection: high target beats stale low target even if low key is newer` ✅
  - `proposal attestation selection: newest attestation slot wins when target ties` ✅
  - `proposal attestation selection: 8-key cap is not filled exclusively by stale low targets` ✅
- The full suite timed out after 900s later in existing long-running tests (`EXIT:124`), after reaching node test 75/211; no failures were observed before the timeout.

### DIFF
--- a/pkgs/node/src/forkchoice.zig
+++ b/pkgs/node/src/forkchoice.zig
@@ -3345,113 +3345,6 @@ fn createTestRoot(fill_byte: u8) types.Root {
     return root;
 }
 
-fn makeProposalSelectionTestAttData(slot: types.Slot, target_slot: types.Slot, root_byte: u8) types.AttestationData {
-    const root = createTestRoot(root_byte);
-    const source_root = createTestRoot(0x01);
-    return .{
-        .slot = slot,
-        .head = .{ .root = root, .slot = slot },
-        .target = .{ .root = root, .slot = target_slot },
-        .source = .{ .root = source_root, .slot = 1 },
-    };
-}
-
-fn appendProposalSelectionPayloads(
-    allocator: Allocator,
-    list: *AggregatedPayloadsList,
-    count: usize,
-) !void {
-    for (0..count) |i| {
-        var proof = try types.SingleMessageAggregate.init(allocator);
-        errdefer proof.deinit();
-        try types.aggregationBitsSet(&proof.participants, @intCast(i), true);
-        try list.append(allocator, .{ .slot = @intCast(i), .proof = proof });
-    }
-}
-
-test "proposal attestation selection: high target beats stale low target even if low key is newer" {
-    var high = makeProposalSelectionTestAttData(90, 100, 0xA0);
-    var stale = makeProposalSelectionTestAttData(120, 10, 0xB0);
-    var high_payloads: AggregatedPayloadsList = .empty;
-    var stale_payloads: AggregatedPayloadsList = .empty;
-
-    var candidates = [_]ProposalAttestationCandidate{
-        .{ .att_data = &stale, .payloads = &stale_payloads },
-        .{ .att_data = &high, .payloads = &high_payloads },
-    };
-    std.mem.sort(ProposalAttestationCandidate, &candidates, {}, proposalAttestationCandidateLessThan);
-
-    try std.testing.expectEqual(@as(types.Slot, 100), candidates[0].att_data.target.slot);
-    try std.testing.expectEqual(@as(types.Slot, 10), candidates[1].att_data.target.slot);
-}
-
-test "proposal attestation selection: payload count wins before attestation slot when target ties" {
-    const allocator = std.testing.allocator;
-    var previous_slot = makeProposalSelectionTestAttData(99, 100, 0xA1);
-    var current_slot = makeProposalSelectionTestAttData(100, 100, 0xA2);
-    var previous_payloads: AggregatedPayloadsList = .empty;
-    defer deinitAggregatedPayloadsList(allocator, &previous_payloads);
-    var current_payloads: AggregatedPayloadsList = .empty;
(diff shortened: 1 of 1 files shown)


pkgs/node/src/forkchoice.zig
```diff
@@ -304,6 +304,101 @@ const StoredAggregatedPayload = types.StoredAggregatedPayload;
...
+/// oldest still-unjustified historical target. `AttestationData` identity
+/// includes the attestation slot, so a stalled node can retain hundreds of keys
+/// for the same logical vote. If we sort ascending by target slot, stale low
+/// targets fill the 8-key cap and fresh high-target aggregates (including whole
+/// subnets) never reach the block. Prefer the deepest target first, then the
+/// newest attestation slot for equal targets; use payload count and the shared
+/// lexicographic att-data order only as deterministic final tie-breakers.
+fn proposalAttestationCandidateLessThan(_: void, a: ProposalAttestationCandidate, b: ProposalAttestationCandidate) bool {
+    if (a.att_data.target.slot != b.att_data.target.slot) return a.att_data.target.slot > b.att_data.target.slot;
+    if (a.att_data.slot != b.att_data.slot) return a.att_data.slot > b.att_data.slot;
```
ch4r10t33r: Question on the tie-break order: when targets tie, the newest attestation slot wins before payload count is considered. Aggregation lags a slot or two, so the newest key for a target can be thin (1 payload) while the previous slot's key has full subnet coverage. Both usually fit inside the 8-key cap so this rarely matters, but if the cap is ever tight, ranking payload count above attestation slot on target ties would favor coverage over freshness. Was newest-first deliberate here (to match the #987 aggregator behavior), or worth flipping?

pkgs/node/src/forkchoice.zig
```diff
@@ -304,6 +304,101 @@ const StoredAggregatedPayload = types.StoredAggregatedPayload;
...
+        .{ .att_data = &previous_slot, .payloads = &previous_payloads },
+        .{ .att_data = &current_slot, .payloads = &current_payloads },
+    };
+    std.mem.sort(ProposalAttestationCandidate, &candidates, {}, proposalAttestationCandidateLessThan);
+
+    try std.testing.expectEqual(@as(types.Slot, 100), candidates[0].att_data.slot);
+    try std.testing.expectEqual(@as(types.Slot, 99), candidates[1].att_data.slot);
+}
+
+test "proposal attestation selection: 8-key cap is not filled exclusively by stale low targets" {
```
ch4r10t33r: These three tests only exercise the comparator on hand-built arrays — none of them runs `getProposalAttestationsUnlocked` against a populated `latest_known_aggregated_payloads` map. That's the path that broke in #989 (cap interplay, `processed_att_data` dedup, loop restarts after justification advances). Suggest adding one selection-level test: populate the map with >8 stale low-target keys plus fresh high-target keys carrying payloads, call the selection, and assert the result includes the high-target keys. That also covers items 1–2 of the issue's test plan.

Minor: the other 42 tests in this file (and `createTestRoot`, which these helpers call) live at the bottom (~line 3050+); consider moving these three down there for consistency.

ch4r10t33r (commented):
Direction is right: descending target matches where aggregators concentrate proofs (`selectJustificationAdvancingKey` picks the highest justification-advancing target), so the cap now holds keys that can actually justify. I checked the downstream path — the cap loop, `compactAttestations`, and the STF don't depend on the old ascending order, and the comparator is a total order (map keys are unique, so the final att-data tie-break always resolves). Two notes inline.

pkgs/node/src/forkchoice.zig
zclawz: Good call — for proposal building, coverage should win inside the same target. I flipped the target-tie order to prefer payload count before attestation slot, with newest slot retained as the next tie-breaker. The comment now documents why this differs from the #987 aggregator policy: builder selection is optimizing block coverage under the cap, not publish freshness.

pkgs/node/src/forkchoice.zig
zclawz: Addressed. I moved the comparator-only tests down with the rest of the forkchoice tests and added a selection-path regression that populates `latest_known_aggregated_payloads` with >8 stale low-target keys plus fresh high-target keys, calls `getProposalAttestations`, and asserts the high-target keys survive the 8-key cap. The new focused tests pass in the node suite.

ch4r10t33r (approved):
LGTM. Both review notes addressed — tie-break flipped to coverage-over-freshness within a target with a comment explaining why, and the new selection-path test drives `getProposalAttestations` end-to-end against a populated payload map (8 stale + 4 fresh) and asserts the cap retains the high-target keys. Clean fix for #989.

ch4r10t33r (approved):

PR #994, opened 2026-06-11
max aggregation flag & ssz update

anshalshukla:
- --max-aggregations-per-tick (default 1) - configurable cap on how many AttestationData the aggregator proves + publishes per tick. The keys form a greedy justification path (each prove advances latest_justified, the next is sourced at that target). Default 1 reproduces today’s single-aggregation behaviour exactly; 0 is floored to 1. Values > 1 run the extra proves in parallel on the shared ThreadPool — pair with a lower --rayon-threads to avoid CPU oversubscription. Threaded through NodeOptions → ForkChoice → aggregateUnlocked.

- Cached hashTreeRoot - bumped blockblaz/ssz.zig to pick up a drop-in TreeHasher and wrapped the hot BeamState collections (HistoricalBlockHashes, JustificationRoots, JustifiedSlots, JustificationValidators, Validators) so Merkle roots are cached incrementally instead of recomputed each call.


pkgs/node/src/forkchoice.zig
```diff
@@ -4135,45 +4215,125 @@ fn selectJustificationAdvancingKey(
...
+
+    var len: usize = 0;
+    var current = latest_justified;
+    while (len < max_count and len < out.len) {
+        const key = selectBestPreferredKey(snap, att_data_keys, current, out[0..len]) orelse break;
+        out[len] = key;
+        len += 1;
+        // Project justification forward: this key's target is the source the
+        // next chain step must match.
+        current = key.target;
```
ch4r10t33r, line 4325: The chain is speculative: key2 is sourced at key1's target on the assumption key1 actually justifies it. If that target doesn't get justified this round (e.g. not enough stake), the key2/key3 aggregates are sourced at a not-yet-justified checkpoint and add nothing until it does. Fine since it's bounded by max_count and opt-in (default 1) — worth a one-line note in the flag doc that the path is speculative.

ch4r10t33r (commented):
One note on the multi-key path.

ch4r10t33r (approved):
LGTM.

pkgs/node/src/forkchoice.zig
anshalshukla: key1 is actually based upon the local justified view of the node so either it is already justified across the network or will be eventually or there is some equivocation to happen

PR #996, opened 2026-06-12
metrics, node: add zeam_block_proof_merge_time_seconds (Type-2 merge latency by component count)

GrapeBaBa:
## What

Adds a dedicated metric for the **Type-2 multi-message STARK merge** in `buildBlockProof` — the recursive-STARK operation that folds the per-attestation Type-1 proofs + the proposer signature into the single `SignedBlock.proof`.

- **`zeam_block_proof_merge_time_seconds`** — a `HistogramVec` labeled by **`components`** = number of distinct `AttestationData` merged into the block proof. Buckets `0.1 … 5s`.
- Recorded in `chain.zig:buildBlockProof`, timing **only** `buildType2BlockProof` (taken *after* `prove_gate` is acquired, so it measures pure merge compute, not gate-wait).
- Also emits a log line: `Type-2 block-proof STARK merge slot=N components=C: X.XXXs`.

## Why

The merge sits on the **proposal critical path** between `produceBlock` and `publishBlock` — the `SignedBlock` can't be assembled until `.proof` (the Type-2) is built, so a block produced on time at i0 is published an interval (or more) late. Until now this cost was only visible buried inside the catch-all `lean_block_building_time_seconds` (whole `proposeImpl`) or inferable from produce→publish log gaps. A dedicated, component-labeled histogram makes the latency — and its scaling with block fullness — observable directly in prod (`rate(_sum)/rate(_count)` by `components`).

## Measured (3-validator beam sim, ReleaseFast)

| `components` | mean | min | max |
|---|---|---|---|
| 0 (proposer-sig-only proof) | **0.68s** | 0.62 | 0.76 |
| 1 | **1.19s** | 1.05 | 1.38 |

- **~0.68s floor** even for an empty block (irreducible recursive-STARK base cost).
- **~0.51s per additional `AttestationData`** → extrapolates to ~4.8s at `MAX_ATTESTATIONS_DATA=8`, consistent with prior recursive-merge measurements.

This quantifies why a timely block can't carry a freshly-merged proof under the per-interval budget, and motivates moving the merge off the propose→publish path (precompute) rather than bounding it with a deadline (a recursive-STARK prove can't be truncated mid-flight).

## Notes

- Pure additive metric; no behavioral change to consensus.
- No Rust touched. `zig fmt --check` passes.

### DIFF
--- a/pkgs/metrics/src/lib.zig
+++ b/pkgs/metrics/src/lib.zig
@@ -181,6 +181,8 @@ const Metrics = struct {
     // Block production metrics
     lean_block_building_time_seconds: BlockBuildingTimeHistogram,
     lean_block_building_payload_aggregation_time_seconds: BlockPayloadAggregationTimeHistogram,
+    // zeam-specific: Type-2 multi-message STARK merge time (buildBlockProof), the proposal critical-path cost.
+    zeam_block_proof_merge_time_seconds: BlockProofMergeTimeHistogram,
     lean_block_aggregated_payloads: BlockAggregatedPayloadsHistogram,
     lean_block_building_success_total: BlockBuildingSuccessCounter,
     lean_block_building_failures_total: BlockBuildingFailuresCounter,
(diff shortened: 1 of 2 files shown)


ch4r10t33r (dismissed):
LGTM

g11tech (dismissed):

g11tech (approved):

PR #998, opened 2026-06-14
add block proposal metrics and logs

anshalshukla:
Adds info logs and Prometheus counters marking the start/finish of each aggregation phase during block production, so the proposer critical path can be traced end-to-end from logs and dashboards.


ch4r10t33r (approved):
LGTM

PR #1002, opened 2026-06-16
forkchoice: finalized-from-canonical-head + gossip-validation hardening & head-slot pruning (leanSpec #1001, #1020)

GrapeBaBa:
## Summary

Ports two **leanSpec** fork-choice changes into zeam (the `#1001`/`#1020` below are *leanSpec* PR numbers, not zeam's):

### 1. Derive `latest_finalized` from the canonical head — leanSpec #1001 (`9b74af36`)
`fcStore.update` advanced `latest_finalized` by an independent monotonic max over every imported block's post-state. A fork that finalized a higher slot but then lost head selection latched its finalized checkpoint in the store, diverging from the head-state finalized the state transition validates against (`is_justifiable`) — **freezing finalization**.

Now `latest_finalized` is **derived from the canonical head chain** in `updateHeadUnlocked`, recomputed on every head change (import, acceptance ticks, proposals): each `ProtoNode` caches its post-state finalized slot (`stateFinalizedSlot`), and the head's parents are walked to the ancestor at that slot (with checkpoint-sync guards). Head selection still starts from justified — no clamp to finalized descendants. The independent-max branch (and its now-dead parameter) are removed.

### 2. Close the gossip-validation gap + prune by head slot — leanSpec #1020 (`875b794a`)
leanSpec #1020 added gossip-attestation rules — source must be an ancestor of target, target an ancestor of head, and the attestation slot must not precede the head slot — to reject attestations that could crash a node on a single message. zeam's production gossip path and the fork-choice spectest runner each carried a **separate, lighter copy** of the validation, and both lacked these rules: the live node accepted what the spec rejects, and the runner couldn't catch it because it validated against its own copy.

- **Converged** into one shared `ForkChoice.validateAttestationDataForGossip` (all fork-choice-derivable checks + the new #1020 rules, built on `isAncestorOf`). `chain.validateAttestationData` delegates to it (closing the live gap) and layers the BeamChain-only checks (signature, empty aggregation bits, per-participant registry). The runner deletes its copy and calls the same method, so the spectest now exercises the **production** predicate. The runner's signature/proof-failure detector accepts both the old `expectedError` and the new `rejectionReason` fixture fields, so it stays correct across spec versions.
- **Prune fix:** attestation pools are pruned by `head.slot` (not `target.slot`), matching leanSpec — an entry is stale once its head is at/below the finalized slot. Flipped across the signature map and both payload maps together.

## Verification
- 54/54 fork-choice unit tests pass (incl. a new regression test for the losing-fork latch).
- **Full spectest green at the committed leanSpec pin (e8014f9):** `All 471 tests passed` — the two stale-prune fixtures now pass, and the stricter gossip rules cause no regression.
- Cross-checked against the leanSpec #1001-merge fixtures (3bd2cd58): the #1020 gossip-negative, slot-before-head, empty-participants, registry, proof, and tick-to-slot vectors pass.
- `zig build` + `zig fmt` clean; merged latest `main`.

## Deferred (separate increment)
Bumping the leanSpec submodule pin past `e8014f9` is out of scope here — it requires spectest-runner schema rewrites for `networking_codec` & `slot_clock`, the `rejectionReason` rename in those runners, 3 new state-transition consensus rules, and one interval-2 aggregation-rotation gap.

### DIFF
--- a/pkgs/cli/src/test_driver.zig
+++ b/pkgs/cli/src/test_driver.zig
@@ -565,28 +565,11 @@ fn parseNonNegativeU64(value: JsonValue) !u64 {
 }
 
 fn validateAttestationDataForGossip(driver: *ForkChoiceDriverState, data: types.AttestationData) !void {
-    const source_node = driver.fork_choice.getProtoNode(data.source.root) orelse return error.UnknownSourceBlock;
-    const target_node = driver.fork_choice.getProtoNode(data.target.root) orelse return error.UnknownTargetBlock;
-    const head_node = driver.fork_choice.getProtoNode(data.head.root) orelse return error.UnknownHeadBlock;
-
-    if (data.source.slot > data.target.slot) return error.SourceCheckpointExceedsTarget;
-    if (data.head.slot < data.target.slot) return error.HeadOlderThanTarget;
-    if (source_node.slot != data.source.slot) return error.SourceCheckpointSlotMismatch;
-    if (target_node.slot != data.target.slot) return error.TargetCheckpointSlotMismatch;
-    if (head_node.slot != data.head.slot) return error.HeadCheckpointSlotMismatch;
-
-    if (!driver.fork_choice.checkpointIsAncestor(data.source, data.target)) {
-        return error.SourceCheckpointNotAncestorOfTarget;
-    }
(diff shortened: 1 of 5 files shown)


ch4r10t33r (approved):
LGTM. please merge once CI completes.

PR #1003, opened 2026-06-16
type1 aggregation in interval 4

anshalshukla:
Moves the proposer’s Type-1 (single-message attestation) aggregation off the interval-0 critical path. The node that will propose slot N now pre-aggregates its block’s attestation set one interval early (interval-in-slot 4 of slot N-1), so at proposal time the Type-1 work is already folded into fork-choice and the interval-0 window goes almost entirely to the Type-2 block-proof merge.

Release note: this renames the CLI/config deadline knob from `--proposal-deadline-pct` / `proposal_deadline_pct` to `--type1-aggregation-deadline-pct` / `type1_aggregation_deadline_pct`, and changes the default deadline budget from 50% to 20%. Deployment scripts and Ansible configs passing the old flag need to be updated before rollout.


ch4r10t33r (commented):
A few notes on the interval-4 warm-up.

pkgs/node/src/forkchoice.zig
```diff
@@ -4309,6 +4348,26 @@ fn selectBestPreferredKey(
 /// single-aggregation behaviour.
 ///
 /// Pure / unit-testable: no I/O, no allocator, no locking.
+/// Select every proposal-eligible att_data key (time-forward: target strictly
+/// after source, which excludes genesis source==target self-votes — the proposer
+/// drops those at build time, so pre-aggregating them is wasted prove). Per-key
+/// no-improvement skipping is handled later by prepareAggregateAttData /
+/// anyStoredProofCoversAll, so this applies only the cheap eligibility filter.
+/// `out` must be sized >= `att_data_keys.len`; returns the filled prefix.
+fn selectProposalEligibleKeys(
```
ch4r10t33r: This selects every eligible att_data with no cap, and the warm-up then proves them in parallel via computeAggregatedSignaturesForKeys (each prove drives rayon internally). That is the same ThreadPool x rayon oversubscription #925 sequentialized to avoid. With several att_data on a box with few cores, each prove runs slower and the whole batch can take much longer than its serial cost. Since the warm-up has the full interval, consider capping the parallelism or proving serially here. Do you have numbers on how long the parallel warm-up takes for a realistic key count?

pkgs/node/src/chain.zig
```diff
@@ -5181,6 +5217,181 @@ pub const BeamChain = struct {
...
+        var window_buf: [2]types.Slot = undefined;
+        var window_len: usize = 0;
+        if (slot > 0) {
+            window_buf[window_len] = slot - 1;
+            window_len += 1;
+        }
+        window_buf[window_len] = slot;
+        window_len += 1;
+
+        self.prove_gate.lock();
```
ch4r10t33r: This is a blocking lock, and the interval-0 getProposalAttestations path also takes prove_gate blocking. If the interval-4 warm-up is still running when interval-0 starts (it holds prove_gate for the whole batch, and the FFI cannot be interrupted), the proposal waits here instead of proceeding. That can delay the very block this change is meant to speed up. Consider tryLock (skip the refresh and compact what is already warmed) or a tight deadline so a slow warm-up never blocks the proposal.

pkgs/cli/src/main.zig
```diff
@@ -99,10 +99,10 @@ pub const NodeCommand = struct {
     /// for fewer sub-threshold aggregates on chatty subnets.
     @"min-aggregation-inputs": u32 = types.default_min_aggregation_inputs,
     /// Single-message attestation aggregation deadline budget (see
-    /// `ChainOpts.proposal_deadline_pct`): the proposer truncates attestation
+    /// `ChainOpts.type1_aggregation_deadline_pct`): the proposer truncates attestation
     /// gathering at this percent of the proposal interval; the remainder is for
     /// block signing and the Type-2 multi-message merge.
-    @"proposal-deadline-pct": u32 = node_lib.default_proposal_deadline_pct,
+    @"type1-aggregation-deadline-pct": u32 = node_lib.default_type1_aggregation_deadline_pct,
```
ch4r10t33r, line 149: This renames --proposal-deadline-pct to --type1-aggregation-deadline-pct and changes the default from 50 to 20. Any scripts or ansible passing the old flag will fail to parse. Worth calling out in the PR body / release notes.

ch4r10t33r (approved):
LGTM

PR #1008, opened 2026-06-24
deps: zig-libp2p v0.2.27 (zquic v1.7.57) — QUIC connection sharding + gossip/req-resp fixes → devnet finalizes

ch4r10t33r:
**Dependency-only** bump (`build.zig.zon` only — zero zeam-side code changes): `zig_libp2p` → **v0.2.27** (which pins **zquic v1.7.57**). This is the cumulative transport / gossip / mesh-formation work that took the 32-validator, 4-subnet QUIC devnet from *"produces blocks but never finalizes"* to **finalizing**. Supersedes #1006.

## Result
The devnet now **finalizes** — `lean_latest_justified_slot` and `lean_latest_finalized_slot` both advance and track the head, with subnet aggregation reaching 6–8/8. Connected peers hold at the full mesh (31/31).

### QUIC connection sharding — the throughput foundation
A single drive thread saturated at the full 32-peer mesh (multi-hundred-ms iterations → no-ACK teardowns → mesh churn). Connections are now **sharded across N drive threads** (quinn model): one demux thread routes datagrams to the owning shard by shard-tagged connection-ID; a peer→shard ownership table routes all directed work; gossip-outbox + hook-queue + conn-manager are funnelled/per-shard. `drive_shards` **auto-detects from core count**, behaviour-preserving at N=1.

### Gossipsub / mesh formation — the finality unlock
- **`mesh_n_low` 6 → 8** so sparse per-subnet `attestation_<n>` topic meshes converge to (near-)full — aggregators were parked at 5–6/8, never reaching ⅔ quorum.
- **Priority/bulk atomic-frame `/meshsub` outbox** — large blocks no longer head-of-line-block (nor wire-desync) the small, time-sensitive attestation frames sharing the per-peer stream.
- **Wedge-reopen-in-place** — a flow-control-stalled gossip stream reopens on the *same* connection instead of tearing it down, so peers hold the full mesh instead of dropping and failing to re-dial.
- Inbound gossip-validation backlog hardening (batch-drain + never-drop-blocks).

### QUIC transport reliability
- **req/resp slot-leak fix** — every raw-app stream-open path (req/resp, gossip, publish, identify, autonat) now releases its zquic stream slot (+ a 15s orphan reaper); fixes the inbound-leg req/resp fallback failing permanently after ~64 requests. Plus uncapped server-initiated stream tracking and a zquic retired-slot guard against retransmit slot-resurrection.
- **Flow-control windows** widened 10/15 → 16/24 MiB (absorbs transient read-lag on long-lived gossip streams).
- onAck made O(acked) (deque), non-reaping ACK flush, interleaved inbound socket drain (earlier mesh-churn fixes).

## Pins
`zig-libp2p v0.2.27` → `zquic v1.7.57` — both released and merged to their respective mainlines. Deps-only; no zeam code changes required (the runtime talks to `NetworkInterface`, not transport internals).

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -61,8 +61,8 @@
             .hash = "thread_pool-0.1.0-m6B8f9j4AAAWx5c9ytLrH_NlntTMxSg049v76xVsgoSw",
         },
         .zig_libp2p = .{
-            .url = "https://github.com/blockblaz/zig-libp2p/archive/refs/tags/v0.2.8.tar.gz",
-            .hash = "zig_libp2p-0.2.8-lil2hMMAJgB0siAnkR_hJS_rMSNip0Y9MqloAV0Wei2k",
+            .url = "https://github.com/blockblaz/zig-libp2p/archive/refs/tags/v0.2.9.tar.gz",
+            .hash = "zig_libp2p-0.2.9-lil2hOoJJgC7Z6D39lgDRfggxa3u53CKhskqm_5extmd",
         },
     },
     .paths = .{""},
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -61,8 +61,8 @@
             .hash = "thread_pool-0.1.0-m6B8f9j4AAAWx5c9ytLrH_NlntTMxSg049v76xVsgoSw",
         },
         .zig_libp2p = .{
-            .url = "https://github.com/blockblaz/zig-libp2p/archive/refs/tags/v0.2.21.tar.gz",
-            .hash = "zig_libp2p-0.2.21-lil2hOJjJgDauxNPFNTHlMPz99a_nPRzNPCy0gPXzQOh",
+            .url = "https://github.com/blockblaz/zig-libp2p/archive/refs/tags/v0.2.22.tar.gz",
+            .hash = "zig_libp2p-0.2.22-lil2hDptJgAY41I9uPkFoDHK2FepMK9cEoUAZwUmB5EB",
         },
     },
     .paths = .{""},
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -61,8 +61,8 @@
             .hash = "thread_pool-0.1.0-m6B8f9j4AAAWx5c9ytLrH_NlntTMxSg049v76xVsgoSw",
         },
         .zig_libp2p = .{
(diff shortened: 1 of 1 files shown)


g11tech (approved):

build.zig.zon
```diff
@@ -61,8 +61,8 @@
             .hash = "thread_pool-0.1.0-m6B8f9j4AAAWx5c9ytLrH_NlntTMxSg049v76xVsgoSw",
         },
         .zig_libp2p = .{
-            .url = "https://github.com/blockblaz/zig-libp2p/archive/refs/tags/v0.2.8.tar.gz",
-            .hash = "zig_libp2p-0.2.8-lil2hMMAJgB0siAnkR_hJS_rMSNip0Y9MqloAV0Wei2k",
+            .url = "https://github.com/blockblaz/zig-libp2p/archive/refs/tags/v0.2.27.tar.gz",
+            .hash = "zig_libp2p-0.2.26-lil2hN15KABKqKziF62qP3iW26XbtL-0Z17ErgwjII_F",
```
GrapeBaBa, line 65: why hash has 0.2.26 prefix?

ch4r10t33r: Good catch, but it's expected. The version segment in a Zig package hash comes from the dependency's own `build.zig.zon` `.version` field, not the git tag. zig-libp2p's v0.2.27 release didn't bump that field — it's still `.version = "0.2.26"` on both the v0.2.26 and v0.2.27 tags — so the prefix reads `0.2.26` either way.

The content multihash is what actually pins the code, and it's correct for v0.2.27. Verified with `zig fetch`:

```
v0.2.27 -> zig_libp2p-0.2.26-lil2hN15KABKqKziF62qP3iW26XbtL-0Z17ErgwjII_F   (== this PR)
v0.2.26 -> zig_libp2p-0.2.26-lil2hKV2JgCuv5ABt4SLeYtDWJADnODSv4M0nc-DCqmD
```

Same `0.2.26` prefix, different multihash — the PR resolves the v0.2.27 tarball. The stale `.version` is a zig-libp2p packaging-hygiene issue; I'll bump it in the next release so the prefix matches the tag.

PR #1022, opened 2026-07-03
node: make source-side attestation checks gossip-only in fork choice

GrapeBaBa:
## Problem

Block-included attestations are validated via `chain.validateAttestationData(data, is_from_block=true)`, which delegates to `forkChoice.validateAttestationDataForGossip`. Its source/target/head existence + ancestry checks query the **proto-array**. Finalization prunes pre-finalized ancestors from proto-array (`forkChoice.rebase`), so a valid on-chain vote whose **source** sits at/below the finalized boundary is dropped (`UnknownSourceBlock`, or `SourceCheckpointNotAncestorOfTarget`) and never applied to the LMD tracker.

Under catch-up sync (finalization advancing in jumps) this fires on nearly every block-included attestation. An observed devnet run: **728 such drops across 292 slots** while the node stalled — finalized frozen, head stuck, own duties silenced by the behind-peers gate.

## Why source doesn't belong in this check (for block attestations)

- `source` is **FFG-only**; the LMD tracker consumes only `head`.
- For a block attestation, source/target/head are already validated by the STF (`process_attestations`: `has_correct_source_root`, `is_source_justified`, `has_correct_head_root`) against the block's **own canonical chain** — the correct reference frame.
- For reference: beacon's `validate_on_attestation` never checks source at all; leanSpec validates block attestations in the STF, not the fork-choice gate.

So the source-side fork-choice checks are redundant for block attestations and only false-reject on pruning.

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -4677,19 +4677,26 @@ pub const BeamChain = struct {
         // All fork-choice-derivable validation lives in one shared method on
         // ForkChoice (proto-array existence/slots, source/target/head
         // ancestry, slot-before-head, and the interval-based future-slot
-        // bound). The future-slot bound applies only to gossip — block-
-        // included attestations are trusted under the block's own validation.
+        // bound). `is_gossip` (= !is_from_block) selects the checks that apply
+        // only to a wire vote; a block-embedded attestation is trusted under the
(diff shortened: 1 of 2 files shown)


ch4r10t33r (approved):
LGTM

PR #1024, opened 2026-07-03
node: load finalized state by forkchoice root

zclawz:
## Summary
- make `getFinalizedState()` cache DB-loaded states by the exact forkchoice finalized root
- make the DB fallback load `DbStatesNamespace[finalized_checkpoint.root]` instead of `loadLatestFinalizedState()`
- avoid keying the cache on the returned state's embedded `latest_finalized`, which can legitimately lag the state/block being served

## Why
Hive rpc-compat test 204 waits until `/lean/v0/fork_choice` reports a non-genesis finalized checkpoint, then requires `/lean/v0/states/finalized` to return a state whose slot reaches that observed finalized slot.

The old failing Hive result was against the stale `b95f3f1` image, and the latest image has not yet had a completed public Hive run. But the current code still had a possible mismatch: if the finalized state was no longer in memory, the RPC fallback reloaded "database latest finalized" rather than the exact finalized root already reported by forkchoice. If that side metadata lagged or was stale, `/states/finalized` could serve an older state than `/fork_choice` just advertised.

This PR makes the fallback root-addressed so the endpoint follows the forkchoice checkpoint directly.

## Validation
- `zig fmt pkgs/node/src/chain.zig`
- `git diff --check`

`zig build test --summary all` could not run in this workspace because dependency build scripts fail against the installed Zig 0.15.2 before reaching zeam code, e.g. `Build.Graph` has no field `io` in `gremlin`, `libxev`, and `zig_yaml` build scripts.

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -264,13 +264,15 @@ pub const BeamChain = struct {
     /// Lock-free monotonic counter; safe to read from any thread.
     states_kept_existing_count: std.atomic.Value(u64) = std.atomic.Value(u64).init(0),
     // Cached finalized state loaded from database (separate from states
-    // map to avoid affecting pruning).
+    // map to avoid affecting pruning). The companion root is the forkchoice
+    // finalized block root used to load this state.
     //
     // Migrated from raw `*types.BeamState` to `*RcBeamState` so the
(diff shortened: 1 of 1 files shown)


ch4r10t33r (approved):
LGTM

PR #1026, opened 2026-07-05
Optimizations

anshalshukla:
- level 2 aggregation in interval 3
- smart block deconstruction
- latest justified source attestation selection
- metrics additoins

### DIFF
--- a/pkgs/node/src/chain.zig
+++ b/pkgs/node/src/chain.zig
@@ -399,6 +399,21 @@ pub const BeamChain = struct {
     // `cacheBlockAndFetchParent`).
     rejected_block: ?BlockCallback(RejectedBlockFn) = null,
 
+    // -- deconstructed-aggregate publish backchannel ---------------------
+    //
+    // Fires from the `onBlock` re-aggregation path (aggregators only) for
+    // each per-attestation Type-1 recovered by splitting the imported
+    // block's merged Type-2 proof. Lets the libxev-side node publish the
+    // recovered proof as a `SignedAggregatedAttestation` on the aggregation
+    // gossip topic (the chain has no `node` pointer in `onBlock`).
+    //
+    // Ownership: the callback takes ownership of the passed
+    // `SignedAggregatedAttestation` (owned-proof clone) and MUST deinit it
+    // — `BeamNode.publishCommittedAggregation` already does. Threading: same
+    // off-loop chain-worker / propose-worker context as the aggregate worker,
+    // which already calls `publishCommittedAggregation` from a pool worker.
+    publish_aggregate: ?BlockCallback(PublishAggregateFn) = null,
+
     // Queue for blocks that arrived before forkchoice had ticked to their slot.
     // When a peer gossips a block for the current slot before our local interval
     // timer fires, the forkchoice rejects it with FutureSlot.  We hold such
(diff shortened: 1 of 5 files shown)


g11tech (approved):
lgtm on high level, get review of @ch4r10t33r as well

ch4r10t33r (commented):
@anshalshukla this PR reverts https://github.com/blockblaz/zeam/pull/1022 ? I believe @GrapeBaBa 's changes were introduced to fix an issue identified by Ethlambda team in their local testing.

ch4r10t33r (commented):
Two notes. Also worth rebasing: this branch is a few commits behind main (its base predates #1022), so please rebase and re-run tests before merge.

pkgs/node/src/forkchoice.zig
```diff
@@ -1516,11 +1516,19 @@ pub const ForkChoice = struct {
...
+        // space. Use pre_state.latest_justified (the head state the proposer builds
+        // on, which equals the fork-choice store's justified at the canonical head)
+        // — the STF-consistent reference the rest of this selector already uses.
+        const pre_state_justified = pre_state.latest_justified;
         {
             var payload_it = self.latest_known_aggregated_payloads.iterator();
             while (payload_it.next()) |entry| {
                 const att_data = entry.key_ptr.*;
                 if (att_data.target.slot <= att_data.source.slot) continue;
+                if (!att_data.source.eql(&pre_state_justified)) continue;
```
ch4r10t33r, line 1531: This narrows block packing to only attestations sourced at the current justified checkpoint, but the STF accepts any justified source: process_attestations gates on isSlotJustified(source_slot), which is true for any slot in justified_slots, not just the latest. So an attestation sourced at an older-but-still-justified checkpoint can justify its target in the STF, and dropping it here can lose finality-relevant votes in recovery/catch-up scenarios. This also reverses the earlier deliberate behavior (the old test: build_block must accept any attestation whose source slot is in justified_slots). Can you justify latest-source-only against the lean 3SF finalization rule before this lands?

pkgs/node/src/chain.zig
```diff
@@ -3966,31 +4015,49 @@ pub const BeamChain = struct {
...
+                }
+
                 var split_mask = self.allocator.alloc(bool, aggregated_attestations.len) catch break :reagg;
                 defer self.allocator.free(split_mask);
                 var selected: usize = 0;
                 for (aggregated_attestations, 0..) |agg_att, i| {
                     split_mask[i] = false;
                     if (selected >= MAX_REAGGREGATIONS_PER_BLOCK) continue;
-                    if (agg_att.data.target.slot <= justified_slot) continue;
+                    if (!agg_att.data.source.eql(&parent_justified)) continue;
```
ch4r10t33r, line 4062: Same concern as the block-packing filter: this drops re-aggregation for any source that is not exactly the parent state's latest justified, but the STF justifies from any justified source. If the latest-source-only rule is correct, please keep the two filters and the finality argument together (ideally split this source-selection change into its own PR from the interval-3 aggregation and metrics, since it is a consensus-semantic change and the rest is additive).

pkgs/node/src/forkchoice.zig
anshalshukla: yeah but since now the attestation are there from state justified as source than we do not need to depend on those attestations, it was required when source was store justified

ch4r10t33r (approved):
LGTM

